PR #25, opened 2017-05-05
Eric wip

ebariaux:


manager/server/src/main/java/org/openremote/manager/server/security/ManagerIdentityService.java
```diff
@@ -231,6 +232,13 @@ public UserConfiguration mergeUserConfiguration(UserConfiguration userConfigurat
         return persistenceService.doReturningTransaction(em -> em.merge(userConfiguration));
     }
 
+    public boolean isUserInTenant(String userId, String tenantId) {
```
christianbauer: The parameter name should be realmId instead of tenantId, we use 

- "tenantRealm" or "realm" for the unique realm name
- "realmId" or "id" for Keycloak entity identifier
- "tenantDisplayName" or "displayName" for what is shown to users

manager/server/src/main/resources/demo/rules/DemoApartmentNotificationWhenAlarm.drl
```diff
@@ -0,0 +1,17 @@
...
+import elemental.json.Json;
+
+global java.util.logging.Logger LOG;
+global org.openremote.model.rules.Users users;
+global org.openremote.model.rules.Assets assets;
+
+rule "Send notification when alarm triggered in a room"
+when
+    $residence : AssetState(type == AssetType.RESIDENCE, attributeName == "alarmEnabled", value.asBoolean)
+    $room : AssetState(parentId == $residence.id, type == AssetType.ROOM, attributeName == "presenceDetected", value.asBoolean)
```
christianbauer: value.asBoolean should be valueAsBoolean with the new value model

manager/shared/src/main/java/org/openremote/manager/shared/security/User.java
```diff
@@ -19,20 +19,47 @@
...
 import javax.validation.constraints.Size;
 
+/**
+ * This can be used (among other things) to query the Keycloak USER_ENTITY table in JPA queries.
+ */
+@Entity
+@Subselect("select * from USER_ENTITY") // Map this immutable to an SQL view, don't use/create table
 public class User {
 
+    @Formula("(select ra.VALUE from REALM_ATTRIBUTE ra where ra.REALM_ID = REALM_ID and ra.name = 'displayName')")
```
christianbauer: This should probably be the "tenantRealm"/"realm" unique name and not the display name, see previous comment.

manager/test/src/test/groovy/org/openremote/test/rules/apartment/ApartmentNotificationWhenAlarmTest.groovy
```diff
@@ -0,0 +1,66 @@
...
+import org.openremote.manager.server.setup.SetupService
+import org.openremote.manager.server.setup.builtin.ManagerDemoSetup
+import org.openremote.model.AttributeEvent
+import org.openremote.test.ManagerContainerTrait
+import spock.lang.Specification
+import spock.util.concurrent.PollingConditions
+
+import static org.openremote.manager.server.setup.builtin.ManagerDemoSetup.DEMO_RULE_STATES_APARTMENT_1
+
+/**
```
christianbauer: Please disable this in the IDE templates

PR #93, opened 2018-10-02
Controller protocol

jvervier:
Hi,
Here is the new Controller Protocol code.

There is only one little issue as I can't have a Controller Protocol in the attribute type list in an Agent (did I forget something or using wrong docker profile? I used demo one).

From Http Client, we used the Request classes to manage requests.


richturner:
I see the controll protocol listed in the manager:

image

jvervier:
I missed something in execution on my local, it's ok too on my side :-)

richturner:
OK great, I couldn't get the test to pass so made some changes there, also went through your code and made a few minor changes mostly relating to multi threading but please take a look and we can discuss if necessary.

jvervier:
Ok, I check everything, I'll made a check by finishing documentation with the demo controller today.

Should I do the merge after or is it your role?

richturner:
OK great, feel free to do the merge when you are ready.

PR #94, opened 2018-10-25
Fixing issue and null case in polling and heartbeat

jvervier:
During my preparation of a demo, I notice two little problems : 

1. A null case not managed in polling
2. A "death alive" task 'heartbeat' running in some scenario (fixed by setting a flag to true)

I'm waiting for your approval and review :-)


jvervier:
As we notice strange behaviour in test with Eric, we add an initial status check and fix socket timeout issue

Issue #111, opened 2019-03-06
or-map uses @polymer


milovanderlinden:
openremote/ui/component/or-map/src/markers/or-map-marker-asset.ts references @polymer/polymer and uses `observe` which is not available in lit-element. I will create a pull request for this issue

milovanderlinden:
This is also the case for or-map-marker.ts

PR #112, opened 2019-03-06
Fix for #111

milovanderlinden:


richturner:
Thanks for the pull request Milo, I have resolved this issue as part of a larger or-map and or-map-marker update.

PR #165, opened 2020-04-22
Asset groups

vlichtenberg:


ui/component/or-asset-viewer/src/index.ts
```diff
@@ -303,12 +321,19 @@ export class OrAssetViewer extends subscribe(manager)(translate(i18next)(LitElem
...
             return;
         }
 
         return html`
+            <style>
+               .panel {
+                    position: relative;
+               }
+            </style>
+
```
richturner, line 336: I'm going to move this style into style.ts; keeps styles in a central place and avoids creating a style DOM node for every panel

PR #170, opened 2020-05-12
Attribute detail card improvements

vlichtenberg:


ui/app/main/src/components/pages/page-insights.ts
```diff
@@ -110,8 +111,17 @@ const viewerConfig: DataViewerConfig = {
             gridRowStart: "1"
         },
         defaults: [{
-            assetId:"4m4dWZUAV4IepHKc8Wwv85",
+            assetId:"35Q8RvOhBU3boCB6IUW1Dj",
```
richturner, line 114: Guess this is for your testing purposes and will be removed

ui/component/or-attribute-card/src/index.ts
```diff
@@ -131,21 +221,24 @@ export class OrAttributeCard extends LitElement {
...
             style
         ];
     }
 
     connectedCallback() {
         super.connectedCallback();
-        this._style = window.getComputedStyle(this);
+
+        if (!this.assetId || !this.attributeName) { return false; }
+
```
richturner, line 236: These properties can be changed during the lifecycle of the component so shouldn't be checked here but in `shouldUpdate` or `updated`

richturner: `connectedCallback` and `disconnectedCallback` should be used for attaching/detaching event handlers etc. to prevent memory leaks

ui/component/or-attribute-card/src/index.ts
```diff
@@ -221,53 +375,125 @@ export class OrAttributeCard extends LitElement {
...
+                            data-mdc-dialog-action="yes"
+                            @click="${this._setAttribute}"></or-input>
+
+                    </footer>
+                    </div>
+                </div>
+                <div class="mdc-dialog__scrim"></div>
+            </div>
+        `;
+
```
richturner, line 428: Should be using `or-mwc-dialog` like `or-asset-viewer` does

ui/app/main/src/components/pages/page-insights.ts
vlichtenberg: Yes. Robin has put this into his version of the insights page, probably as a working stub, and I connected the KPI cards to it (and therefore I had to fit this in with "my local manager build"). This will be dropped once the localstorage functionality comes in.

PR #176, opened 2020-05-19
Mqtt broker

Miggets7:


container/src/main/java/org/openremote/container/security/ClientCredentialsAuthFrom.java
```diff
@@ -0,0 +1,43 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.openremote.container.security;
+
+import javax.ws.rs.FormParam;
+
+public class ClientCredentialsAuthFrom extends AuthForm {
```
richturner, line 24: Typo in class name

container/build.gradle
```diff
@@ -58,8 +58,10 @@ dependencies {
         exclude group: "logkit"
         exclude group: "commons-logging"
     }
+
+    compile "io.moquette:moquette-broker:$moquetteVersion"
```
richturner, line 62: Think this should be in the manager module not the container

manager/src/main/java/org/openremote/manager/mqtt/AssetInterceptHandler.java
```diff
@@ -0,0 +1,190 @@
...
+                CancelEventSubscription<AttributeEvent> cancelEventSubscription = new CancelEventSubscription<>(AttributeEvent.class, subscriptionId);
+                messageBrokerService.getProducerTemplate().sendBodyAndHeader(ClientEventService.CLIENT_EVENT_QUEUE, cancelEventSubscription, WebsocketConstants.SESSION_KEY, connection.clientId);
+            }
+        }
+    }
+
+    @Override
+    public void onMessageAcknowledged(InterceptAcknowledgedMessage interceptAcknowledgedMessage) {
+
+    }
```
richturner, line 180: Can this override be removed as unused

manager/src/main/java/org/openremote/manager/mqtt/AssetInterceptHandler.java
```diff
@@ -0,0 +1,190 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Logger;
+
+import static org.openremote.manager.mqtt.MqttBrokerService.TOPIC_SEPARATOR;
+import static org.openremote.model.Constants.KEYCLOAK_CLIENT_ID;
+
+public class AssetInterceptHandler implements InterceptHandler {
```
richturner, line 54: Maybe give this a more generic name rather than `AssetInterceptHandler` as we could allow other `SharedEvent`s to be pubbed/subbed over `MQQT`

manager/src/main/java/org/openremote/manager/mqtt/AssetInterceptHandler.java
```diff
@@ -0,0 +1,190 @@
...
+        MqttConnector.MqttConnection connection = mqttConnector.createConnection(interceptConnectMessage.getClientID(), interceptConnectMessage.getUsername(), interceptConnectMessage.getPassword());
+        String suppliedClientSecret = new String(interceptConnectMessage.getPassword(), StandardCharsets.UTF_8);
+        connection.accessToken = identityProvider.getKeycloak().getAccessToken(connection.realm, new ClientCredentialsAuthFrom(connection.username, suppliedClientSecret)).getToken();
+    }
+
+    @Override
+    public void onDisconnect(InterceptDisconnectMessage interceptDisconnectMessage) {
+        MqttConnector.MqttConnection connection = mqttConnector.removeConnection(interceptDisconnectMessage.getClientID());
+        if (connection != null) {
+            CancelEventSubscription<AttributeEvent> cancelEventSubscription = new CancelEventSubscription<>(AttributeEvent.class);
```
richturner, line 102: Should use `EventSubscriptions:cancelAll()` to remove any subscriptions this client has made

manager/src/main/java/org/openremote/manager/mqtt/AssetInterceptHandler.java
```diff
@@ -0,0 +1,190 @@
...
+            messageBrokerService.getProducerTemplate().sendBodyAndHeader(ClientEventService.CLIENT_EVENT_QUEUE, cancelEventSubscription, WebsocketConstants.SESSION_KEY, connection.clientId);
+        }
+    }
+
+    @Override
+    public void onConnectionLost(InterceptConnectionLostMessage interceptConnectionLostMessage) {
+        MqttConnector.MqttConnection connection = mqttConnector.removeConnection(interceptConnectionLostMessage.getClientID());
+        if (connection != null) {
+            CancelEventSubscription<AttributeEvent> cancelEventSubscription = new CancelEventSubscription<>(AttributeEvent.class);
+            messageBrokerService.getProducerTemplate().sendBodyAndHeader(ClientEventService.CLIENT_EVENT_QUEUE, cancelEventSubscription, WebsocketConstants.SESSION_KEY, connection.clientId);
```
richturner, line 112: Should use `EventSubscriptions:cancelAll()` to remove any subscriptions this client has made

manager/src/main/java/org/openremote/manager/mqtt/AssetInterceptHandler.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private Map<String, Object> prepareHeaders(MqttConnector.MqttConnection connection) throws VerificationException {
+        AccessToken accessToken = AdapterTokenVerifier.verifyToken(connection.accessToken, identityProvider.getKeycloakDeployment(connection.realm, KEYCLOAK_CLIENT_ID));
+        Map<String, Object> headers = new HashMap<>();
+        headers.put(Constants.AUTH_CONTEXT, new AccessTokenAuthContext(connection.realm, accessToken));
+        headers.put(WebsocketConstants.SESSION_KEY, connection.clientId);
+        headers.put(ClientEventService.HEADER_CONNECTION_TYPE, ClientEventService.HEADER_CONNECTION_TYPE_MQTT);
+        return headers;
+    }
+}
```
richturner, line 190: Make this static

manager/src/main/java/org/openremote/manager/mqtt/KeycloakAuthorizatorPolicy.java
```diff
@@ -0,0 +1,111 @@
...
+
+        String assetId = topic.getTokens().get(1).toString();
+        Asset asset = assetStorageService.find(assetId);
+        if (asset == null) {
+            LOG.info("Asset not found for assetId: " + assetId);
+            return false;
+        }
+
+        if (!asset.getRealm().equals(connection.realm)) {
+            LOG.info("Asset not in same clientId");
```
richturner, line 81: Probably should say `Asset not in the same realm`

manager/src/main/java/org/openremote/manager/mqtt/MqttConnector.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Logger;
+
+import static org.openremote.manager.mqtt.KeycloakAuthenticator.MQTT_CLIENT_ID_SEPARATOR;
+
+public class MqttConnector {
```
richturner, line 13: This class seems to have little purpose...can it be removed?

proxy/haproxy.cfg
```diff
@@ -68,5 +68,18 @@ frontend https
 
     use_backend proxy_backend
 
+listen mqtt
```
richturner, line 71: Have you tried this out?

test/src/test/groovy/org/openremote/test/notification/NotificationTest.groovy
```diff
@@ -581,15 +581,16 @@ class NotificationTest extends Specification implements ManagerContainerTrait {
...
             assert ((EmailNotificationMessage) notificationMessages.get(0)).getText() == "Hello world!"
             assert ((EmailNotificationMessage) notificationMessages.get(0)).getSubject() == "Test"
             assert ((EmailNotificationMessage) notificationMessages.get(1)).getText() == "Hello world!"
             assert ((EmailNotificationMessage) notificationMessages.get(1)).getSubject() == "Test"
-            assert toAddresses.size() == 3
+            assert toAddresses.size() == 4
             assert toAddresses.any { it == "testuser2@openremote.local" }
             assert toAddresses.any { it == "testuser3@openremote.local" }
             assert toAddresses.any { it == "building@openremote.local" }
+            assert toAddresses.any { ((String) it).startsWith("service-account") } //mqtt service account
```
richturner, line 593: What is this?

manager/src/main/java/org/openremote/manager/mqtt/KeycloakAuthorizatorPolicy.java
```diff
@@ -0,0 +1,111 @@
...
+                    LOG.log(Level.INFO, "Couldn't verify token", verificationException);
+                    return false;
+                }
+            } else {
+                LOG.log(Level.INFO, "Couldn't verify token", e);
+                return false;
+            }
+        }
+    }
+}
```
richturner, line 111: The idea we discussed was to use the existing infrastructure, so this code should just create `EventSubscription`s and push through a route in `ClientEventService` that is used by Websocket and MQTT clients.

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -162,68 +173,92 @@ public void configure() throws Exception {
...
+                            .when(exchange -> isGatewayClientId(getClientId(exchange)))
+                                .to(GatewayService.GATEWAY_EVENT_TOPIC)
+                            .otherwise()
+                                .to(ClientEventService.CLIENT_EVENT_TOPIC)
+                        .endChoice()
+                    .otherwise()
+                        .process(exchange -> LOG.fine("Unsupported message body: " + exchange.getIn().getBody()))
+                    .end();
+
+                from(ClientEventService.CLIENT_EVENT_QUEUE)
```
richturner, line 205: This route is for outbound messages to `clients` shouldn't be `SUBSCRIBE` messages going in this direction....the only change to this route should be that after split() the `HEADER_CONNECTION_TYPE` should be checked and messages either routed to MQTT or Websocket...`splitForSubscribers` should be updated to add this header to each split message

richturner: Actually ok to use this route for inbound and outbound message processing can always adjust camel routes later if we need to but my previous comment still applies about `splitForSubscribers`

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -162,68 +173,92 @@ public void configure() throws Exception {
...
+                    .routeId("ToClientWebsocketEvents")
+                    .choice()
                     .when(bodyAs(String.class).startsWith(EventSubscription.SUBSCRIBE_MESSAGE_PREFIX))
                         .convertBodyTo(EventSubscription.class)
                         .process(exchange -> {
                             String sessionKey = getSessionKey(exchange);
                             EventSubscription subscription = exchange.getIn().getBody(EventSubscription.class);
                             AuthContext authContext = exchange.getIn().getHeader(Constants.AUTH_CONTEXT, AuthContext.class);
+                            sessionKeyConnectionTypeMap.put(sessionKey, (String) exchange.getIn().getHeader(HEADER_CONNECTION_TYPE));
```
richturner, line 214: Shouldn't be adding/removing sessionKeys for subsribe/unsubscribe messages...should only do this on connect/disconnect

richturner: A single client could have multiple subscriptions

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -142,6 +152,7 @@ public void configure() throws Exception {
                         .when(header(WebsocketConstants.SESSION_OPEN))
                             .process(exchange -> {
                                 String sessionKey = getSessionKey(exchange);
+                                sessionKeyConnectionTypeMap.put(sessionKey, HEADER_CONNECTION_TYPE_WEBSOCKET);
```
richturner, line 155: Move `SESSION_OPEN`, `SESSION_CLOSE` to `ClientEventService` and send message to `CLIENT_EVENT_QUEUE` with `HEADER_CONNECTION_TYPE` can then handle processing of client connection open/close in a generic way then

richturner: Sending a `CONNECTED` message back to the client might not make sense for `MQTT`, not sure it is that useful for `Websocket` either really but I would need to check if anything relies on it (don't think the new `core` component does)

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
richturner: Move `SUBSCRIBE`, `CANCEL`, `RENEW` logic into `CLIENT_EVENT_QUEUE` processing of these messages shouldn't matter where they come from

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -142,6 +152,7 @@ public void configure() throws Exception {
...
                                 sendToSession(sessionKey, "CONNECTED");
```
richturner, line 156: Delete this as not used anywhere and not very useful

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
richturner: `routeId` should be updated also as not just for `websocket` messages any more

test/src/test/groovy/org/openremote/test/notification/NotificationTest.groovy
Miggets7: Needed in order to make the test work. Because a Mqtt Keycloak client is added to the demo setup code. Keycloak adds an service user and it seems it also gets an email address.

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -209,22 +241,24 @@ public void configure() throws Exception {
                             }
                         })
                         .choice()
-                            .when(exchange -> isGatewayClientId(getClientId(exchange)))
-                                .to(GatewayService.GATEWAY_EVENT_TOPIC)
-                            .otherwise()
-                                .to(ClientEventService.CLIENT_EVENT_TOPIC)
+                            .when(header(Constants.AUTH_CONTEXT).isNotNull())
```
richturner, line 244: I think you should look for the `HEADER_CONNECTION_TYPE` because long term we might have anonymous access for public assets

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -162,68 +173,92 @@ public void configure() throws Exception {
...
-                    .end();
-
-                from(ClientEventService.CLIENT_EVENT_QUEUE)
-                    .routeId("ToClientWebsocketEvents")
-                    .choice()
                     .when(body().isInstanceOf(SharedEvent.class))
-                    .split(method(eventSubscriptions, "splitForSubscribers"))
-                    .to("websocket://" + WEBSOCKET_EVENTS)
+                        .split(method(eventSubscriptions, "splitForSubscribers"))
+                        .to("websocket://" + WEBSOCKET_EVENTS)
```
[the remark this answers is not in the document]
Miggets7: This already happens, maybe you looked at an older commit?

manager/src/main/java/org/openremote/manager/mqtt/AssetInterceptHandler.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Override
+    public void onDisconnect(InterceptDisconnectMessage interceptDisconnectMessage) {
+        MqttConnector.MqttConnection connection = mqttConnector.removeConnection(interceptDisconnectMessage.getClientID());
+        if (connection != null) {
+            CancelEventSubscription<AttributeEvent> cancelEventSubscription = new CancelEventSubscription<>(AttributeEvent.class);
+            messageBrokerService.getProducerTemplate().sendBodyAndHeader(ClientEventService.CLIENT_EVENT_QUEUE, cancelEventSubscription, WebsocketConstants.SESSION_KEY, connection.clientId);
+        }
+    }
```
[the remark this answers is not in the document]
Miggets7: This already happens, maybe you looked at an older commit?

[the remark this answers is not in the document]
richturner: That's an old comment, don't know why it has an updated time on it

PR #187, opened 2020-05-29
Input improvements

vlichtenberg:
What needs to be done still:

1. If a field is "dirty", it should change the helper text to "press enter to send". I've indicated in the code where I thought that should apply, but for some reason it doesn't.
2. Same goes for when the field is submitted: it should say "sending" and change back to the last updated state when the promise is resolved.
3. The way the field value is being submitted to the api feels a bit convoluted at this point. Is there a tighter approach to make the input field and button play together?
4. Don wants the time format of "last changed" to be in a specific style (eg. without seconds indication). I was wondering where I can set this, as options like LTS, LLL and LLLL being used in the translation file don't suffice for what Don wants.


ui/component/or-attribute-input/src/index.ts
```diff
@@ -133,19 +144,59 @@ export function getAttributeValueTemplate(
...
                     value = JSON.stringify(value, null, 2);
                 }
-                return Util.getAttributeValue(attribute, attributeDescriptor) ? Util.getAttributeValue(attribute, attributeDescriptor) : value;
+                currentValue = Util.getAttributeValue(attribute, attributeDescriptor) ? Util.getAttributeValue(attribute, attributeDescriptor) : value;
+                return currentValue;
+            };
+
+            // todo: using this fn to put a value into current value so the button knows about it is a bit of a detour but it works for now
+            const setHelperTextChanged = (event: any) => {
+                currentValue = event.path[0].value;
```
vlichtenberg, line 153: This feels like sorcery. I can't seem to find a way to fetch the current input value in another way though.

ui/component/or-attribute-input/src/index.ts
```diff
@@ -133,19 +144,59 @@ export function getAttributeValueTemplate(
...
+                helperText = i18next.t("sending");
+
                 if (inputType === InputType.JSON && value !== null && typeof(value) !== "string") {
                     if (typeof(value) === "string") {
                         value = JSON.parse(value);
                     }
                 }
+
+                const response = manager.rest.api.AssetResource.writeAttributeValue(assetId!, attribute.name!, value)
+                    .then(() => helperText = i18next.t("updatedWithDate", { date: new Date().toLocaleString() } as i18next.TOptions<i18next.InitOptions>))
```
vlichtenberg, line 168: This actually updates the helper text.

ui/component/or-attribute-input/src/index.ts
```diff
@@ -133,19 +144,59 @@ export function getAttributeValueTemplate(
...
+
+            // todo: using this fn to put a value into current value so the button knows about it is a bit of a detour but it works for now
+            const setHelperTextChanged = (event: any) => {
+                currentValue = event.path[0].value;
+                helperText = i18next.t("pressEnterToSend");
             };
 
             const setValue = (value: any) => {
+                iconTrailing = "send-clock";
+                helperText = i18next.t("sending");
```
vlichtenberg, line 159: This does not update the helper text for some reason. It does reach this point in the code though.

PR #191, opened 2020-06-12
Added support for IKEA Tradfri devices

RensvdLinden:
Added support for IKEA Tradfri devices


agent/build.gradle
```diff
@@ -28,6 +28,13 @@ dependencies {
     }
 
     compile "net.sf.saxon:Saxon-HE:$saxonHEVersion"
+
+    //IKEA Tradfri agent dependencies
+    compile "org.eclipse.californium:californium-core:2.1.0"
```
richturner, line 33: Any dep should have the version number placed in the `gradle.properties` file and referenced using avariable like everything else.

agent/build.gradle
```diff
@@ -28,6 +28,13 @@ dependencies {
...
+    compile "org.eclipse.californium:scandium:2.1.0"
+    compile "org.apache.commons:commons-lang3:3.9"
+    compile "com.fasterxml.jackson.core:jackson-databind:2.10.2"
```
richturner, line 36: Are we sure we need to explicitly declare these as they are dependencies from elsewhere in the codebase

agent/src/main/java/org/openremote/agent/protocol/tradfri/TradfriConnection.java
```diff
@@ -0,0 +1,133 @@
...
+import org.openremote.model.syslog.SyslogCategory;
+
+import java.util.function.Consumer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Logger;
+
+import static org.openremote.model.syslog.SyslogCategory.PROTOCOL;
+
+public class TradfriConnection {
```
richturner, line 21: Not totally clear what the difference between a connection and a gateway is...they look to be just the same thing

agent/src/main/java/org/openremote/agent/protocol/tradfri/TradfriAsset.java
```diff
@@ -0,0 +1,62 @@
...
+
+public abstract class TradfriAsset {
+    private String name;
+    protected Asset asset;
+    protected AssetAttribute protocolConfiguration;
+    protected MetaItem agentLink;
+    protected ProtocolAssetService assetService;
+    protected Device device;
+
+    public TradfriAsset(AssetAttribute protocolConfiguration, MetaItem agentLink, ProtocolAssetService assetService, Device device, AssetType assetType, HashMap<String, Device> tradfriDevices) {
```
richturner, line 21: Haven't dug into this but wondering why protocolConfiguration and assetService are being passed in here.
Why is this called an Asset....in OR we have an Asset class and this causes a bit of confusion.
What is the difference between TradfriAsset(s) and Tradfri device(s)

agent/src/main/java/org/openremote/agent/protocol/tradfri/device/DeviceObserver.java
```diff
@@ -0,0 +1,96 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.openremote.agent.protocol.tradfri.device.event.*;
+import org.openremote.agent.protocol.tradfri.payload.DeviceResponse;
+import org.openremote.agent.protocol.tradfri.util.CoapClient;
+
+import java.util.ArrayList;
+
+/**
+ * The class that observes a device to automagically detect changes
+ * @author Stijn Groenen
```
richturner, line 13: We don't put author and version info in the javadocs...authors can be added to contributors.txt

agent/src/main/java/org/openremote/agent/protocol/tradfri/device/event/Event.java
```diff
@@ -0,0 +1,19 @@
+package org.openremote.agent.protocol.tradfri.device.event;
+
+import org.openremote.agent.protocol.tradfri.device.Device;
+
+/**
+ * The class that represents an event that occurred to an IKEA TRÅDFRI device
+ * @author Stijn Groenen
+ * @version 1.0.0
+ */
+public class Event {
```
richturner, line 10: Empty class has no purpose and adds no value

agent/src/main/java/org/openremote/agent/protocol/tradfri/device/event/EventHandler.java
```diff
@@ -0,0 +1,34 @@
+package org.openremote.agent.protocol.tradfri.device.event;
+
+import java.lang.reflect.ParameterizedType;
+
+/**
+ * The class that handles events for IKEA TRÅDFRI devices
+ * @author Stijn Groenen
+ * @version 1.0.0
+ */
+public abstract class EventHandler<T extends Event> {
```
richturner, line 10: The generic use of `Event` offers nothing for the reason mentioned above regarding `Event` being empty

agent/src/main/java/org/openremote/agent/protocol/tradfri/payload/AuthenticateRequest.java
```diff
@@ -0,0 +1,45 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.openremote.agent.protocol.tradfri.util.ApiCode;
+
+/**
+ * The class that contains the payload for a request to authenticate to the IKEA TRÅDFRI gateway
+ * @author Stijn Groenen
+ * @version 1.0.0
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class AuthenticateRequest {
```
richturner, line 13: A class to contain a single `string` is a bit overkill

profile/tradfri.yml
```diff
@@ -0,0 +1,120 @@
+# OpenRemote v3
```
richturner, line 1: Think we discussed putting additional settings commented out in the `deploy.yml` profile with comments explaining what they are for so we can get rid of this profile file

agent/src/main/java/org/openremote/agent/protocol/tradfri/TradfriConnection.java
StijnGroenen: The Gateway class is the class that represents the IKEA TRÅDFRI gateway. This class is part of the IKEA TRÅDFRI library and is not dependent on OpenRemote classes.
The TradfriConnection contains all the methods and logic used by the OpenRemote protocol to manage a connection with the IKEA TRÅDFRI gateway (like the Consumers for the ConnectionStatus class).

agent/src/main/java/org/openremote/agent/protocol/tradfri/TradfriAsset.java
StijnGroenen: I changed it so TradfriAsset now extends the Asset class of OpenRemote.
The protocol configuration is no longer passed, but the asset service is used by the asset to merge new values (when updates occur).

agent/src/main/java/org/openremote/agent/protocol/tradfri/device/event/Event.java
StijnGroenen: The Event class has been removed.
The Event class was used as a parent for the DeviceEvent class and the GatewayEvent class. Though it did not provide any functionality, it was used as a shared parent which was used to limit the generic type of the EventHandler class.

agent/src/main/java/org/openremote/agent/protocol/tradfri/payload/AuthenticateRequest.java
StijnGroenen: This class represents the payload of a request to authenticate a new client to the IKEA TRÅDFRI gateway. This class is used by an object mapper to convert the property in this class to a valid JSON body for the authentication request.

richturner: Guessed it was to help `jackson` with serialisation....there are other ways to do this but this is quick and easy

PR #196, opened 2020-06-26
ArtNet protocol Bug Fix

DamienKusters:
We discovered a bug while testing OpenRemote with the protocol on a remote server.


RamonPeek:
^^ The bug occurred because we were reading hardcoded data for the Json-object which is used for the Value-asset. It always prompted to R,G,B,W only, even if other values were specified in the 'required values' property. We have completely removed the hardcoded constant, and replaced it with the dynamically loaded properties from the 'required values'.

richturner:
The `gradlew` file shouldn't be modified please make this PR only show the single file change to the protocol and then I can accept. Thanks

DamienKusters:
Quickly reverted the gradlew file 😉 
> The `gradlew` file shouldn't be modified please make this PR only show the single file change to the protocol and then I can accept. Thanks

richturner:
Great thanks for the quick fix

PR #249, opened 2020-09-01
Explicitly define healthchecks in docker-compose config

kklin:
Blimp (and Kubernetes) ignores healthchecks defined in `Dockerfile`s.
This commit adds them explicitly so that depends_on behaves as expected
with Blimp.

### DIFF
--- a/profile/deploy.yml
+++ b/profile/deploy.yml
@@ -87,6 +87,13 @@ services:
         options:
             max-size: 10m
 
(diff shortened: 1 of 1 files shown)


aktur:
Would be nice to have it merged in order to proceed with checking blimp.

richturner (approved):
Thanks for the contribution!

PR #460, opened 2021-07-20
Feature/role management

vlichtenberg:


richturner:
Regular user creation looks good.

Service user creation/update is broken though, specifically the Role assignment does not work

vlichtenberg:
I've just pushed a patch to fix the user creation (service and normal). It was trying to push something to an array that sometimes didn't exist.
I've clicked around on the page for a while and everything seems to work as expected now.

richturner:
Seems the same to me

Create a service user with assigned role(s) and save
A 404 is generated and role(s) are not shown when you expand the user
Update the service user roles and save and same problem again

richturner:
Some other feedback:

- Role dropdown should be called Roles now (as it is multiselect - should have changed before now but quick change incorporated here is fine)
- Role(s) dropdown shouldn't be required anymore

PR #464, opened 2021-08-12
Improvement/add asset dialog

DonWillems:
Improved flow for adding new asset. First select the type on the left, then set the name and parent (+optional attributes and later required attribute values) for that asset.

Unsure about the use of .mdc-list-group__subheader to style the headers in the asset type list..


ui/component/or-asset-tree/src/or-add-asset-dialog.ts
```diff
@@ -104,27 +116,67 @@ export class OrAddAssetDialog extends LitElement {
...
                 border-width: 1px 0;
             }
+
+            .msg {
+                display: flex;
+                justify-content: center;
+                align-items: center;
+                text-align: center;
+                height: 100%;
+                font-family: "Segoe UI", Helvetica,Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol";
```
vlichtenberg, line 131: Technically your message should be in a span, so the font family will come through by itself, but it's minor.

ui/component/or-asset-tree/src/or-add-asset-dialog.ts
```diff
@@ -211,32 +256,44 @@ export class OrAddAssetDialog extends LitElement {
...
+                </div>
+            </div>
             
             ${!attributes
                 ? html``
                 : html`
-                    <div style="margin-top: 0.5em">
-                        <strong>${i18next.t("attribute_plural")}</strong>
+                    <div>
+                        <div class="heading">${i18next.t("attribute_plural")}</div>
```
vlichtenberg, line 276: Actually while at it, I think it would be better to do h1/h2/h3/h4/etc, but I don't think they're well defined yet.

vlichtenberg (approved):
Left a few minor comments, but looks good!

PR #576, opened 2021-12-24
Feature/filter asset tree

jvervier:
Hello,

Feel free to check how I implemented this. I'm fully open to get your mind and maybe to see if I should more/less integrate the filter into or-mwc-input or not :-)


ui/component/or-asset-tree/src/index.ts
```diff
@@ -269,6 +271,13 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
     protected _expandedNodes: UiAssetTreeNode[] = [];
     protected _initCallback?: EventCallback;
 
+    @property({attribute: false})
+    protected _filterValue: string = '';
+    @property({type: Boolean})
```
richturner, line 276: For fields that should cause a re-render but are protected then the `@state` directive should be used rather than the `@property`...this was introduced in a more recent lit version so lots of our codebase needs updating but good to adopt this for new code:

https://lit.dev/docs/components/decorators/#built-in-decorators

ui/component/or-asset-tree/src/index.ts
```diff
@@ -327,17 +339,35 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
...
+                                this._onFilterValueChange(e);
+                              } }">
+                              </or-mwc-input>
+            </div>
+            
             ${!this._nodes
                 ? html`
                     <span id="loading"><or-translate value="loading"></or-translate></span>`
-                : html`
+                : (this._nodes.length === 0 && this._isFiltering
```
richturner, line 360: Can this be combined with the first check (i.e. `!this._nodes || this._isFiltering`)

ui/component/or-asset-tree/src/index.ts
```diff
@@ -538,6 +568,99 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
...
+        if (!this._filterValue.startsWith(oldFilterValue) || this._filterValue.length === 0) {
+            this._onCancelFilterChange();
+        }
+
+        if (this._searchInputTimer) {
+            clearTimeout(this._searchInputTimer);
+        }
+
+        this._searchInputTimer = setTimeout(async () => {
+            if (this._nodes && this._filterValue.length > 2) {
```
richturner, line 597: With async ops like this it's worth checking `isConnected` status of component on completion to avoid doing work if the component is no longer part of the DOM

ui/component/or-mwc-components/src/or-mwc-input.ts
```diff
@@ -125,7 +125,8 @@ export enum InputType {
     CRON = "cron",
     DURATION = "duration",
     DURATION_TIME = "duration-time",
-    DURATION_PERIOD = "duration-period"
+    DURATION_PERIOD = "duration-period",
+    FILTER = "filter"
```
richturner, line 129: Don't think we want `filter` as an input type; for background, `or-mwc-input` is intended to be a simple component that just applies material design styles to basic input fields.

Filter here is being used to apply some styling which should be achievable by just using `InputType#TEXT` and `css` from outside the `or-mwc-input` component.

ui/component/or-mwc-components/src/or-mwc-input.ts
```diff
@@ -665,6 +667,9 @@ export class OrMwcInput extends LitElement {
     @property({type: Boolean})
     public autoSelect?: boolean;
 
+    @property({type: Boolean})
+    public eventOnInput?: boolean;
+
```
richturner, line 672: The properties of this component generally replicate the standard HTML input attributes; for a particular instance of `or-mwc-input` you can just attach event handlers that you want:

```
<or-mwc-input @input=.....
```

ui/component/or-mwc-components/src/or-mwc-input.ts
```diff
@@ -1203,6 +1209,10 @@ export class OrMwcInput extends LitElement {
                     }
 
                     if (!(this.type === InputType.RANGE && this.disableSliderNumberInput)) {
+                        if (type === InputType.FILTER) {
+                            outlined = true;
+                        }
+
```
richturner, line 1215: Same as comment at the top about `FILTER` being used for styling...just use the property:

```
<or-mwc-input outlined ... />
```

ui/component/or-asset-tree/src/index.ts
```diff
@@ -305,8 +314,11 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
         return false;
     }
 
-    protected render() {
+    protected isFilteringAndNotMatching(assetId: string | undefined): boolean {
+        return assetId !== undefined && this._filterValue.length > 0 && this._foundAssets.length > 0 && this._foundAssets.indexOf(assetId) === -1;
+    }
```
richturner, line 319: Is this function necessary? Can this logic be stored as state in the `UiAssetTreeNode` when you iterate through the tree to apply filtering?

ui/component/or-asset-tree/src/index.ts
```diff
@@ -269,6 +271,13 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
     protected _expandedNodes: UiAssetTreeNode[] = [];
     protected _initCallback?: EventCallback;
 
+    @property({attribute: false})
+    protected _filterValue: string = '';
+    @property({type: Boolean})
+    protected _isFiltering: boolean = false;
+    protected _foundAssets: string[] = [];
+    protected _searchInputTimer: ReturnType<typeof setTimeout> | undefined = undefined;
```
richturner, line 279: `protected _searchInputTimer?;` saves having to explicitly state `| undefined` also happy to have type as `number`

ui/component/or-asset-tree/src/index.ts
```diff
@@ -538,6 +568,99 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
...
+                        predicateType: "string",
+                        match: AssetQueryMatch.CONTAINS,
+                        value: this._filterValue,
+                        caseSensitive: false
+                    }]
+                };
+
+                const response = await manager.rest.api.AssetResource.queryAssets(query);
+
+                this._foundAssets = response.data.map<string>((asset: Asset) => asset.id ? asset.id : '' );
```
richturner, line 618: If state becomes part of the `UiAssetTreeNode` as mentioned above then I don't see a need to store `_foundAssets`

ui/component/or-asset-tree/src/index.ts
```diff
@@ -538,6 +568,99 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
...
+
+        if (this._searchInputTimer) {
+            clearTimeout(this._searchInputTimer);
+        }
+
+        this._searchInputTimer = setTimeout(async () => {
+            if (this._nodes && this._filterValue.length > 2) {
+                let newNodes: UiAssetTreeNode[] = this._nodes;
+                this._nodes = [];
+                this._isFiltering = true;
```
richturner, line 600: As there doesn't seem to be a visual difference between loading and filtering the value of `_isFiltering` seems very limited. Previously when the component was loading `_nodes` was undefined so no need for additional boolean flag.

richturner: Also shouldn't the UI state change as soon as filter input occurs? Not 1.5s later?

richturner (changes requested):
Happy new year Jérôme! I've put some comments, happy to discuss anytime.

jvervier:
Hi @richturner, I'm back from holiday. I'll take a look on this on the next day :-)

ui/component/or-mwc-components/src/or-mwc-input.ts
jvervier: Hi @richturner , for this, using @input force me to leave to focus on the input field before any change has been "returned" to the asset-tree listening for input.
But if in or-mwc-input, I also add @input on inputElement (in addition to @change and with the same event handling), then it's working.

ui/component/or-mwc-components/src/or-mwc-input.ts
jvervier: I handle to get rid of this new type :-)

ui/component/or-asset-tree/src/index.ts
jvervier: indeed, I forget that point

ui/component/or-asset-tree/src/index.ts
jvervier: (y)

ui/component/or-asset-tree/src/index.ts
jvervier: Ok

ui/component/or-mwc-components/src/or-mwc-input.ts
jvervier: Removed as I removed the type FILTER

ui/component/or-asset-tree/src/index.ts
jvervier: Indeed, I include necessary info now in the recursive function and avoid the usage of an additional attribute in the class :-)

ui/component/or-asset-tree/src/index.ts
jvervier: In previous comment, I succeed to get rid of this :-)

ui/component/or-asset-tree/src/index.ts
jvervier: I put this in place as the request was to have a visual effect of the activity of filtering. 1.5s was discussed with Don but maybe we can decrease the value a bit? @DonWillems any comments?

jvervier:
Hi @richturner I put up to date the branch and also include your observations (please see comments for details)

ui/component/or-asset-tree/src/index.ts
jvervier: I migrate to standard number

jvervier:
Hi @richturner, I saw your point in your changes. You extracted more the change from mwc-input. Which is indeed the best compare to my initial idea to integrate things into mwc-input (with trailing icon action for example,...).

Do you want I renew a review on your changes?

richturner:
Always good if someone else can look over your work, happy for you to try out the updated code and look through changes and put any comments here; if it useful we can always have a call to discuss anything that isn't clear.

jvervier:
> Always good if someone else can look over your work, happy for you to try out the updated code and look through changes and put any comments here; if it useful we can always have a call to discuss anything that isn't clear.

Hello, I discover two things : 
- it was decided to make filtering only when at least 3 characters was entered
- I'm getting an expandable icon in front of asset with no children.

I'll bring a proposition for tomorrow

jvervier:
@richturner Here it is :-)

Note also, there is no more visual effect showing filtering is on going. As it was in the original request, should we put this back in the solution?

richturner:
> * it was decided to make filtering only when at least 3 characters was entered

@DonWillems @jvervier The 3 character requirement seemed a little strange, esp as I was un-aware of it I was trying to filter assets beginning with en but nothing happened. Was this requirement intended to reduce the number of matches?

richturner:
> * I'm getting an expandable icon in front of asset with no children.

Good spot I forgot to update that state during filtering

richturner:
> Note also, there is no more visual effect showing filtering is on going. As it was in the original request, should we put this back in the solution?

I think I mentioned that the 3s delay was detrimental and was removed; for slow connections then some visual indicator should be used but I didn't understand the need to simulate this with timeouts

DonWillems:
Since there is a short delay before searching when you are typing, I think it's fine to remove the 3 character requirement.

Don't know why I added it to be honest...

jvervier:
Ok :-) I'll so remove the 3 char check

jvervier:
Hi @richturner , I moved the matchers a bit in another way (to also prepare future features). Let me know if it's ok :-)

richturner:
Looks good!

richturner:
Squash and merge if finished

jvervier:
Hi @richturner, I can rebase the branch on the master but for the merge of the pull request, I do not have rights to do it :-)

richturner:
I've squash merged it now :-)

PR #578, opened 2021-12-30
Feature/map pins

vlichtenberg:
~Just to show Don a diff of what has been done so far~


ui/component/core/src/index.ts
```diff
@@ -109,6 +109,34 @@ export enum MapType {
     RASTER = "RASTER"
 }
 
+//todo: fix the ts problem
+export interface AttributeMarkerColours {
+    type: "string" | "boolean";
+    [value: string]: string;
+    //@ts-ignore
```
vlichtenberg, line 116: @DonWillems wished for labels to be toggled, but obviously we can't do it this way because of line 115 in this file.
Was in a rush now so couldn't really find a solution to prep the model for this feature.

ui/app/manager/src/pages/page-map.ts
```diff
@@ -99,7 +101,40 @@ export function pageMapProvider(store: EnhancedStore<MapStateKeyed>, config?: Pa
         ],
         pageCreator: () => {
             const page = new PageMap(store);
-            if(config) page.config = config;
+
+            page.config = config || {};
+
+            // todo: remove stub
```
vlichtenberg, line 107: Should get rid of this as it should come through the manager from a config file set by the user.

ui/component/or-map/src/markers/or-map-marker-asset.ts
```diff
@@ -1,28 +1,18 @@
 import {PropertyValues} from "lit";
 import {customElement, property} from "lit/decorators.js";
 import {OrMapMarker} from "./or-map-marker";
-import {AttributeEvent, GeoJSONPoint, AssetEvent, AssetEventCause, Asset, SharedEvent, AssetDescriptor, WellknownAttributes} from "@openremote/model";
```
richturner, line 4: Re-formatting mixed in with code changes is a big headache to figure out what are changes and what's formatting. We need to put some standard formatting rules in place but for now please don't re-format whole files at the same time as making change.

ui/component/core/src/index.ts
```diff
@@ -109,6 +109,37 @@ export enum MapType {
     RASTER = "RASTER"
 }
 
+export type BaseMapMarkerConfig = {
```
richturner, line 112: Doesn't seem like my previous comment has been implemented. Don't want this config in the core component it needs to be part of the `ManagerAppConfig`

richturner (approved):
Looks good

PR #617, opened 2022-02-25
Feature/panel linked users

vlichtenberg:


manager/src/web/shared/locales/cn/or.json
```diff
@@ -435,5 +435,7 @@
     "invalidJSON": "Invalid JSON",
     "invalidDate": "Invalid date",
     "keyAlreadyExists": "Key already exists"
-  }
+  },
+  "yes": "是的",
```
vlichtenberg, line 439: If only we had a Chinese speaker in our midst... oh wait!

vlichtenberg:
Just found out I'm missing the following logic still:
`Only show panel if the user is allowed to read user data (has write:admin or read:admin or is superuser), unless the user without read:users permission is the one linked to the asset. Then show the panel with only themself there.`
I'll try to fit this in before this Friday.

vlichtenberg:
I've pushed an update, checks whether user is superuser, has write:admin or read:admin. It does indeed produce a console error on the roles request (403), most likely because user doesn't have read:users permission. I'll have to catch that as well and try to put in logic to produce the panel after all when they do.

DonWillems:
It looks like this for me. Also known to happen in the group panel.
image

vlichtenberg:
> It looks like this for me. Also known to happen in the group panel. image

It doesn't do that for me on Chrome. Are you using something else?
I had this happen 2 weeks ago too but then it hasn't bothered me since then. Hmm.

vlichtenberg:
- "fixed" the minimum height of the table
- Discussed the requirement of showing/hiding the panel based on the user having restricted access regardless of their roles. It doesn't make a lot of sense at this point so we're leaving it out for now.

richturner (approved):

PR #648, opened 2022-03-21
Feature/asset tree filter improvement

jvervier:
Hi @richturner ,

Things move ! I'll mail you for one specific details :-)


jvervier:
Hi @richturner ,

I made a proposal for the input text field issue with the absolute clear-icon which is a trailing icon with action outside the or-mwc-input.

Everything is ready to be checked

regards

ui/component/or-asset-tree/src/index.ts
```diff
@@ -346,27 +434,102 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
                               outlined="true"
                               @input="${(e: KeyboardEvent) => {
                                   // Means some input is occurring so delay filter
-                                  this._onFilterInput(e);                                  
+                                  this._onFilterInputEven(e);
```
richturner, line 437: Typo

ui/component/or-asset-tree/src/index.ts
```diff
@@ -575,30 +738,215 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
...
+    protected _onFilterInput(newValue: string | undefined, force: boolean): void {
+        if (newValue) {
+            this._clearIconContainer.classList.add("visible");
+        } else {
+            this._clearIconContainer.classList.remove("visible");
+        }
+
+        let currentFilter: OrAssetTreeFilter = this.parseFromInputFilter(newValue);
+
+        if (this._filter.asset === currentFilter.asset &&
```
richturner, line 927: There is a util helper method for object deepEquals checking

ui/component/or-asset-tree/src/index.ts
jvervier: Fixed

ui/component/or-asset-tree/src/index.ts
jvervier: You're right, I moved to Util.equals

jvervier:
Hi @richturner , Just get your comments into the code. Pushed on the branch

ui/component/model/src/index.ts
```diff
@@ -1,2 +1,2521 @@
-export * from "./model";
-export {AssetModelUtil} from "./util";
```
richturner, line 2: Not sure how this has happened but the autogenerated contents get output to `model.ts` but they seem to have been copied into here?

richturner (changes requested):

ui/component/or-asset-tree/src/index.ts
```diff
@@ -649,42 +998,145 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
...
+            attributes: attributeCond
         };
 
         const response = await manager.rest.api.AssetResource.queryAssets(query);
+
         const foundAssetIds: string[] = response.data.map((asset: Asset) => asset.id!);
 
         return (asset) => {
-            return foundAssetIds.includes(asset.id!);
+            let attrValueCheck = true;
```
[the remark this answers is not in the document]
jvervier: My bad for that.

I'll implement the comparaison on assetQuery side

ui/component/model/src/index.ts
jvervier: MMMh I do not intend myself to copy this content. Don't really know what could have create such case.

ui/component/or-asset-tree/src/index.ts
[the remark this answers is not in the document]
jvervier: @richturner The only things is : using the ValuePredicateUnion need to know the attribute type. By doing it after the call, we know the attribute data type and can handle the right comparaison.
If we need to know the Attribute type before executing the request we need to get the attribute type (linked to the attribute name provided in the search) by another way (or another request).

For that part I prefer to get your final word on this. If you prefer we can schedule a discussion about that :-)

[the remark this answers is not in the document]
jvervier: Thinking about this, I could also use the content on the right of the filter attribute value "attribute:valueComparaison" and test the type of content to make assumptions we're comparing a number attribute value,... in order to use the right Type in the ValuePredicateUnion.

ui/component/model/src/index.ts
jvervier: I don't really know which action lead to that change but I fix this by reverting the changes

richturner (approved):

PR #666, opened 2022-04-01
send input changed event on keyup as an input's value isn't updated y…

vlichtenberg:
…et on keydown

### DIFF
--- a/ui/component/or-mwc-components/src/or-mwc-input.ts
+++ b/ui/component/or-mwc-components/src/or-mwc-input.ts
@@ -1278,6 +1278,10 @@ export class OrMwcInput extends LitElement {
                                 if ((e.code === "Enter" || e.code === "NumpadEnter")) {
                                     this.onValueChange((e.target as HTMLInputElement), (e.target as HTMLInputElement).value, true);
                                 }}}"
+                            @keyup="${(e: KeyboardEvent) => {
+                                if(e.code === "Enter" || e.code === "NumpadEnter") return false;
+                                this.onValueChange((e.target as HTMLInputElement), (e.target as HTMLInputElement).value, false);
+                            }}"
                             @blur="${(e: Event) => {if ((e.target as HTMLInputElement).value === "") this.reportValidity()}}"
                             @change="${(e: Event) => this.onValueChange((e.target as HTMLInputElement), (e.target as HTMLInputElement).value)}" />`;
 


richturner (changes requested):
This has a lot of implications. or-input-changed event is used to indicate when attribute inputs should send attribute events to the backend and this should not happen on every keypress

ui/component/or-asset-viewer/src/index.ts
```diff
@@ -1038,6 +1040,7 @@ export class OrAssetViewer extends subscribe(manager)(translate(i18next)(LitElem
         this.addEventListener(OrChartEvent.NAME, () => OrAssetViewer.generateGrid(this.shadowRoot));
         this.addEventListener(OrAttributeHistoryEvent.NAME, () => OrAssetViewer.generateGrid(this.shadowRoot));
         this.addEventListener(OrEditAssetModifiedEvent.NAME, (ev: OrEditAssetModifiedEvent) => this._onAssetModified(ev.detail));
+        this.addEventListener('keyup', () => this._assetModified = true)
```
vlichtenberg, line 1043: It's really just about this. It does listen to any keyup happening in the assetviewer. I tried with change but there's no getting out of blurring (defeating the purpose of this ticket) and a cleaner keypress event is unusable because it doesn't include backspace.
Could scope this a bit better by putting custom events on all the or-attribute-input instances in the panel, but that wouldn't include other fields on the page. Thought this would be better in general.

ui/component/or-asset-viewer/src/or-edit-asset-panel.ts
```diff
@@ -298,7 +298,11 @@ export class OrEditAssetPanel extends LitElement {
                 <td class="padded-cell mdc-data-table__cell expander-cell"><or-icon icon="chevron-right"></or-icon><span>${attribute.name}</span></td>
                 <td class="padded-cell mdc-data-table__cell">${Util.getValueDescriptorLabel(attribute.type!)}</td>
                 <td class="padded-cell overflow-visible mdc-data-table__cell">
-                    <or-attribute-input ${ref(attributeInputRef)} compact .comfortable="${true}" .assetType="${assetType}" .label=${null} .readonly="${false}" .attribute="${attribute}" .assetId="${this.asset.id!}" disableWrite disableSubscribe disableButton @or-attribute-input-changed="${(e: OrAttributeInputChangedEvent) => this._onAttributeModified(attribute, e.detail.value)}"></or-attribute-input>
```
vlichtenberg, line 301: I was editing this, but ultimately ended up doing nothing to it. Thought I'd leave it in a more readable state.

richturner (approved):

Issue #302, opened 2021-02-01
FE: list of styling updates


DonWillems:
- [x] Asset-viewer header sticky with shadow on scroll.
- [ ] Header: add dividers
- [x] service user creation: remove/change regenerate secret section as it will be auto generated after creation.
- [x] Assets: Linked users panel: If the user does not have a first and last name, the username isn't shown at all.
- [x] Keep uncollapsed state of asset tree in memory so when returning to the Assets page it looks the same (correct for potential changes in assets)
- [x] Insights: Chart: 'Add' button should not be clickable of no attribute is selected.
- [x] 'undefined' console log on insights page when adding attributes to chart or card. Check other places too.
- [x] Add configuration item dialog on modify asset page has padding causing a horizontal slider. Remove.
- [x] 'Add another condition' text in When-Then rules should be clickable as is the '+'
- [ ] Give dialogs the appropriate dialogHostElement so that they fit correctly in the element of the page they are relevant to.
- [x] Assets: when navigating from the map (view asset) to the asset page, the selected asset should not only be shown as 'selected' in the tree, but that asset should also be shown within view, without requiring to scroll down in a long list of assets
- [x] History data table should show latest timestamp at top
- [ ] Rules: Asset type list should be split in category "Used in project" and "Other". New endpoint needed for this?// Should be solved when only adding relevant asset types to a project?
- [ ] or-map: if 'center on my location' is used and user location does not exist in mapsettings bounds, show error message "location not within map bounds" as 'snackbar'
- [x] Replace logo and logo-mobile with smaller version. (in demo folder)
- [x] Off/On label of light switch is wrong way around
- [x] Asset: Edit: Parent edit button should be outlined not primary
- [x] Add 20px space at bottom of Asset viewer grid
- [x] Add unit to attributes in tooltips of chart
- [x] Attribute selection dialog (e.g. in Insights charts) should have the attributes in alphabetical order (use util)
- [x] Add titles to icon buttons throughout interface
- [x] Assets: Edit mode: Add config item modal has double borders for content.
- [x] Assets: Add attribute: Two selects of 'Custom attribute' should be hidden, so no height reserved.
- [x] Asset tree: Group/parent collapses when selected
- [x] Default language in manager config is not applied.
- [x] Show which language is currently selected
- [x] Manager interconnect: page title styling mobile
- [x] Max height of select lists. Example: add custom attribute, valueType
- [x] Expand add attribute modal in edit mode only when the two extra fields are needed.
[...]

PR #688, opened 2022-04-22
Bugfix/302 fe list of styling updates

vlichtenberg:
Solves a few issues in #302


DonWillems:
Everything is in https://github.com/openremote/openremote/pull/692

PR #708, opened 2022-05-16
Update cloudformation-create-ec2.yml

aktur:


aktur:
obsolete with the next

PR #711, opened 2022-05-18
Update cloudformation-create-ec2.yml

aktur:
On request of Pierre added alarm and recovery for instance. There are still Route 53 healthchecks to be added. Due to not be able to push to master and test it on the AWS stack, I've developed the code outside on different AWS account. PR -> merge -> deploy -> test loop would take too much time which I don't have from Pierre.


richturner:
Does this replace #708 as I see the same changes but this PR is more extensive? Should just be done in the one branch as it's related.

As for testing why can't you just manually create a host using this cloudformation template?

If you wanted to get more accurate CI/CD test scenario then you can add the github-cli user credentials to AWS CLI locally and call `.ci_cd/aws/provision_host.sh` just like the github `Privision` workflow does or use the github UI and set `Use workflow from` to your branch.

richturner (approved):

PR #715, opened 2022-05-21
Added HealthCheck

aktur:
On Pierre's request a HealthCheck with corresponding Alarm.


richturner:
I've added the following permission to the `github-cli` user in the `standard-resources` Cloud formation stack of the developers member account otherwise github CI/CD pipeline won't be able to create EC2 instances using this template:
```
- route53:CreateHealthCheck
```

PR #727, opened 2022-06-03
Styling of asset tree filtering settings

DonWillems:
- Some styling changes to the filter settings panel
- Remove icon and clear button, add placeholder text. To make a bit more space for input, and because the clear button was overlapping the text.
- Add clear button to the settings panel with the functionality the clear icon had before.

Should _onFilterInputEvent(e: KeyboardEvent) in or-asset-tree index be removed altogether?

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -55,12 +55,15 @@
 name: CI/CD
 
 on:
-  # Push excluding tags and workflow changes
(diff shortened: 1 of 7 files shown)


jvervier (approved):
Hi @DonWillems , 

Ok for me, but I don't see reason to remove _onFilterInputEvent(e: KeyboardEvent).

Issue #714, opened 2022-05-20
Can't launch from 8443 port


DanielBustillos:
**Describe the bug**
Following the instructions from the Quickstart. When I try to launch the by:

`OR_HOSTNAME=ip OR_SSL_PORT=8443 docker-compose -p openremote up -d`

I can't access to the site via the ip. But I changed the port to 443 and now it works. I use AWS EC2 and I have both ports opened. Changing the port in the Quickstart should facilitate the installation.

**To Reproduce**
1. Follow the Quickstart from the readme and run the suggested command to launch from an specific host:
`OR_HOSTNAME=192.168.1.1 OR_SSL_PORT=8443 docker-compose -p openremote up -d`
3. Access to the ip in a web browser adding https: https://192.168.1.1
4. No access to the site

**Desktop (please complete the following information):**
 - OS: ubuntu 20.04 LTS,  AWS EC2 instance
 - Safari

richturner:
Did you follow the instruction underneath in bold:

**NOTE: When chaning OR_SSL_PORT you will need to alter the KEYCLOAK_FRONTEND_URL in the `docker-compose.yml` file (see comments in the file for details)**

richturner:
There may well be an issue with Letsencrypt SSL cert generation when using a non standard port as this functionality is generally in place for running on a local network with self signed or existing SSL cert.

DanielBustillos:
> Did you follow the instruction underneath in bold:
> 
> **NOTE: When chaning OR_SSL_PORT you will need to alter the KEYCLOAK_FRONTEND_URL in the `docker-compose.yml` file (see comments in the file for details)**

Yes, but why not to include the 443 port in the instructions?  No extra config required in the `docker-compose.yml`

richturner:
Because keycloak doesn't recognise these as the same origin:

```
https://mydomain.com:443/
https://mydomain.com/
```

DanielBustillos:
The instructions to launch should include port 443 no 8443. If you follow the instructions to the letter, you will not be able to access the launched site.

richturner:
Hi, where in the instructions do you see `8443` mentioned?

DanielBustillos:
Thank you for your response Rich, 8443 is mentioned in the startup command of the instructions. I've open a pr  #741   with a minor change in the startup command.

richturner:
The section of the README you have updated is called  `Changing host and/or port` and it uses 8443 as an alternative port suggestion.

[...]

PR #741, opened 2022-06-16
Minor change in startup command

DanielBustillos:
Changed to port 443 so openremote can be launched without extra configuration.

#714


richturner:
Don't think this PR makes sense.

PR #743, opened 2022-06-17
667 multi select assets and move them in tree

jvervier:
Hello @richturner ,

Here is my changes to handle multi select, changing parent on multi select and also the drag&drop in the asset tree

Regards,

### DIFF
--- a/model/src/main/java/org/openremote/model/asset/AssetResource.java
+++ b/model/src/main/java/org/openremote/model/asset/AssetResource.java
@@ -35,6 +35,8 @@ import org.openremote.model.value.MetaItemType;
 
 import javax.annotation.security.RolesAllowed;
 import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
 import javax.ws.rs.*;
 import javax.ws.rs.core.Response;
 import java.util.List;
@@ -277,4 +279,21 @@ public interface AssetResource {
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
     Asset<?>[] queryAssets(@BeanParam RequestParams requestParams, AssetQuery query);
+
+    /**
+     * Change parent for a set of asset
+     */
+    @PUT
+    @Path("{parentAssetId}/child")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    void updateParent(@BeanParam RequestParams requestParams, @PathParam("parentAssetId") @NotNull(message = "Parent reference required") String parentId, @QueryParam("assetIds") @Size(min = 1, message = "At least one child to update parent reference") List<String> assetIds);
+
+    /**
+     * Remove parent reference from each asset referenced in the query parameter assetIds
+     */
+    @DELETE
+    @Path("/parent")
+    @Produces(APPLICATION_JSON)
+    void updateNoneParent(@BeanParam RequestParams requestParams, @QueryParam("assetIds") @Size(min = 1, message = "At least one child to update parent reference") List<String> assetIds);
 }
--- a/ui/component/or-asset-tree/package.json
+++ b/ui/component/or-asset-tree/package.json
@@ -19,6 +19,7 @@
     "@mdi/js": "^5.9.55",
     "@openremote/core": "workspace:*",
     "@openremote/model": "workspace:*",
+    "@openremote/or-asset-viewer": "^1.1.0",
     "@openremote/or-icon": "^1.1.0",
     "@openremote/or-mwc-components": "^1.1.0",
     "@openremote/or-translate": "^1.1.0",
--- a/ui/component/or-asset-viewer/src/index.ts
+++ b/ui/component/or-asset-viewer/src/index.ts
@@ -223,15 +223,18 @@ export class OrAssetViewerSaveEvent extends CustomEvent<SaveResult> {
     }
 }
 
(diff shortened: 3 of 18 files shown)


DonWillems:
Hi @jvervier, two issues I noticed:
1. When dragging selected item(s) over other assets, the dark gray background flickers a bit, this should be a constant colour.
Reproduce: select an asset and (slowly) drag it over the assets in the asset tree. You will notice a slight flicker. This could be a bit more polished and have a constant colour while moving over it.
2. [Minor] Changing parent of multiple assets to one of those selected assets, causes them to move until reaching the item moved too and showing the snackbar. Happens both for parent select dialog and drag & drop.
Reproduce: Select 5 assets on the same level > drag to the 3rd selected item and release > the first two items will have moved under the 3rd item, an error message will pop up, and the 4th and 5th items will stay in their original position.
Expected: none of the items move because there is an error in the bulk action.

jvervier:
Hi @DonWillems ,

I pushed changes to fix issues I look on today. The only remaining is the blinking (I'm still looking) and the red 'impossible' icon that I can't reproduce (even with brave on Mac). I'll proceed with tests on a windows station.

jvervier:
Hi @richturner ,

After discussing with @DonWillems, we can move forward now with the review of this PR as it's OK in term of functionality.

Let me know if there is any issue in the code :-)

model/src/main/java/org/openremote/model/asset/AssetResource.java
```diff
@@ -277,4 +279,21 @@ void deleteAllUserAssetLinks(@BeanParam RequestParams requestParams,
...
+    void updateParent(@BeanParam RequestParams requestParams, @PathParam("parentAssetId") @NotNull(message = "Parent reference required") String parentId, @QueryParam("assetIds") @Size(min = 1, message = "At least one child to update parent reference") List<String> assetIds);
+
+    /**
+     * Remove parent reference from each asset referenced in the query parameter assetIds
+     */
+    @DELETE
+    @Path("/parent")
+    @Produces(APPLICATION_JSON)
+    void updateNoneParent(@BeanParam RequestParams requestParams, @QueryParam("assetIds") @Size(min = 1, message = "At least one child to update parent reference") List<String> assetIds);
 }
```
richturner, line 299: Is there a reason for 2 methods rather than one?

ui/component/or-asset-tree/package.json
```diff
@@ -19,6 +19,7 @@
     "@mdi/js": "^5.9.55",
     "@openremote/core": "workspace:*",
     "@openremote/model": "workspace:*",
+    "@openremote/or-asset-viewer": "^1.1.0",
```
richturner, line 22: Don't see any imports from this added package? Would also need to be added to tsconfig.json if there was a dependency

ui/component/or-asset-viewer/src/index.ts
```diff
@@ -221,6 +223,22 @@ export class OrAssetViewerSaveEvent extends CustomEvent<SaveResult> {
     }
 }
 
+export class OrAssetViewerChangeParentEvent extends CustomEvent<any> {
```
richturner, line 226: Detail of the custom event shouldn't be any but `{parentId: string, assetIds: string[]}`

ui/component/or-asset-viewer/src/index.ts
```diff
@@ -1105,6 +1124,9 @@ export class OrAssetViewer extends subscribe(manager)(translate(i18next)(LitElem
     @property({type: String})
     public assetId?: string;
 
+    @property({type: Array})
+    public assetsIds: string[] = [];
+
```
richturner, line 1129: Why do we need two properties for selected assets?

richturner (commented):
See my comments

model/src/main/java/org/openremote/model/asset/AssetResource.java
jvervier: first is to attach to one parent new children.
second is the case where asset are put on root level of the asset tree (and so no parent and so no parent asset ID provided). I tried to avoid to put "undefined" or any strange value on the first PUT.

ui/component/or-asset-tree/package.json
jvervier: I forget to clean this ! It'll be cleared

ui/component/or-asset-viewer/src/index.ts
jvervier: right

ui/component/or-asset-viewer/src/index.ts
jvervier: Asset viewer is only containing info for one asset (to show it), here it's to handle the case when more than one asset is selected in the asset tree.

Of another way should be to create an OrMultipleAssetsViewer component only dedicated to show the button to change parent for multiple selected asset in the tree.

jvervier:
Fix little issues and put comment for the others

richturner (approved):

ui/component/or-asset-viewer/src/index.ts
richturner: The one property `assetIds` should be sufficient and a single item in the array indicates single asset selection

richturner: I need to checkout the code and see what happens to the asset viewer during multiselect as not clear to me why the asset viewer needs to know anything.

richturner (approved):

PR #750, opened 2022-06-27
Feature/realm role input

robin-dekkers:


DonWillems (changes requested):
Looks good. Only remarks directly related to this PR would be that the text in the button should be '# Assets' instead of 'linked assets'. It does not show the selected assets count anymore, which it should. ({{number}} parameter in translations)

Remarks sort of related, can be picked up in separate PR if they do not fit:
- Now that we have 'Realm roles', maybe the other roles should be named 'Manager roles'
- There is no space after the comma of each selected role (", ") in the multiselect field, which would be nicer.
- The permissions (under the roles select) are alphabetically horizontal instead of vertical. Vertical they would be nicely split in read: and write:...

DonWillems (dismissed):

ui/app/manager/src/pages/page-users.ts
```diff
@@ -804,6 +809,22 @@ export class PageUsers extends Page<AppStateKeyed> {
...
+                                    <or-mwc-input
+                                            ?readonly="${readonly}"
+                                            ?disabled="${isSameUser}"
+                                            .value="${user.realmRoles && user.realmRoles.length > 0 ? user.realmRoles.filter(r => !this._realmRolesToExclude.includes(r.name)).filter(r =>!r.composite).map(r => r.name) : undefined}"
+                                            .type="${InputType.SELECT}" multiple
+                                            .options="${realmRoleOptions}"
+                                            .label="${i18next.t("realm_role_plural")}"
+                                            @or-mwc-input-changed="${(e: OrInputChangedEvent) => {
+                                                const roleNames = e.detail.value as string[];
+                                                user.realmRoles = this._realmRoles.filter(cr => roleNames.some(name => cr.name === name)).map(r => {
```
richturner, line 822: This looks like it will remove the excluded roles from the user which it should not @DonWillems do I need to resolve this ready for tomorrow?

richturner (dismissed):

ui/app/manager/src/pages/page-users.ts
richturner: Resolved now

richturner (approved):

PR #751, opened 2022-06-27
UI updates: Asset viewer enhancements

richturner:


DonWillems (changes requested):
Looks good, few remarks:
- History: Attribute selection: I see what the idea was of moving the attribute selection as some sort of higher level control over the panel, but I don't think introducing that placement/interaction in the history graph makes sense for now. I think it's better to keep the original layout (with the change that it is initially empty until you select the attribute). I will practice some css to get it a bit nicer looking soon (something like images below). And then I will finally spent some time to come with a real solution.
Wide screen:
proposalSelects
A bit less wide screen..:
proposalSelects2

- The top and bottom of the panels container has an additional 10px that shouldn't be there.
10pxTooMuchTopBottom

- The config order does not translate to a single column layout that ignores the column number. e.g. suppose a config with panels with colums 0 0 1 0 1 0, becomes 000011 in the single column layout instead of 001010. I assume this is because they are separated into 2 column containers and you would need a re-render to order the panels according to the config? If so, and it is not really workable, keep it as is (meaning, it is fine if the second column just goes below the first one)
- The page is not full height, so if a select field or menu is near the bottom of the last panel they get cut off:
pageNotFullHeight

PR #760, opened 2022-07-10
702 non existing asset type warning

jvervier:
Hi @DonWillems and @richturner ,

Here are my fix for improvement on asset tree filter.

I'm off next week but able to pass to check remarks/fix to provide ;-)


DonWillems (approved):
Looks good. Approved, just two notes:
1. I've made a commit fixing the color of the asset type icon that is selected in the options panel, and removed a class set on the label so that it aligns with the input. Please check if I've done that correctly.
2. I found that unlike the asset and attribute name, the AssetType needs to match the case. I see that this is not something currently available in the AssetQuery, so lets leave it for now and I will add it to the backlog.

richturner (approved):

PR #763, opened 2022-07-14
Remove `btmesh-data` unused `volume`

PululuK:
The **btmesh-data** volume is no longer used at all so,  we can remove.


richturner:
I have already made this adjustment whilst also updating other parts of the file so nothing to do here. Thanks again

PR #779, opened 2022-08-01
Several small UI updates/fixes

DonWillems:
@richturner if you find the time, what do you think of this change? If this is incorrect, I can just revert that part for now. Others can review the other changes.
- Asset viewer: Modify: Replaced 'AnyType' reference in schema of SimulatorReplayDatapoint as it gave errors and did not show in the data fields in the vertical data viewer.

Changes:
- Asset tree: Added icon toggle to close the filter options
- Asset viewer: Fixed console error when selecting multiple assets
- Asset viewer: Made selected file input fullwidth and with label
- Asset viewer: Made textarea attribute inputs resizable vertically
- Asset viewer: Removed 10px at top and bottom of columns in asset viewer
- Asset viewer: Modify: Set 'custom attribute' as default selected when adding a new attribute
- Asset viewer: Modify: Made multi-select fields fixed width so that they not overflow the page
- Asset viewer: Modify: Made parent field read-only
- Asset viewer: Modify: Added border and title to parent select dialog


MartinaeyNL (commented):
Looks fine to me in general

MartinaeyNL (approved):

Issue #697, opened 2022-05-04
Assets: Linked users: service users username not shown


DonWillems:
Reproduce:
- Log in to a project where assets are created by provisioning. Or link a service user to an asset manually
- Go to the asset page to view the linked users panel
- See the linked user, but no username (because a service user does not have a first and last name?)

Expected:
- See the service user username in the table.

PR #783, opened 2022-08-05
Minor UI features, styling updates and bugfixes

MartinaeyNL:
Did updates and fixes from for example #302, or ones that @DonWillems suggested.

Changelog:
> [FEATURE]  Service users on Users page now include a "copy secret" action button.
> [FEATURE]  Expanded parents in asset tree are now saved in Redux store, to keep them opened/closed after page switching.*
> [FEATURE]  Added (currently unused) autoScroll on Users page, for scrolling towards the table entry after user creation.

> [BUGFIX]  Linked user on asset page now shows firstName/lastName correctly (#697)
> [BUGFIX]  Small fix on deletion of users; it does not open the details of other users anymore.

> [STYLING]  Added header shadow on asset-viewer when scrolling.
> [STYLING]  Replaced the empty secret field on service user creation to text; "will be generated on creation"
> [STYLING]  Username field on Users page is now disabled after creation.
> [STYLING]  Now showing snackbar on user save/update.

*I'm not sure whether using Redux store for this is a good practice, so I would appreciate feedback on this.
Especially if we want to include states of other pages in the future.


MartinaeyNL:
Currently a draft, doing last bugfixes before it's ready for review.

MartinaeyNL:
After discussion with Rich I will reconsider the 'expanded parents/selected asset on page switching',
since it's not practical to save it in the current practice used. Especially since the asset ID is saved in the URL, and deleted on page switch.

I will also reconsider the feature together with Don; why would a menu button serve as "go back to where you were" option.

DonWillems:
Regarding the 'expanded parents/selected asset on page switching'. I think it is a nice addition:

In use when going from the asset page, to e.g. rules, to map, back to asset page, it was annoying to find the tree fully collapsed again every time. So the request wasn't for the same asset to be open again (which isn't implemented) but for the tree to look familiar to me/reduce time clicking the structure open again. 
Using the back button results in the tree to be open up to the selected asset, but all other uncollapsed assets and/or children of the asset that was selected before, will not be in its previous uncollapsed state. 
With this addition, also when using the back button the experience has improved I think.

@MartinaeyNL 
The snackbar on service user creation with the 'Copy secret' button. The copied value is 'undefined', not the generated secret.

MartinaeyNL:
Ready for merge

richturner (approved):

PR #807, opened 2022-09-15
Bugfix/rules issue

jvervier:
Hi @richturner & @DonWillems ,

As this issue is quite blocking the release, I propose to first apply this solution and propose in a second time another one with a better separation between Asset and Ruleset (by making de deserialise different for Ruleset).


richturner:
I have done the re-factor we discussed to get this completed and to avoid polluting the asset meta item types.

I also introduced coverage for this in tests.

As a side note: Several of the changed files could be un-done (un-used imports and spacing changes).

richturner (approved):

richturner (approved):

PR #820, opened 2022-10-03
Fix language bug, styling update parameter dialog

DonWillems:
- Fix bug where language set in manager_config is overwriting local storage on every refresh (and is setting the language of each configured realm on each refresh)
- Add title/styling to parameter dialog for agent link. (Unrelated to the bug though...)


richturner (approved):
Good work finding the cause

PR #848, opened 2022-10-25
Feature/improve asset tree dom size

jvervier:
Chrome based browser is consuming more time to handle "Recalculating style" compared to other browser. In our test case, one asset has 5000 children assets.

The idea of this proposal is to reduce the quantity of asset attached to the html document by only attaching asset in the document html code when they're effectively visible on the screen (showing only what's expanded).

I also fix a little undefined issue (adding some tests in if).

I moved Martin's code into the _treeNodeTemplate method.

This change in the asset tree size is solving asset page loading issue with Chrome based browser (time consumption as the "recalculating time" is taking time).

This solution is a first part as, when we'll expand a asset with a HUGH number of asset, the asset page loading issue will consume time. Don is actually thinking about some other possibility to improve user experience.

Next step could be to consider a way, when more than a given number of asset are visible, to paging the html content depending of what the asset tree height permit to see a number of asset.


MartinaeyNL:
Just to mention; I included a fix for the "recalculating style" / forced reflow bug of `or-mwc-input` in #851.
I used an [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) instead, making it listen to visual changes of an element, which solved the problem while keeping the same behavior of changing the menu width :)

richturner (approved):

PR #851, opened 2022-10-27
Rules performance improvements

MartinaeyNL:
Changelog:

- Added batching and caching for the API requests for Assets; now only 1 request is made per AssetType.
- Solved frontend `or-mwc-input` performance bug in select using [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API); not causing "recalculate style" / forced reflow anymore.
- Improved or-mwc-input search; now also including a `searching` property and `or-mwc-inputsearch-changed` event.
- Added "loading" placeholders to prevent empty, unexplained areas.
- Minor bugfixes / styling changes.


MartinaeyNL:
Double-checked the code, just to be sure. Ready for review and merge @richturner

ui/component/or-rules/src/json-viewer/or-rule-action-attribute.ts
```diff
@@ -108,51 +123,73 @@ export class OrRuleActionAttribute extends translate(i18next)(LitElement) {
...
+                }
+                
+                return html`
+                    <or-mwc-input id="matchSelect" class="min-width" .label="${i18next.t("asset")}" .type="${InputType.SELECT}"
+                                  .searchable="${this._assets!.length > 25}" .options="${idOptions}" .value="${idValue}" .readonly="${this.readonly || false}"
+                                  @or-mwc-input-changed="${(e: OrInputChangedEvent) => { this._assetId = (e.detail.value); this._assetSearchValue = ''; this.refresh(); }}"
+                                  @or-mwc-inputsearch-changed="${debounce((event: OrInputSearchChangedEvent) => { this._assetSearchValue = event.detail.value; }, 500)}"
```
richturner, line 183: As previous comment, maybe directly use `@keyup` here if `input-changed` isn't suitable

MartinaeyNL:
Replaced the custom event with a callback property.
I tried using `@keyup` directly, but decided not to since it would become a big mess with shadow dom querying.
A `searchProvider` was also considered, but since it is only a `string[].filter()` I thought it was unnecessairy to make custom logic possible outside of the component.

For now the `searchCallback` property is optional to listen for changes, and setting `searchable` to true is enough to enable the functionality.

MartinaeyNL:
Went with a more abstract implementation using an `optionsProvider`.
This is an optional asynchronous function where you can return a customized list of options.
Included placeholders when the `Promise` is not resolved yet, showing "loading" or "no results found" respectively.

`searchable` property can still be used standalone, to filter within the `options` array.
If used with `optionsProvider`, it will add the current search input as function parameter.

Ready to merge.

MartinaeyNL:
Ready to merge.

ui/component/or-rules/src/json-viewer/or-rule-action-attribute.ts
```diff
@@ -107,52 +122,81 @@ export class OrRuleActionAttribute extends translate(i18next)(LitElement) {
...
-            <or-mwc-input id="matchSelect" class="min-width" .label="${i18next.t("asset")}" .type="${InputType.SELECT}" @or-mwc-input-changed="${(e: OrInputChangedEvent) => this._assetId = (e.detail.value)}" .readonly="${this.readonly || false}" .options="${idOptions}" .value="${idValue}"></or-mwc-input>
-            ${attributes.length > 0 ? html`
-                <or-mwc-input id="attributeSelect" class="min-width" .label="${i18next.t("attribute")}" .type="${InputType.SELECT}" @or-mwc-input-changed="${(e: OrInputChangedEvent) => this.setActionAttributeName(e.detail.value)}" .readonly="${this.readonly || false}" .options="${attributes}" .value="${this.action.attributeName}"></or-mwc-input>
-                ${attributeInput}
-            ` : html`
-                <or-translate value="No attributes with write permission"></or-translate>
-            `}
+            
+            <!-- Show SELECT input with 'loading' until the assets are retrieved -->
+            ${when((!this._assets), () => html`
```
richturner, line 130: Here all the initial async options gathering is done at this higher level which comes back to my point about handling it at a higher level rather than in the `or-mwc-input` component

richturner (changes requested):

richturner (dismissed):

MartinaeyNL:
Still busy with last fixes, will mention when ready for merge.

richturner (approved):

PR #852, opened 2022-10-27
Fix bug where rule engine asset states weren't updated when asset was modified

richturner:


richturner:
@MartinaeyNL please take a look at this bug fix for the rules engine

test/src/test/resources/org/openremote/test/rules/BasicBrokenRules.groovy
```diff
@@ -1 +1,3 @@
+package org.openremote.setup.integration.rules
+
 THIS IS NOT A VALID GROOVY FILE.
```
MartinaeyNL, line 3: Empty groovy test file?

MartinaeyNL (commented):
Reviewed. Nice work with a good set of fixes and updates.
Only spotted an empty groovy file, thats all.

MartinaeyNL (approved):

PR #854, opened 2022-10-31
Hotfix for or-mwc-input select; ripple not behaving to Material Design spec

MartinaeyNL:
Includes a hotfix for `or-mwc-input` SELECT type when `outlined` property was enabled.
The ripple had unusual behavior, and it is now behaving to the Material Design specifications.

### DIFF
--- a/ui/component/or-mwc-components/src/or-mwc-input.ts
+++ b/ui/component/or-mwc-components/src/or-mwc-input.ts
@@ -1007,7 +1007,7 @@ export class OrMwcInput extends LitElement {
                                      aria-expanded="false"
                                      aria-disabled="${""+(this.disabled || this.readonly)}"
                                      aria-labelledby="label selected-text">
-                                    <span class="mdc-select__ripple"></span>
+                                    ${!outlined ? html`<span class="mdc-select__ripple"></span>` : undefined}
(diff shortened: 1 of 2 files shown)


ui/component/or-mwc-components/src/or-mwc-input.ts
```diff
@@ -1007,7 +1007,7 @@ export class OrMwcInput extends LitElement {
                                      aria-expanded="false"
                                      aria-disabled="${""+(this.disabled || this.readonly)}"
                                      aria-labelledby="label selected-text">
-                                    <span class="mdc-select__ripple"></span>
```
DonWillems, line 1010: Isn't it weird to put undefined there (why not just if !outlined put html there? no else)? Or is that common practice?

MartinaeyNL: Used it myself in multiple places, for example the dashboard builder. Would imagine leaving it `undefined` would render nothing instead of an empty html, which would be uneccessairy. Had no issues using it yet.

DonWillems (approved):

MartinaeyNL:
Ready to merge.

richturner (approved):

PR #881, opened 2022-11-25
Enable update of agent related root assets

rainerhitz:
Code change is needed for agents that are used to update assets no matter if they are root elements or not.

### DIFF
--- a/manager/src/main/java/org/openremote/manager/agent/AgentService.java
+++ b/manager/src/main/java/org/openremote/manager/agent/AgentService.java
@@ -185,16 +185,6 @@ public class AgentService extends RouteBuilder implements ContainerService, Asse
     @Override
     public <T extends Asset<?>> T mergeAsset(T asset) {
         Objects.requireNonNull(asset.getId());
-        Objects.requireNonNull(asset.getParentId());
-
-        // Do basic check that parent is at least an agent...doesn't confirm its' the correct agent so
-        // that's up to the protocol to guarantee
-        // TODO: need to revisit this once agent generated assets logic finalised
-//        if (!getAgents().containsKey(asset.getParentId())) {
-//            String msg = "Cannot merge protocol-provided asset as the parent ID is not a valid agent ID: " + asset;
-//            LOG.warning(msg);
-//            throw new IllegalArgumentException(msg);
-//        }
 
         // TODO: Define access permissions for merged asset (user asset links inherit from parent agent?)
         LOG.fine("Merging asset with protocol-provided: " + asset);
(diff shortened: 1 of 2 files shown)


manager/src/main/java/org/openremote/manager/agent/AgentService.java
```diff
@@ -185,16 +185,6 @@ public void configure() throws Exception {
...
-        // that's up to the protocol to guarantee
-        // TODO: need to revisit this once agent generated assets logic finalised
-//        if (!getAgents().containsKey(asset.getParentId())) {
-//            String msg = "Cannot merge protocol-provided asset as the parent ID is not a valid agent ID: " + asset;
-//            LOG.warning(msg);
-//            throw new IllegalArgumentException(msg);
-//        }
 
         // TODO: Define access permissions for merged asset (user asset links inherit from parent agent?)
         LOG.fine("Merging asset with protocol-provided: " + asset);
```
richturner, line 190: There is a risk here that protocols can commit assets across realms which we cannot allow.

`assetService` in `AbstractProtocol` needs to proxy the merge asset and check that ```realm = agent realm``` so:

```
ProtocolAssetService protocolAssetService = container.getService(ProtocolAssetService.class);
        assetService = new ProtocolAssetService() {
            @Override
            public <T extends Asset<?>> T mergeAsset(T asset) {
                if (TextUtil.isNullOrEmpty(asset.getRealm())) {
                    asset.setRealm(getAgent().getRealm());
                } else if (!Objects.equals(asset.getRealm(), getAgent().getRealm())) {
                    Protocol.LOG.warning("Protocol attempting to merge asset into another realm: " + agent);
                    throw new IllegalArgumentException("Protocol attempting to merge asset into another realm");
                }
                return protocolAssetService.mergeAsset(asset);
            }

           ....
     }
```

Same for all other operations.

richturner (changes requested):

manager/src/main/java/org/openremote/manager/agent/AgentService.java
rainerhitz: What do you think about the following ?

 ```
@Override
 public void subscribeChildAssetChange(String agentId, Consumer<PersistenceEvent<Asset<?>>> assetChangeConsumer) {
     Asset<?> asset = protocolAssetService.findAsset(agentId);
     if (asset != null) {
         if (!Objects.equals(asset.getRealm(), getAgent().getRealm())) {
             Protocol.LOG.warning("Protocol attempting to subscribe to asset change events from another realm: " + agent);
             throw new IllegalArgumentException("Protocol attempting to subscribe to asset change events from another realm");
         }                       
     }
     protocolAssetService.subscribeChildAssetChange(agentId, assetChangeConsumer);
}
```
Same for  `unsubscribeChildAssetChange`.

richturner: I don't think this step is necessary; you need to know the agent ID already which is enough of a 'protection' against incorrect usage.

Looking up the agent again using `protocolAssetService.findAsset(agentId)` and then comparing it with what is in the agent map doesn't achieve anything (unless you want to check that something hasn't modified the agent realm in the map).

Subscribe and unsubscribe are 'safe' as you need to already know the agent ID.

richturner (approved):

PR #884, opened 2022-12-01
Manager configuration V1 

NealGeilen:
Creating the first version of version of a configuration page.
This update places the settings of the `manager_config.json` inside the manager with the help of form fields and a JSON editor. 

The first version consists of the visual settings for realms.


ui/app/manager/src/index.ts
```diff
@@ -257,10 +239,10 @@ fetch(configURL).then(async (result) => {
         // Add config prefix if defined (used in dev)
         if (CONFIG_URL_PREFIX) {
             Object.values(orAppConfig.realms).forEach((realmConfig) => {
-                if (typeof (realmConfig.logo) === "string") {
+                if (typeof (realmConfig.logo) === "string" && !realmConfig.logo.startsWith(" data:")) {
```
MartinaeyNL, line 242: Just checking whether having an extra spacebar is correct here. (" data:")

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -0,0 +1,186 @@
...
+        if (!manager.authenticated) {
+            return html`
+                <or-translate value="notAuthenticated"></or-translate>
+            `;
+        }
+
+
+
+        const app = this
+        document.addEventListener('saveLocalManagerConfig', (e:CustomEvent) => {
```
MartinaeyNL, line 141: You can safely use `this.addEventListener()` here as far as I'm aware. (same for all other `document` references in the file)

model/src/main/java/org/openremote/model/configuration/ManagerConf.java
```diff
@@ -0,0 +1,25 @@
...
+    protected String styles = null;
+    protected String logo = null;
+    protected String logoMobile = null;
+    protected String favicon = null;
+    protected String language = null;
+    protected ManagerHeaders[] headers = null;
+}
+
+enum ManagerHeaders {
+    rules, insights, gateway, logs, account, users, assets, roles, realms, logout, language, export, map, appearance, provisioning
```
MartinaeyNL, line 24: Don't know whether the backend should be aware of the available pages.

NealGeilen: This is part of the validation.

ui/component/or-components/src/or-file-uploader.ts
```diff
@@ -0,0 +1,151 @@
...
+            this.dispatchEvent(new CustomEvent("change", {
+                detail: { value: files },
+            }));
+        }
+        this.loading = false;
+        this.requestUpdate();
+    }
+
+
+    convertBase64 (file:any) {
```
MartinaeyNL, line 106: Would rather put this in a more global file. `util.ts` in the component/core has more small functions like this.

ui/component/or-configuration/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "@openremote/or-icon": "^1.1.0",
+    "@openremote/or-mwc-components": "^1.1.0",
+    "@openremote/or-translate": "^1.1.0",
+    "@openremote/rest": "workspace:*",
+    "ace-builds": "^1.4.13",
+    "linqts": "^1.12.6",
+    "lit": "^2.0.2",
+    "moment": "^2.24.0",
+    "resize-observer": "^1.0.3",
+    "rrule": "^2.6.4",
```
MartinaeyNL, line 32: Small note; double check the dependencies. Don't know if you need this.

ui/component/or-configuration/src/or-conf-realm/or-conf-realm-card.ts
```diff
@@ -0,0 +1,351 @@
...
+                    `)
+      .setDismissAction(null));
+
+  }
+
+
+  render() {
+    const colors = this._getColors();
+    const app = this;
+    document.addEventListener('saveManagerConfig', () => {
```
MartinaeyNL, line 254: This will endlessly create eventlisteners, making multiple active at the same time. Make sure only 1 exists on `firstUpdated()` and then saved onto state, or what ever you solution you find suitable ;)

ui/component/or-configuration/src/style.ts
```diff
@@ -0,0 +1 @@
+
```
MartinaeyNL, line 1: Empty file, or maybe you were planning to move some css in here. Just mentioning. ;)

ui/component/or-configuration/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "@openremote/or-attribute-input": "^1.1.0",
+    "@openremote/or-components": "^1.1.0",
+    "@openremote/or-icon": "^1.1.0",
+    "@openremote/or-mwc-components": "^1.1.0",
+    "@openremote/or-translate": "^1.1.0",
+    "@openremote/rest": "workspace:*",
+    "ace-builds": "^1.4.13",
+    "linqts": "^1.12.6",
+    "lit": "^2.0.2",
+    "moment": "^2.24.0",
```
MartinaeyNL, line 30: Could upgrade to 2.29.4 since it is used at multiple places, according to yarn.lock

ui/app/manager/src/pages/page-configuration.ts
NealGeilen: After just validation. it won't work when using `this`

manager/src/main/java/org/openremote/manager/web/ManagerWebService.java
```diff
@@ -77,7 +77,7 @@ protected interface ServerVariableMixin {
     public static final String OR_APP_DOCROOT = "OR_APP_DOCROOT";
     public static final String OR_APP_DOCROOT_DEFAULT = "manager/src/web";
     public static final String OR_CUSTOM_APP_DOCROOT = "OR_CUSTOM_APP_DOCROOT";
-    public static final String OR_CUSTOM_APP_DOCROOT_DEFAULT = "ui/app";
+    public static final String OR_CUSTOM_APP_DOCROOT_DEFAULT = "deployment/manager/app";
```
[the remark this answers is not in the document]
NealGeilen: (https://github.com/openremote/openremote/blob/master/profile/deploy.yml#L228) Changed it so it has the same value as the value that is used within docker.

richturner:
Some feedback:

* Realm selection should be sorted alphabetically
* Primary menu order determines the default page (i.e. the first item in the array) so ideally the order should be configurable

NealGeilen:
> * [ ]  Primary menu order determines the default page (i.e. the first item in the array) so ideally, the order should be configurable
@richturner Discussed this with @DonWillems and this will be a feature in the future. 
This is now possible through the JSON editor

ui/app/manager/src/index.ts
```diff
@@ -170,6 +151,7 @@ fetch(configURL).then(async (result) => {
     orApp.managerConfig = appConfig.manager;
 
     orApp.appConfigProvider = (manager) => {
+        manager.managerAppConfig = appConfig as unknown as ManagerConf
```
richturner, line 154: This soft cast to `unknown` then to `ManagerConf` I guess is because `ManagerAppConfig` model is missing the `manager` key which should be fixed by comments above in `ManagerConf`

ui/component/or-components/src/or-file-uploader.ts
```diff
@@ -0,0 +1,133 @@
+import { css, html, LitElement, PropertyValues } from "lit";
```
richturner, line 1: Copyright

ui/component/core/src/util.ts
```diff
@@ -1056,3 +1056,18 @@ export function enableScroll() {
...
+
+        fileReader.onload = () => {
+            resolve(fileReader.result);
+        };
+
+        fileReader.onerror = (error) => {
+            reject(error);
+        };
+    });
+}
```
richturner, line 1073: * `file:any` is not true...`FileReader` expects a `Blob` object 
* This should be named `blobToBase64(blob:Blob)` and also it should used the code from `or-asset-viewer/index.ts:469` (where the URL prefix is stripped) and then this util method can be re-used here and in `or-asset-viewer`

ui/component/or-components/src/or-file-uploader.ts
richturner: * Should just have a simple `get Files(): FileList {}` so it behaves like the standard File input element just with styling added (don't do any encoding and exposing that as a custom event)
* `or-asset-viewer` `fileupload-elem` should also be replaced with this component

ui/component/or-app/tsconfig.json
```diff
@@ -20,6 +20,7 @@
     { "path": "../../component/or-data-viewer" },
     { "path": "../../component/or-log-viewer" },
     { "path": "../../component/or-chart" },
+    { "path": "../../component/or-configuration" },
```
richturner, line 23: Should be part of the `manager` app (see below), will be removed from here

richturner (changes requested):
Happy to discuss anything I've mentioned, not the easiest going through so much code and there's lots of feedback unfortunately but we'll get there

manager/src/main/java/org/openremote/manager/web/ManagerWebService.java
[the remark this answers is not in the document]
NealGeilen: Ok, I will change it back. Is there another variable that I can use? so that it also works in development? If not I will go and hardcode a path. It's just not my preference

model/src/main/java/org/openremote/model/configuration/ManagerConf.java
NealGeilen: Oke Just want to mention that one of the given requirements was backend validation. 

But if it's not needed I will remove it.

ui/component/or-components/src/or-info.ts
```diff
@@ -0,0 +1,60 @@
+import { css, html, LitElement } from "lit";
```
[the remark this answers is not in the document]
NealGeilen: This component is no longer used and thus removed

model/src/main/java/org/openremote/model/manager/Credentials.java
```diff
@@ -0,0 +1,25 @@
+/*
```
richturner, line 1: Can be replaced with `org.openremote.model.auth.UsernamePassword`, feel free to adjust the ts code to use this type also

ui/component/or-app/src/index.ts
```diff
@@ -15,15 +15,15 @@ import "./or-header";
...
+import manager, {DefaultColor2, DefaultColor3, DefaultColor4, Util, BasicLoginResult, normaliseConfig, Manager} from "@openremote/core";
 import {DEFAULT_LANGUAGES, HeaderConfig} from "./or-header";
 import {OrMwcDialog, showErrorDialog, showDialog} from "@openremote/or-mwc-components/or-mwc-dialog";
 import {OrMwcSnackbar} from "@openremote/or-mwc-components/or-mwc-snackbar";
 import {AnyAction, Store, Unsubscribe} from "@reduxjs/toolkit";
 import {AppStateKeyed, updatePage, updateRealm} from "./app";
 import { InputType, OrInputChangedEvent } from "@openremote/or-mwc-components/or-mwc-input";
 import { ORError } from "@openremote/core";
-import { Realm } from "@openremote/model";
+import { Auth, ManagerAppConfig, ManagerConfig, Realm } from "@openremote/model";
```
richturner, line 26: `ManagerAppConfig` shouldn't be needed in here

richturner (changes requested):
Well done for decoding my previous feedback.

Only a couple of minor things, once done I will approve.

ui/component/or-app/src/index.ts
NealGeilen: It is. Because I need a way to hold onto the original settings before they are separated in other settings. Some settings of the `ManagerAppConfig` are used in `ManagerConfig` but not all. 

I found it not logical that the `manager_config.json` is loaded twice. With every page visited it's loaded. And when I don't use this it will be needed to load a second time when a user is on the config page.

DonWillems:
@NealGeilen Two questions about the manager_config as I've used it in other projects:
1. I've never used `"manager":{}` that is on the same level as `"realms"` in the JSON. What could it be used for?
2. Most of the time I have `"loadLocales": true,` at the same level as "realms". Is there any use to that?

NealGeilen:
> @NealGeilen Two questions about the manager_config as I've used it in other projects:

Both items will no longer show up automatically after saving. This was a feature/bug from the backend validation that is removed. 

> 1. I've never used `"manager":{}` that is on the same level as `"realms"` in the JSON. What could it be used for?

The manager object has [these](https://github.com/openremote/openremote/pull/884/files#diff-373830cf8a9db4a81646924b06dea40d40cbc6bc0dc329ae50a2fa91b6d90a7c) possible values. It's more technical and would not be used in most cases. 

> 2. Most of the time I have `"loadLocales": true,` at the same level as "realms". Is there any use to that?
This is a setting which is unclear to me what it exactly does. I added it because I saw it that is being used in some of our projects.

richturner (approved):

Issue #875, opened 2022-11-23
Rules: incorrect behaviour related to realms as admin user


DonWillems:
When logged in as an admin user you should be able to see the rules in each realm, but only of the realm you are in. You should also be able to see the assets of other realms while creating a global rule. You should not be able to see assets of other realms while creating a realm rule.

Bugs:
- [x] As admin user I can access assets from another realm while creating a realm rule. Expected: only see assets of the realm I'm in.
- [x] As admin user when I start creating a rule, switch realms, and save it, it is saved in the realm where it was 'created'. Expected: When switching trigger discard dialog, or if not possible empty the page/deselect rule.
- [x] Rules list does not show correct rules when switching back and forth between realms (only in case of Master realm?)

MartinaeyNL:
> As admin user when I start creating a rule, switch realms, and save it, it is saved in the realm where it was 'created'. Expected: When switching trigger discard dialog, or if not possible empty the page/deselect rule.

I was not able to show the discard dialog, since there is no easy way to "cancel" or block the realm switching.
It now simply erases your changes immediately.

PR #890, opened 2022-12-13
Hotfix for webhook, and other frontend bugfixes for Rules

MartinaeyNL:
Changelog:

**Webhook related:**
- Hotfix for headers not working in webhook
- Hotfix for bug where the HTTP webhook request was sent twice
- Hotfix for asset payload formatting incorrectly to JSON
- When using "content-type" header, it now overrides the default JSON type correctly when formatting the payload. It is also case-insensitive as it should be.
- Successful webhook responses are now also shown in logs.
- Added groovy test for verifying headers, media types, and payloads.

**Rules in general:**
- Fix a major case where caching for assets in rules would not work. (originally added in *#851*)
- Fix for *#875* where all assets of all realms were shown in rules.
- Added caching for or-rules-list, since it was requesting the list way too many times.
- Fix for *#875* where or-rules-list would misbehave on realm change. This was due to async issues where the previous realm was used.


richturner (changes requested):

MartinaeyNL:
Applied the requested changes.
I also reverted some changes, such as the bugfix for showing the incorrect list of rules. An explanation:
> Before, the list of rules got fetched multiple times with different realms in the payload. This caused the incorrect list of rules.
> Initially I implemented a `onRealmChange()` method to only load the rules of the new selected realm. However the solution was way simpler: the `refresh()` method in or-rule-list requests before updating the realm, and was unnecessairy due to the fetch done in `requestUpdate()`. Removing it solved the issue.

Furthermore, I have been writing a test for the webhook.
However I'm not done with it yet, as I need some help to intercept the requests with a mock server.
Since many developers are on holidays, I will continue on that later.

MartinaeyNL:
Added groovy test for webhook, verifying the media types, injected headers and custom payloads.
Ready for a final review.

ui/component/or-rules/src/json-viewer/or-rule-json-viewer.ts
```diff
@@ -120,8 +120,8 @@ export class OrRuleJsonViewer extends translate(i18next)(LitElement) implements
     @property({attribute: false})
     protected _ruleset!: RulesetUnion;
 
-    @state()
-    protected _loadedAssets: Map<string, Asset[]> = new Map<string, []>();
+    @state() // to be exact: Map<Realm name, Map<AssetType name, Asset[]>>
+    protected _loadedAssets: Map<string, Map<string, Asset[]>> = new Map<string, Map<string, Asset[]>>();
```
richturner, line 124: Components shouldn't be aware of different realms; this is a higher level function that is only applicable to super users; we shouldn't optimise for this scenario as most users will be regular users working within a single realm, don't mind if there is a performance hit switching realms.

I have always tried to avoid complexity of making components "realm aware". When the realm changes then there should be an ability to clear loaded assets so a `reset()/refresh()` function of some kind

richturner (changes requested):
Just minor point about not making components realm aware

richturner (dismissed):

richturner (approved):

Issue #896, opened 2022-12-21
Rules: scheduler dropdown doesn't adjust to width of options


pierrekil:
![Screenshot 2022-12-21 at 15 34 46](https://user-images.githubusercontent.com/11444149/208947995-d2e75aab-e09a-4584-a931-b81c32e280cf.png)

PR #907, opened 2022-12-27
Fix for incorrect select menu width

MartinaeyNL:
Made some bugfixes to cases where the `or-mwc-input`'s width was behaving incorrectly.
Mainly happening in dialogs, where the width is initially 0px. (or relatively small)

Changelog:
- Changed logic from MutationObserver to ResizeObserver, to support width changes as described above.
- Fixed individual case as described in #896

.
Requires manual testing through the UI, since the observer change could be breaking in some edge cases.


MartinaeyNL:
Tested on a deployment with @pierrekil and @DonWillems . No cases discovered where the `or-mwc-input` was broken after the change.

richturner (approved):

PR #909, opened 2022-12-29
Bugfix maplibre-gl-geocode

DasAuryn:
After searching with geocode, the marker now appears correctly


richturner (dismissed):
Thanks for your first contribution

richturner (approved):

PR #910, opened 2023-01-03
Fix DB migration script naming issue and fail hard and fast on future issues

richturner:
… naming issues


richturner:
Need to fix a naming issue with the DB migration script for adding functions and also change config so future naming issues won't fail silently

DonWillems (approved):

PR #915, opened 2023-01-12
Show MQTT sessions in user page of manager and allow MQTT connection status to be linked to attributes

richturner:


richturner:
As well as implementing what is described in the title I also:

- Improved MQTT authentication and handling of transient credentials (for auto provisioned devices)
- Noticed a `<or-mwc-table>` usage that didn't translate column headers so fixed that in here also.

MartinaeyNL (approved):

PR #917, opened 2023-01-12
CI/CD: Fix issue with latest docker buildx version

richturner:


richturner:
@MartinaeyNL it took a while to figure out the CI/CD docker issue but it turns out a new version of buildx was released a couple of days ago and that has changed the structure of the docker manifest so it is no longer compatible with the current version of the `docker manifest inspect` tool.

Switched over to using buildx to inspect the manifest as recommended in a buildx github issue; hopefully it will work!

MartinaeyNL (approved):

PR #923, opened 2023-01-17
Map customization and settings 

NealGeilen:


manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -78,6 +87,31 @@ public class MapService implements ContainerService {
...
+            mapSettings.putPOJO("options", mapConfiguration);
+            mapper
+                    .setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY)
+                    .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+                    .enable(SerializationFeature.INDENT_OUTPUT);
+
+            out.write(mapper.writeValueAsString(mapSettings).getBytes());
+            out.close();
+            this.setData();
+        } catch (IOException exception) {
```
richturner, line 107: Best practice is to use try-resource with anything that implements `Closeable`; see:

https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

model/src/main/java/org/openremote/model/map/MapResource.java
```diff
@@ -25,11 +25,19 @@
...
 @Tag(name = "Map")
 @Path("map")
 public interface MapResource {
 
+    /**
+     * Saves the settings for maps
+     */
+    @PUT
+    @Produces(MediaType.APPLICATION_JSON)
+    Object saveSettings(@BeanParam RequestParams requestParams, Object mapConfig) throws Exception;
```
richturner, line 39: Is there a reason to use `Object` here rather than your new `MapConfig`?

ui/app/manager/src/components/configuration/or-conf-map/index.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  updated(changedProperties: PropertyValues) {
+    super.updated(changedProperties);
+  }
+
+  render() {
+    const app = this;
+    return html`
+      <div class="panels">
+        ${Object.entries(this.config === undefined ? {} : this.config).map(function([key , value]){
```
richturner, line 128: Use arrow function

richturner (changes requested):
Just a couple of minor points

richturner (dismissed):

ui/app/manager/src/components/configuration/or-conf-map/index.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  protected _showAddingRealmDialog(){
+    this._addedRealm = null;
+    const _AddRealmToView =  () => {
+      if (this._addedRealm){
+        if (!this.config){
+          this.config = {}
+        }
+        this.config[this._addedRealm] = {bounds: [4.42, 51.88, 4.55, 51.94], center: [4.485222, 51.911712], zoom: 14, minZoom: 14, maxZoom: 19, boxZoom: false}
```
richturner, line 70: @DonWillems These hardcoded values are strange; will leave for now and merge this but might want an issue to address this

richturner (approved):

PR #930, opened 2023-01-26
Logging improvements

richturner:
- Cleanup of logging now prefer System.Logger
- Camel JMX support now enabled
- Cleanup of CI/CD and updated packages
- Removed old OR2.x controller protocol


richturner:
@MartinaeyNL This PR looks bigger than it is; mostly cleanup of logging so lots of log statements have changed across files.

PR #932, opened 2023-01-30
Feature: Datapoints improvement for charting with use of algorithms/calculations

MartinaeyNL:
Changelog (30/01/2023):

- Added migration script registering the timescaledb extension and creating the 'hypertable'.
- Add AssetDatapointQuery model that includes from- and to-timestamp, amount of data points, asset ID, attribute name, and the “hyperfunction” to use.
- Added /asset/datapoint/query HTTP POST endpoint, which uses AssetDatapointQuery.
- Written several SQL queries using “hyperfunctions” such as LTTB and ASAP_SMOOTH,
that are used by the new /asset/datapoint/query endpoint.
- Added 2 docker compose profiles with timescaledb image for testing purposes.
- Adjusted or-chart component to use the new query endpoint.


MartinaeyNL:
Should specify this code is currently in a **research stage**. The endpoint, queries, and frontend code are purely experimental for prototyping purposes. It has not been polished nor made efficient at all.

manager/src/main/resources/org/openremote/manager/setup/database/V20230120_01__Timescale.sql
```diff
@@ -0,0 +1,3 @@
+create extension if not exists timescaledb;
+create extension if not exists timescaledb_toolkit; /*Not sure whether this is required*/
```
richturner, line 2: Best to verify this

model/src/main/java/org/openremote/model/query/AssetDatapointQuery.java
```diff
@@ -0,0 +1,32 @@
...
+    public static class Hyperfunction {
+        public Function function;
+    }
+
+    public enum Function {
+        LTTB, // default
+        GP_LTTB,
+        ASAP_SMOOTH
+    }
+}
```
richturner, line 32: Subtypes might be more appropriate if there are parameters associated with each type

richturner (approved):

richturner (approved):

richturner (dismissed):

richturner (dismissed):

richturner (approved):

PR #937, opened 2023-01-31
Dropdown width fix; now only stops observing element on disconnect

MartinaeyNL:
Reverted back to IntersectionObserver, and now only stops observing element on disconnect


MartinaeyNL:
Sorry for the minor edit. Finished now.

Issue #922, opened 2023-01-16
Users: hide realm role: 'create realm', add option to make a Master user 'super user'


DonWillems:
On the Master realm the realm role 'create realm' is visible when creating/editing a user. This role should be hidden (as are others). Then a checkbox to make a user 'Super user' should be added. Only another Super user can do this, and only on for users on the Master realm.

- [ ] Users page: Hide 'create realm' from realm roles select
- [ ] Users page: Add checkbox to set user to 'Super user'. Only if the current user is a Super user. Only for users on the Master realm.
- [ ] Make 'Manager roles' select and the individual roles disabled when Super user select is chosen, as that user has all access.
- [x] Make sure only the Super user 'admin' has the effects of the manager_config disabled. Other Super users should see the impact of these settings.

Note for future improvements, to be done later:
- Add explanation to items on users page
- Revisit all roles
- Set roles in separate panel so more space for explanations of roles?

richturner:
Super user is just a realm role called `admin` so I have just used this rather than complicate things further.

It doesn't make sense to disable the openremote client roles as they are used in conjunction with the `Super user` status...definitely a case for this to change but that relates to roles cleanup

PR #938, opened 2023-01-31
UI updates: Super user and realm page bug fixes/improvements

richturner:


richturner:
Fixes #922

richturner:
@DonWillems Also added login theme to realm UI...longer term this should be a list of available themes but at least it can be changed now

DonWillems (approved):

PR #960, opened 2023-02-14
JSON rules support for multiple attributes within a when condition

richturner:
Had to make some adjustments to backend code but seems to be functioning now, deployed to `test2`

Also made some adjustments to UI code to try and improve responsiveness a little as well as being able to show multiple attributes


DonWillems:
@richturner Looks good and works well. However I found an issue with using multiple 'any of this type' conditions in one AND When block. Whether using the same asset type or two different asset types, the rule does not trigger.
Is this something we want to fix in this PR or open a separate issue for it?

DonWillems (approved):

PR #970, opened 2023-02-21
Enhancement/small styling updates

DonWillems:
- Add link to user rows in the Linked Users panel on asset-viewer
- Add message when a service user has no active MQTT sessions
- Fix issue for configuration item dialog when screen height is too low


ui/app/manager/src/pages/page-assets.ts
```diff
@@ -343,4 +345,11 @@ export class PageAssets extends Page<AssetsStateKeyed>  {
             callHandler: !silent
         });
     }
+
+    protected _onLoadUserEvent(event: OrAssetViewerLoadUserEvent, silent: boolean = false) {
```
MartinaeyNL, line 349: Just checking with @richturner here.

I don't know how we should practice page switching.
Especially since this event is navigating to `users` (so outside of assets context)
Other than in `or-map-asset-card.ts` / `page-map.ts` we do not practice it anywhere,
so just checking whether the practice is okay.

This is what @richturner implemented in 2020 in `page-map.ts` navigating to an individual asset:
```
protected onLoadAssetEvent(loadAssetEvent: OrMapAssetCardLoadAssetEvent) {
    router.navigate(getAssetsRoute(false, loadAssetEvent.detail));
}
```

richturner (approved):

PR #972, opened 2023-02-27
create customAsset.java

AMadany1:
create a new asset type


richturner:
Thanks for the PR but this doesn't belong in our main code base; custom asset types are for custom projects

PR #973, opened 2023-02-27
Create CustomAsset.java

AMadany1:


richturner:
Thanks for the PR but this doesn't belong in our main code base; custom asset types are for custom projects

PR #974, opened 2023-02-27
Fix scroll issue for assets and rules

DonWillems:
When the asset tree/rules list becomes too long, it made the whole page scroll instead of only the list. Same for the asset-viewer/rules viewer that makes the list scroll if too high.


MartinaeyNL (approved):
I assume that you've checked whether it doesn't break anywhere else in the UI, so approved.

PR #991, opened 2023-03-13
Added momentary button icon

accabog:


pierrekil:
Hi Alexandru, thanks for your contribution(s)! We are happy to integrate those, but require signing a contributor license agreement. Can you drop me an e-mail at info@openremote.io so I can explain?

pierrekil:
Hi Alexandru, can you add your e-mail in the text send to info@openremote.io :-)

DonWillems:
Hi @accabog,

I'm always a bit reluctant adding more configuration items, especially if they are quite niche. Another direction could be to use the same button as used for 'Modify' in the top right of the assets page, but without the icon. And we use the attribute name as button text. What do you, and others, think?

image

accabog:
Hi @DonWillems ,
My issue is that if you add an attribute of type "setAsMomentary"  to an asset, you WILL get a round button, but the button will have no icon. The functionality to add the round button is there, just that it is impossible to set any ICON or TEXT to it. It will always look blank and invisible. That is why we need a way to make that button visible and if we can show an ICON on it, even better. 

It would be just like the refresh button on the Insights page
This is what it looks like w/o ICON:
image
image

And with ICON:
image

DonWillems:
Hi Alexandru,

First off, I really appreciate your contributions!
I understand what you did and why you did it. What I meant to ask was, what do you think about using the button type I posted a screenshot of instead of adding an icon to the currently invisible button. So replace the button type as a whole.
For me this has three benefits: 1, users don't need to know about an extra configuration item to use together with an asMomentary button; 2, we don't have to add a configuration item to the already a bit too long list; and 3, it is always clearly a button with an action defined by text (as opposed to only showing an icon that might only approximate the actual action).

So that why I asked what do you think about showing that button type instead of the round one with the icon. I think it would be an improvement. Just to be clear, I'm not asking you to make this change :) I will try to find some time at the end of the day.

Don

DonWillems:
Hmm, we need to discuss this a bit internally. I'll get back to you. Sorry for the wait

delfin2492:
> Hi @accabog,
> 
> I'm always a bit reluctant adding more configuration items, especially if they are quite niche. Another direction could be to use the same button as used for 'Modify' in the top right of the assets page, but without the icon. And we use the attribute name as button text. What do you, and others, think?
> 
> image

How to add button text in the attribute boolean asMomentary?

richturner:
@DonWillems were these concerns addressed? Seems like some useful insight esp. regarding two commands being sent on momentary button press

PR #995, opened 2023-03-14
FIX: offset is automatically calculated by Date function

accabog:
Should not add the offset again, as it is directly and correctly retuned by the Date function.
@pierrekil @DonWillems


DonWillems:
Quick note: we might need to do the same for the DATETIME input type.

richturner (approved):

PR #1004, opened 2023-03-15
Manager persistence updates (standardised storage volume and Flyway env variable)

richturner:
- Adds `OR_STORAGE_DIR` environment variable with default of `/storage` for services and agents to persist data to the file system; this path can be volume mapped in docker to allow persistence between container recreations
- Adds `OR_DB_FLYWAY_OUT_OF_ORDER` allows controlling Flyway DB outOfOrder option for applying migrations out of order if required


richturner:
Also merged in dependabot PR to avoid cluttering the commit history too much

MartinaeyNL (approved):

PR #1009, opened 2023-03-18
Gauge min/max are correctly set when 0, show role name on roles page

DonWillems:
Gauge did not properly set min/max value when changing it to 0.
Decimals set for gauge can't be below 0
Added back the role/permission name on the roles page.


ui/component/or-dashboard-builder/src/widgets/or-gauge-widget.ts
```diff
@@ -190,7 +190,7 @@ export class OrGaugeWidgetSettings extends LitElement {
                             ></or-mwc-input>
                         </div>
                         <div>
-                            <or-mwc-input .type="${InputType.NUMBER}" style="width: 100%;" .value="${config.decimals}" label="${i18next.t('decimals')}"
+                            <or-mwc-input type="${InputType.NUMBER}" style="width: 100%;" .value="${config.decimals}" label="${i18next.t('decimals')}" .min="${0}"
```
MartinaeyNL, line 193: This now prevents people from using negative values.

MartinaeyNL (changes requested):

ui/component/or-dashboard-builder/src/widgets/or-gauge-widget.ts
DonWillems: I did because I don't know what -2 decimals means. That 100 becomes 1? 🙃

MartinaeyNL (approved):

PR #1020, opened 2023-03-30
Fix restricted user AssetEvent bug

### LINKED ISSUE #969: Assets: a restricted user can read all attributes if an asset is updated
Reproduce:
- Log in as admin and restricted user on different browsers/private
- Go to the same asset page
- Make sure some attribute have restricted read off
- The admin changes something on the assets and saves it. So change name/add attribute/add or change config item etc.
- The restricted users see all attribute values, and updates to them, as long as they don't refresh the page.

Expected:
- The restricted user only ever sees the attributes with restricted read true

richturner:
Resolves #969

### DIFF
--- a/test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
+++ b/test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
@@ -27,6 +27,11 @@ import org.openremote.manager.agent.AgentService
 import org.openremote.manager.asset.AssetProcessingService
 import org.openremote.manager.asset.AssetStorageService
 import org.openremote.manager.setup.SetupService
+import org.openremote.model.asset.AssetEvent
+import org.openremote.model.attribute.Attribute
+import org.openremote.model.event.shared.SharedEvent
+import org.openremote.model.value.ValueType
+import org.openremote.setup.integration.KeycloakTestSetup
 import org.openremote.setup.integration.ManagerTestSetup
 import org.openremote.model.asset.AssetFilter
 import org.openremote.model.asset.agent.ConnectionStatus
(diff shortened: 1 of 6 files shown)


test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
```diff
@@ -53,10 +56,12 @@ import static org.openremote.model.Constants.MASTER_REALM_ADMIN_USER
  */
 class WebsocketClientTest extends Specification implements ManagerContainerTrait {
 
-    protected static <T> T messageFromString(String message, String prefix, Class<T> clazz) {
+    protected static <T> T messageFromString(String message) {
         try {
-            message = message.substring(prefix.length())
-            return ValueUtil.JSON.readValue(message, clazz)
+            def isSubscription = message.startsWith(EventSubscription.SUBSCRIBED_MESSAGE_PREFIX)
+            def isTirggered = !isSubscription && message.startsWith(TriggeredEventSubscription.MESSAGE_PREFIX)
```
MartinaeyNL, line 62: Minor typo

MartinaeyNL (changes requested):

test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
```diff
@@ -71,6 +76,194 @@ class WebsocketClientTest extends Specification implements ManagerContainerTrait
...
+                        null,
+                        "building",
+                        "building"))
+        client2.setEncoderDecoderProvider({
+            [new AbstractNettyIOClient.MessageToMessageDecoder<String>(String.class, client2)].toArray(new ChannelHandler[0])
+        })
+
+        and: "we add callback consumers to the clients"
+        def connectionStatus = client.getConnectionStatus()
+        def connectionStatus2 = client.getConnectionStatus()
```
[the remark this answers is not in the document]
richturner: Good spot; doesn't affect the test as the test wouldn't pass if client2 wasn't connected but worth fixing

test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
richturner: Another good spot thanks

MartinaeyNL (approved):

MartinaeyNL (approved):

PR #1029, opened 2023-04-11
Bugfixes for several attribute types in datapoint query

MartinaeyNL:
Changelog:
- Wrote several tests for verifying datapoint queries return the correct values.
- Hotfix for `all` query, where it failed to return anything else other than `number` and `boolean` attribute types.
- Now throws exception when invalid attribute types are used in downsampling queries.
- `or-attribute-history` now uses either `LTTB` or `ALL` depending on the attribute type.


MartinaeyNL:
Had to push a small fix for a bug that was occuring. Now ready for review and merge :)

test/src/test/groovy/org/openremote/test/assets/AssetDatapointQueryTest.groovy
```diff
@@ -0,0 +1,305 @@
...
+        Collections.reverse(allDatapoints) // since they're returned in reverse other (from new to old)
+        allDatapoints.stream().forEach {dp -> {
+            assert dp.timestamp == lttbDatapoints1[index].timestamp
+            assert dp.value == lttbDatapoints1[index].value
+            index++
+        }}
+
+        /* ------------------------- */
+
+        and: "requesting 3 datapoints using the LTTB algorithm should return 10, 30 and 50"
```
richturner, line 108: Comment doesn't match assertions

DonWillems (approved):

PR #1032, opened 2023-04-13
User query enhancements and JSON rule updates and fixes

### LINKED ISSUE #885: When-then rule only checks last entry when multiple "or" conditions are specified
When specifying multiple conditions (of any type) using 'or..', it only uses the last item.
The example that is shown in the screenshot, only triggers at 14:47, not at the other specified times.

image

Tested with multiple different condition types (assets, time of day, etc), on multiple deployments and they behave the same.

I have done no analysis yet, but when the other conditions were correct, it did not execute JsonRulesBuilder's `buildRuleActionExecution()` at all.
(only when making the last condition correct)

### LINKED ISSUE #961: Rules: When-Then: add realm role to linked users action
As an admin I want to send notifications to linked users with a certain realm role, so that only those users get a message and not all linked users.

- [ ] If the action 'push notification' or 'email' is chosen with the recipients 'users', add the items 'Linked: [realm role]' to the 'Users' select list.
- [ ] Add a separation between the items e.g. 'Linked, Linked: installer, Linked: end user' and username e.g. 'admin, James, Tobi' 
- [ ] If e.g. 'Linked: installer' is chosen only users that are linked to the asset that triggers the rule and have the realm role 'installer' will receive the message.

### LINKED ISSUE #1027: Git commit fails with submodule when no .git dir exists

richturner:
Fixes #1027, Closes #961, Fixes #885


richturner:
- Enhancements to `UserQuery` to allow user attributes and realm roles to be queried
- JSON Rules update to add extra options to Notification action with user target (each realm role can be used to filter matched linked users)
- JSON Rules fix for multiple OR conditions where only last would trigger the rule
- Fixed gradle issue when repo used as submodule and `.git` dir is missing
- Updated `@mdi/font` dependency

MartinaeyNL (approved):
Approved! Only saw a single comment I didn't understand; see feedback.

Issue #952, opened 2023-02-10
Map: Create option to add GeoJSON overlay on map


DonWillems:
As an admin I want to add a GeoJSON file I prepared to show on the map per realm, so that meaningful information is added for my customers.

For the first version 1 GeoJSON can be set per realm. This file will have the realmname as filename.
The files should be stored in deployment\map\geojson
File can be uploaded per realm in the mapsettings section of the Appearance page. Existing file will be overwritten by the new one.

Example file can be generated here: https://overpass-turbo.eu/s/1tma
Other tool to draw geometry yourself: https://geojson.io/#map=6.51/51.189/6.886

Only show the geometry, properties can be ignored.

PR #1034, opened 2023-04-18
Added GeoJSON functionality for Map

MartinaeyNL:
This adds new functionality where GeoJSON can be shown on top of the Map.
Using tools like https://geojson.io, you can define points and shapes to be drawn in the color of the realm.
Based on feature request #952 .

The changes are almost frontend-only, with only a minor change in the MapRealmConfig model.
*(which only the frontend uses after all)*

<h2>Feature Changelog:</h2>

- `or-map` now draws GeoJSON on both the Map- and Insights-page respectively.
It uses the primary color of the realm, and supports several geometry types such as points, polygons and lines.
- Added support for specifying GeoJSON sources per realm.
- Added GeoJSON editor to realm map settings on Appearance page.
- Added control to Insights page to enable or disable GeoJSON overlay.
- Minor styling fixes

.
<h2>Technical Changelog:</h2>

- Added `geoJSON` property to MapRealmConfig, that includes `sources` and `layers`.
Sources is a GeoJSON object, whilst layers is a MapLibre specific object to customize styling which is currently unused..
- Added default specification for each geometry type. (lines, polygons, points etc)
- Optimized drawing of GeoJSON layers on `or-map` by grouping sources and layers by geometry type. So, instead of drawing a layer for each of the 1000+ lines in the GeoJSON source, we group all lines together in 1 layer with the same styling.
- Added ability to hide/show geoJSON in `or-map`
- **Other:** fix for interval datapoint query; the previously used "HOUR" interval (without specifying amount) caused a crash.
- **Other:** Internal map resources are now correctly removed during cleanup
- Added necessary translations

.

---

Some examples created with https://overpass-turbo.eu:

| All roads in Rotterdam where the max speed limit is 50: | All touristic places in Rotterdam: |
| --- | --- |
| image | image |


MartinaeyNL:
Ready for a code review. Will verify the functionality together with @pierrekil and @DonWillems by testing on a deployment.

richturner:
I've created a PR to improve the robustness ForecastServiceTest #1037

ui/component/or-map/src/mapwidget.ts
```diff
@@ -45,23 +48,29 @@ export class MapWidget {
...
     protected _showGeoCodingControl: boolean = false;
     protected _showBoundaryBox: boolean = false;
     protected _useZoomControls: boolean = true;
+    protected _showGeoJson: boolean = true;
     protected _controls?: (Control | IControl | [Control | IControl, ControlPosition?])[];
     protected _clickHandlers: Map<OrMapMarker, (ev: MouseEvent) => void> = new Map();
     protected _geocoder?: any;
 
-    constructor(type: MapType, showGeoCodingControl: boolean, styleParent: Node, mapContainer: HTMLElement, showBoundaryBox = false, useZoomControls = true) {
+    constructor(type: MapType, styleParent: Node, mapContainer: HTMLElement, realm?: string, showGeoCodingControl: boolean = false, showBoundaryBox = false, useZoomControls = true, showGeoJson = true) {
```
richturner, line 65: Realm should not be passed in (ideally components should not be realm aware and this should be handled at the higher app level); using `manager.displayRealm` is tolerable for now until some refactoring is done for view settings.

Why isn't `manager.displayRealm` sufficient?

richturner (changes requested):

ui/component/or-map/src/mapwidget.ts
MartinaeyNL: Hi! Just changed it so `or-map` is not realm aware anymore.

The initial intent was to fix geoJSON overlay and other glitches within the Appearance page of the Manager UI.
On that page you can edit mapsettings of each individual realm, making `manager.displayRealm` insufficient.
It caused `smartcity` geoJSON to be shown on the mapsettings of `manufacturer`, which is confusing.

I now changed it to a GeoJSON override property on `or-map`, similar to how we handle properties like `center`.
By default it will use the mapsettings of `manager.displayRealm`, with the Appearance page being the only exception.

MartinaeyNL:
Made the requested changes by cleaning up realm-specific code.
Included some other bugfixes as well.

Ready for review.

ui/component/or-map/src/index.ts
```diff
@@ -436,6 +437,12 @@ export class OrMap extends LitElement {
     @property({type: Boolean})
     public useZoomControl: boolean = true;
 
+    @property({type: Object})
+    public geoJson?: GeoJsonConfig;
+
+    @property({type: Boolean})
+    public showGeoJson: boolean = true;
```
richturner, line 444: Why do we need this boolean property? Surely if you don't want to showGeoJson then just set geoJson to undefined/null

richturner (changes requested):

ui/component/or-map/src/index.ts
MartinaeyNL: The geoJson property acts as an override.
So if it is `undefined`, it will fallback to the geoJson of the `manager.displayRealm`.
`showGeoJson` can be used to hide it no matter what.

Furthermore, it is also a customizable option of the Map widget in the Insights page, where you can hide/show the GeoJSON layer.

richturner: The `geoJson` property seems to only contain the geo JSON that should be currently loaded on the map so therefore whatever higher level component is setting this can also decide whether it should be set based on some other dashboard widget option can't it?

Even if we can justify having both properties on `or-map` I don't see why you also need both on `mapwidget` for the reason above.

richturner (changes requested):

richturner:
How come `page-map` hasn't been updated as part of this work? Surely you would want to be able to display the geo JSON on the map page if it is defined in the manager config?

MartinaeyNL:
> How come `page-map` hasn't been updated as part of this work? Surely you would want to be able to display the geo JSON on the map page if it is defined in the manager config?

By default the map loads the GeoJSON of the current realm. (`manager.displayRealm`).
So `page-map` doesn't have to set the `geoJson` property. Only the Appearance page uses the property to override the GeoJSON that is loaded.

ui/component/or-map/src/index.ts
MartinaeyNL: Example: User is in the smartcity realm.
The behavior of the properties will be the following;
- `geoJson: undefined`: GeoJSON of the SmartCity realm will be loaded.
- `geoJson: {}`: GeoJSON object specified here will be loaded instead.

If combined with showGeoJson:
- `geoJson: undefined; showGeoJson: false`: Don't load the SmartCity GeoJSON, and hide the layer.

richturner (changes requested):

ui/component/or-map/src/mapwidget.ts
```diff
@@ -178,6 +186,14 @@ export class MapWidget {
         return this;
     }
 
+    public setGeoJson(geoJsonConfig?: GeoJsonConfig): this {
+        this._geoJsonConfig = geoJsonConfig;
+        if(this._mapGl) {
+            this.loadGeoJSON(this._geoJsonConfig);
```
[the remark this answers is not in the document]
MartinaeyNL: Good catch! Will fix 👍

richturner:
I should look more closely at the code before commenting :-)

As mentioned previously I don't want components realm aware at all if possible but this can all be refactored another time...if there's too much logic inside the component(s) then you need to add overrides and extra options to control the logic rather than letting the consumers of the components decide what to do in the first place.

richturner (approved):

PR #1035, opened 2023-04-18
Added new asset type for heat meters

rooedw:
Added a new asset type for heat meters, which I think would be beneficial for all users. Especially in the context of the current energy crisis and the need for data-centric solutions which require the use of a variety of different sensors.


pierrekil:
New asset types can be added in a custom project. The current asset types are merely examples and no need to add new ones in the main code.

Issue #1030, opened 2023-04-12
Assets: History: Table: data not immediately shown


DonWillems:
Reproduce:
1. On the assets page make sure you have an attribute with string or object with store data points and some data
2. Select that attribute in the history panel
3. Empty table is shown
4. Press double arrow, or back and forth with normal arrows
5. Data in table is shown

Expected:
step 3. shows a table filled with the correct data.

PR #1039, opened 2023-04-25
Bugfix for attribute history; data changes did not correctly update the UI.

MartinaeyNL:
Bugfix for #1030 .

## Breakdown of the issue
The `or-attribute-history` component was incorrectly updating / rendering the data.
I discovered the following behavior:

1. Data from the correct timestamps is fetched from the database.
2. The `_data` property gets updated, triggering a render.
3. Render takes place.
4. After render, during `updated()`, a new table is created with the new data using `this._getTableTemplate()`.
5. The `_tableTemplate` property gets updated.
6. LitElement tries to update, but we cancel it manually because of insufficient changes. (see snippet)

```typescript
 shouldUpdate(_changedProperties: PropertyValues): boolean {
        let reloadData = _changedProperties.has("period") || _changedProperties.has("toTimestamp") || _changedProperties.has("attribute");
        if (this._dataFirstLoaded && !_changedProperties.get('_loading') && !reloadData) {
            return false;
        }

// If only _tableTemplate is changed, it will return false, since nothing else is updated.
```

## Changes made
I could've included a check with `changedProperties.has('_tableTemplate')` within `shouldUpdate()`.
However, that would mean 2 updates are required for each data update.

So, I changed the `updated()` method to `willUpdate()`. See https://lit.dev/docs/components/lifecycle/#willupdate
Now the table changes **during** the same update, instead of having to trigger a 2nd render. (that initially was cancelled)

.

### DIFF
--- a/ui/component/or-attribute-history/src/index.ts
+++ b/ui/component/or-attribute-history/src/index.ts
@@ -446,6 +446,7 @@ export class OrAttributeHistory extends translate(i18next)(LitElement) {
                 }
             }
         } else {
+            // Updating the table template before render, if necessary, to prevent additional update
             if (!this._tableTemplate || changedProperties.has("_data")) {
                 this._tableTemplate = this._getTableTemplate();
             }
--- a/ui/component/or-attribute-history/src/index.ts
+++ b/ui/component/or-attribute-history/src/index.ts
@@ -335,7 +335,7 @@ export class OrAttributeHistory extends translate(i18next)(LitElement) {
     }
 
     // Method that is executed during an update, to compute (additional) values before rendering.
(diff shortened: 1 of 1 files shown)


MartinaeyNL:
Ready for review. Will keep it a draft for now, since I want to test on a deployment first.

richturner:
@MartinaeyNL Is this still a draft or is it finished?

richturner:
@MartinaeyNL can this be merged?

MartinaeyNL:
Forgot that this PR existed! Thanks.
Just tested it with @DonWillems on a deployment and it's indeed ready to merge.

PR #1045, opened 2023-04-28
Fix for deselecting asset when expanding/collapsing in or-asset-tree

### LINKED ISSUE #976: Assets: When expanding an asset in the tree, asset(s) are deselected
Reproduce:
- Select an asset
- Toggle expand on an unrelated asset
- Asset in the asset-viewer will deselect and other selection in the tree will also deselect

Expanded:
- Asset selection is maintained when expanders are toggled

MartinaeyNL:
Fixes #976 

Previously the asset was deselected when expanding/collapsing assets in the `or-asset-tree` sidebar.
This was caused by `stateChanged()` being triggered by the changes in the `_store` we make for caching the expanded assets.
I now included a check whether the `AppState` has changed (such as URL and parameters), since we don't need to update
the `editMode` nor `assetIds` in other cases.

### DIFF
--- a/ui/app/manager/src/pages/page-assets.ts
+++ b/ui/app/manager/src/pages/page-assets.ts
@@ -209,7 +209,6 @@ export class PageAssets extends Page<AssetsStateKeyed>  {
     // State is only utilised for initial loading, and for changes within the store.
     // On the assets page, we shouldn't change editMode nor assetIds if the URL/state hasn't changed.
     stateChanged(state: AppStateKeyed) {
-        // State is only utilised for initial loading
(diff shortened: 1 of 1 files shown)


MartinaeyNL:
Currently a draft, as I want to test the UI on a deployment somewhere next week.

MartinaeyNL:
Should mention @richturner;

I added a separate variable `_currentStateJsonString` to compare the changes of `stateChanged()`
I'm in favor of another approach, but couldn't think of a better way. Maybe a `changedProperties` parameter of some kind?

MartinaeyNL:
Ready for review and merge.

PR #1052, opened 2023-05-02
Fix for email notifications

### LINKED ISSUE #1044: Cannot send emails when OR_EMAIL_PROTOCOL is set to smtps
Looks like an issue with the mechanism the EmailNotificationHandler uses to send the emails; so the initial connection test during initialisation succeeds but it is never able to actually send an email.

This looks like it might be an issue relating to `smtps` and the static `Transport.send()` method not picking up the fact that we want `smtps` so it doesn’t find the configured java mail properties.

See this forum post for discussion:

https://forum.openremote.io/t/e-mail-notification-failed/1795/35

richturner:
Fixes #1044

### DIFF
--- a/manager/src/main/java/org/openremote/manager/notification/EmailNotificationHandler.java
+++ b/manager/src/main/java/org/openremote/manager/notification/EmailNotificationHandler.java
@@ -51,7 +51,9 @@ public class EmailNotificationHandler implements NotificationHandler {
 
     private static final Logger LOG = Logger.getLogger(EmailNotificationHandler.class.getName());
     protected String defaultFrom;
+    protected String protocol;
     protected Session mailSession;
+    protected Transport mailTransport;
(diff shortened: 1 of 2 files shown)


manager/src/main/java/org/openremote/manager/notification/EmailNotificationHandler.java
```diff
@@ -51,7 +51,9 @@ public class EmailNotificationHandler implements NotificationHandler {
 
     private static final Logger LOG = Logger.getLogger(EmailNotificationHandler.class.getName());
     protected String defaultFrom;
+    protected String protocol;
```
MartinaeyNL, line 54: Is this field necessary? I don't see any usages of it within the changes.

MartinaeyNL (changes requested):

MartinaeyNL (approved):

PR #1054, opened 2023-05-04
[WIP] Support for public available dashboards

### LINKED ISSUE #1046: Insights app: add login/logout button and for master admins a realm switch
As a user I want to log in when people share a public dashboard with me, so that I can see my other dashboards.

As a super admin user I want to switch between realms, so that I can see the insights boards of other realms.

As a user I want to log out, so that I can log in as a different user.

Add a log in or log out button at the bottom of the sidebar.
Add a realm switch button above the log out button in the sidebar for super admins.

### LINKED ISSUE #1047: Insights app: logo for master realm not showing.

### LINKED ISSUE #1050: Insights: page refresh 'crashes' tab
Reproduce:
- Switch between insights pages a few times and refresh
- Page is stuck in 'loading' and GET for /master/dashboards/all/master is not finished:
image

Initially seemed related to #1049 but is also reproduced without the dialog showing up. However, if encountering the dialog, pressing discard, and refreshing, this issue will appear too.

MartinaeyNL:
Added support for making dashboards public, removing the requirements for users to login.
This required several changes to the HTTP API, the Insights page, and the standalone Insights app.
It also fixes #1046, fixes #1047, and fixes #1050
An overview of the changes:

### Feature changelog:
- Major improvements to the `/dashboards` HTTP API endpoints; doesn't require authentication anymore,
and now correctly checks whether a user has access to the data.
- Datapoints query endpoint now supports fetching datapoints of `ACCESS_PUBLIC_READ` assets.
- Added "open in new window" button to Insights page, to directly view dashboards in the Insights app.
- Added `PUBLIC` option to view access dropdown in dashboard settings.

### Changes to Insights standalone App:
- Removed initial login redirect, to allow viewing of public dashboards. Added login and logout button to menu instead. [#1046](https://github.com/openremote/openremote/issues/1046)
- Support for realm switching within Insights app, by clicking "Change realm" button in the menu.
- Fix for 'master' logo not showing in menu. [#1047](https://github.com/openremote/openremote/issues/1047)
- Automatic login redirect if no public dashboards could be found in that realm.
- Visiting dashboard of another realm as superuser, now automatically switches the realm.

.
### Other technical bugfixes:
- Fix for Insights page ending in an infinite loop when visiting specific dashboard. [#1050](https://github.com/openremote/openremote/issues/1050)
- Removed extra fetch in `or-map-widget` to get `AttributeDescriptor`, using `Util` function instead.
- Change of `MapWidgetConfig` model to use `assetId` instead of the whole `Asset` object being included. 
- Fix for `or-map-widget` where reverting the asset type dropdown back to 'none' would cause errors.
- Unsupported value types in `or-map-widget` now fall back to `text`.
- Bugfix for `or-map-widget` where it would crash when the user couldn't access the location attribute of an asset.
- Added `fullWidth` property to `or-mwc-menu`, to adapt to the width of a parent component such as a button.
- Made DefaultLogo PUBLIC for use in other apps that do not practice a header.

.

### DIFF
--- a/manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
+++ b/manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
@@ -56,6 +56,9 @@ public class DashboardResourceImpl extends ManagerWebResource implements Dashboa
     public Dashboard get(RequestParams requestParams, String dashboardId) {
         boolean publicOnly = true;
         String realm = getRequestRealmName();
+        if(realm == null) {
+            throw new WebApplicationException(BAD_REQUEST);
+        }
         try {
             // Realm should be enabled. Also takes unauthenticated users into count.
             if(!isRealmActiveAndAccessible(realm)) {


MartinaeyNL:
Tested on a deployment and ready for a review!
Quite some changes in this PR, so taking a more detailed look would be really appreciated 😄

manager/src/main/java/org/openremote/manager/dashboard/DashboardStorageService.java
```diff
@@ -68,36 +67,40 @@ public void stop(Container container) throws Exception {
...
                 CriteriaQuery<Dashboard> cq = cb.createQuery(Dashboard.class);
                 Root<Dashboard> root = cq.from(Dashboard.class);
+
                 List<Predicate> predicates = new ArrayList<>();
+                if(dashboardId != null) {
+                    predicates.add(cb.like(root.get("id"), dashboardId));
+                }
                 if(realm != null) {
                     predicates.add(cb.like(root.get("realm"), realm));
                 }
```
richturner, line 86: Not sure it should be possible to query dashbaords across realms so this parameter should expect not null or throw

manager/src/main/java/org/openremote/manager/dashboard/DashboardStorageService.java
```diff
@@ -130,10 +134,9 @@ protected <T extends Dashboard> T update(T dashboard, String userId) {
         return persistenceService.doReturningTransaction(em -> {
             Dashboard d = em.find(Dashboard.class, dashboard.getId());
             if(d != null) {
-                if(d.getEditAccess() == DashboardAccess.PRIVATE) {
-                    if(!(d.getOwnerId().equals(userId))) {
+                if(d.getEditAccess() == DashboardAccess.PRIVATE && !(d.getOwnerId().equals(userId))) {
                         throw new WebApplicationException("You are not allowed to edit this dashboard!", FORBIDDEN);
```
richturner, line 138: This is technically leaky as WebApplicationException is a JAXRS exception but this code isn't a JAXRS class

ui/app/insights/src/pages/page-view.ts
```diff
@@ -0,0 +1,366 @@
...
+            this._activePromises.set(topic, promise);
+        }
+        promise.finally(() => { this.completePromise(topic, updateOnComplete); });
+        if(doUpdate) { this.requestUpdate("_activePromises"); }
+        return promise;
+    }
+    completePromise(topic: string, doUpdate: boolean = true) {
+        this._activePromises.delete(topic);
+        if(doUpdate) { this.requestUpdate("_activePromises"); }
+    }
```
richturner, line 326: I haven't spent time figuring out what this custom promise code is here but maybe the `Util.Deferred` class fits your needs?

richturner (changes requested):

manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointResourceImpl.java
```diff
@@ -87,7 +91,13 @@ public ValueDatapoint<?>[] getDatapoints(@BeanParam RequestParams requestParams,
                 throw new WebApplicationException(Response.Status.NOT_FOUND);
             }
 
-            if (!isRealmActiveAndAccessible(asset.getRealm())) {
+            // If not logged in, asset should be PUBLIC READ
+            if(!isAuthenticated() && !asset.isAccessPublicRead()) {
```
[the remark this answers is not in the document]
MartinaeyNL: What check do you mean exactly? Not sure if I got `ACCESS_RESTRICTED_READ` covered, but I thought line 84 had the restricted user logic covered.

ui/app/insights/src/pages/page-view.ts
```diff
@@ -0,0 +1,366 @@
...
+        }
+        let promise = this.getPromise('dashboard/all');
+        if(promise == undefined) {
+            promise = this.registerPromise('dashboard/all', manager.rest.api.DashboardResource.getAllRealmDashboards(realmName), true, false);
+        }
+        try {
+            const response = await promise;
+            return response.data;
+        } catch (ex) {
+            if(isAxiosError(ex) && ex.response.status == 403 && loginRedirect) {
```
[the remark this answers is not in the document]
MartinaeyNL: Resolved since it now won't return a 403 error anymore.

ui/app/insights/src/pages/page-view.ts
```diff
@@ -0,0 +1,366 @@
...
+            flexDirection: 'column',
+            pointerEvents: this.dashboardMenu?.isDrawerOpen() ? 'none' : 'auto'
+        }
+        return html`
+            ${when(!this.viewDashboardOnly, () => {
+                const realms = this._loadedRealms?.map((r) => ({ name: r.name, displayName: r.displayName }));
+                return html`
+                    <dashboard-menu id="dashboard-menu" .dashboards="${this._loadedDashboards}" .realms="${realms}" .realm="${this._realm}"
+                                    .selectedId="${this._selectedId}" .userId="${this._userId}" .loading="${this.isLoading()}" .logoMobileSrc="${logoMobile}"
+                                    @realm="${(ev: CustomEvent) => this.changeRealm(ev.detail.value)}"
```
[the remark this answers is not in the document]
MartinaeyNL: `or-header` has a realm picker to be exact, with the fetching logic of the realms sitting in `or-app`.
The Insights app doesn't use `or-header`, since it uses a custom sidebar/menu. (with options such as switching realms)
The realm picker is also a tiny bit different, being a big button instead of a dropdown text like in the manager.

Since in `or-header` it was implemented with `getContentWithMenuTemplate()` and a simple list of realm strings,
I made a similar implementation for the Insights app.

ui/app/insights/src/pages/page-view.ts
MartinaeyNL: Removed some of the unused functions in here.

Essentially it was meant to keep track of active promises for managing a "loading" state.
The problem previously that we changed the `_loading` state every time a fetch started, and when one was completed.
Like this:
```javascript
protected fetchSomething() {
    this._loading = true;
    fetch().then(() => {
        // make changes
    }).finally() => {
        this._loading = false;
    }
}
```

On this page, several HTTP fetches happen asynchronously, starting a different timestamps.
Causing `_loading` to be set to false, while other fetches are still active.
Sadly `Promise.all()` would not work here, since the promises are sometimes depending on each other,
or starting at different places in the code.

I was a bit experimental by coming up with something like this.
Having a `Map` with a `topic: string` and a `Promise` to store the active promises.
This gave flexibility of checking which fetches are active, do `Promise.all()` to wait for all promises to complete, et cetera.

But I completely understand it's a bit too much for what is necessairy haha

Open for suggestions 😄

ui/app/insights/src/pages/page-view.ts
```diff
@@ -0,0 +1,366 @@
...
+
+        // Clear loadedDashboards when people switch between realms or viewing modes.
+        // For example, if 'view this dashboard only' is active, we only fetch that specific dashboard.
+        // When switched to 'view all dashboards', or the other way around, we should fetch all dashboard(s) of that realm again.
+        if(changedProps.has("_realm") || changedProps.has("viewDashboardOnly")) {
+            this._loadedDashboards = undefined;
+        }
+
+        // Fetch dashboard(s) if none exist, and no related fetch calls are active.
+        if(this._loadedDashboards == undefined && !this.getPromise('dashboard/all') && !this.getPromise('dashboard/' + this._selectedId)) {
```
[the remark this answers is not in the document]
MartinaeyNL: See comment above.

MartinaeyNL:
Applied most of the changes mentioned in the review.
Also added some tests to verify access rights for several cases.

ui/app/insights/src/pages/page-view.ts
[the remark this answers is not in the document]
richturner: @DonWillems My concern here is that there are inconsistencies in our different UIs

manager/src/main/java/org/openremote/manager/dashboard/DashboardStorageService.java
richturner: Don't see a change here....

manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointResourceImpl.java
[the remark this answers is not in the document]
richturner: So a restricted user has to be linked to the asset (you have this check on L84) but also each attribute that restricted users are allowed to acces must have the `ACCESS_RESTRICTED_READ` configuration item for reading the attributes and/or `ACCESS_RESTRICTED_WRITE` configuration item for writing.

This check should also be made for reading datapoints of an attribute

richturner (changes requested):

manager/src/main/java/org/openremote/manager/dashboard/DashboardStorageService.java
```diff
@@ -68,36 +67,40 @@ public void stop(Container container) throws Exception {
...
-    protected <T extends Dashboard> Dashboard[] findAllOfRealm(String realm, String userId) {
-        return this.findAllOfRealm(realm, userId, false);
-    }
-
-    @SuppressWarnings("java:S2326")
-    protected <T extends Dashboard> Dashboard[] findAllOfRealm(String realm, String userId, Boolean editable) {
+    // Querying dashboards from the database
+    // userId is required for checking dashboard ownership. If userId is NULL, we assume the user is not logged in.
+    // canEdit can be used to only return editable dashboards when 'true'.
+    protected <T extends Dashboard> Dashboard[] query(String dashboardId, String realm, String userId, Boolean canEdit) {
```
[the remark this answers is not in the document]
MartinaeyNL: The code here is from an older commit that doesn't include the most recent changes of mine.
I changed the parameter to `editable` in a commit after the 1st review.

ui/app/insights/src/pages/page-view.ts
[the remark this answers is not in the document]
MartinaeyNL: Discussed it with @DonWillems and @pierrekil earlier.
The content of the sidebar is quite different, and the Insights app has a kind of "mobile"-approach.
I get what you're coming from though.

Here are some visuals for context:

**Header in Manager app:**
or-header-mei2023

---

**Sidebar in Insights app:**
insights-sidebar

MartinaeyNL:
Applied the requested changes.

See the last unresolved comment for some context on the requested changes for the header.

richturner (dismissed):

richturner:
Approved to avoid this being held up on the realm picker issue which can always be addressed separately

MartinaeyNL:
Good point about the realm picker being adressed seperately.
I should get used to making smaller PR's as well.

Please **wait with merging**.
I didn't test the latest changes on a deployment, which broke some of the functionality.
Will push a bugfix for it before all hell breaks loose. 🤣

ui/app/insights/src/pages/page-view.ts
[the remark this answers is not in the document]
richturner: How does it look when you use a desktop? Still have the pullout sidebar?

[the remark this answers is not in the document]
MartinaeyNL: Yes. Previously it was the similar to the list on the Insights page, but chose to focus on the UX on consoles.
But we thought it also works fine on desktop.

Would be in a seperate iteration if we want to change it.

manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
```diff
@@ -30,31 +31,68 @@ public DashboardResourceImpl(TimerService timerService, ManagerIdentityService i
...
         try {
+            // if not authenticated, or having no INSIGHTS access, only fetch public dashboards
+            if(isAuthenticated()) {
+                publicOnly = (!hasResourceRole(ClientRole.READ_INSIGHTS.getValue(), Constants.KEYCLOAK_CLIENT_ID));
+            } else {
+                publicOnly = true;
+            }
             if(!isRealmActiveAndAccessible(realm)) {
-                throw new WebApplicationException(FORBIDDEN);
+                publicOnly = true;
```
richturner, line 46: If a realm is inactive then nothing should be retrievable from that realm

manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
```diff
@@ -30,31 +31,68 @@ public DashboardResourceImpl(TimerService timerService, ManagerIdentityService i
...
+        boolean publicOnly;
+        if(realm == null) {
+            throw new WebApplicationException(BAD_REQUEST);
+        }
         try {
+            // if not authenticated, or having no INSIGHTS access, only fetch public dashboards
+            if(isAuthenticated()) {
+                publicOnly = (!hasResourceRole(ClientRole.READ_INSIGHTS.getValue(), Constants.KEYCLOAK_CLIENT_ID));
+            } else {
+                publicOnly = true;
```
richturner, line 43: Just set `publicOnly = true` when creating the variable rather than use an else statement

manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
```diff
@@ -30,31 +31,68 @@ public DashboardResourceImpl(TimerService timerService, ManagerIdentityService i
...
+        boolean publicOnly;
+        String realm = getRequestRealmName();
+        try {
+            // if not authenticated, or having no INSIGHTS access, only take public dashboards into count
+            if(isAuthenticated()) {
+                publicOnly = (!hasResourceRole(ClientRole.READ_INSIGHTS.getValue(), Constants.KEYCLOAK_CLIENT_ID));
+            } else {
+                publicOnly = true;
+            }
+            if(!isRealmActiveAndAccessible(realm)) {
```
richturner, line 67: Same comments as above

richturner (changes requested):
Good to see tests, surprised I hadn't insisted on tests given backend code changes included ;-P

manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
MartinaeyNL: Good spot!

Just noticed as well that `isRealmActiveAndAccessible()` will always return false if you're unauthenticated.
Should I do a manual `realm.getEnabled()` check instead?

richturner:
```
 public boolean isRealmActiveAndAccessible(AuthContext authContext, Realm realm) {
      if (realm == null) {
          return false;
      }

      boolean isSuperUser = authContext != null && authContext.isSuperUser();
      boolean isUsersRealm = isSuperUser || authContext == null || authContext.isRealmAccessibleByUser(realm.getName());

      return isSuperUser || (isUsersRealm && realm.isActive(timerService.getCurrentTimeMillis()));
  }
```

This shouldn't always return false if unauthenticated?

MartinaeyNL:
Took a bit, but now it's totally ready to merge. 👍

manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
```diff
@@ -30,31 +31,66 @@ public DashboardResourceImpl(TimerService timerService, ManagerIdentityService i
...
+            throw new WebApplicationException(ex, INTERNAL_SERVER_ERROR);
         }
     }
 
     @Override
     public Dashboard get(RequestParams requestParams, String dashboardId) {
-        return this.dashboardStorageService.get(dashboardId);
+        boolean publicOnly = true;
+        String realm = getRequestRealmName();
+        try {
```
richturner, line 59: As you have a null check on realm in other methods with `BAD_REQUEST` if null it should be here also for consistency

richturner (changes requested):

richturner (approved):

PR #1073, opened 2023-06-15
Bugfix for dashboard deletion while superuser

MartinaeyNL:
Bugfix for dashboard deletion while superuser, since request is sent with 'master' realm instead of a custom one.
Instead of adding a superuser check, or checking for permission after the query, I just added a `realm` parameter to the URL.

So be aware of the minor API change:
`GET /dashboard/{id}` -> `GET /dashboard/{realm}/{id}`
`DELETE /dashboard/{id}` -> `DELETE /dashboard/{realm}/{id}`

Also added small translation fix.

### DIFF
--- a/manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
+++ b/manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
@@ -53,9 +53,8 @@ public class DashboardResourceImpl extends ManagerWebResource implements Dashboa
     }
 
     @Override
-    public Dashboard get(RequestParams requestParams, String dashboardId) {
+    public Dashboard get(RequestParams requestParams, String realm, String dashboardId) {
         boolean publicOnly = true;
-        String realm = getRequestRealmName();
         if(realm == null) {
             throw new WebApplicationException(BAD_REQUEST);
         }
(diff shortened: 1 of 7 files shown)


richturner:
Request realm should always be in the path of the URL and doesn't need manually adding on endpoints.

You have the request realm and the auth realm.

MartinaeyNL:
> Request realm should always be available as a header in all calls to the API (see `WebResource:getRequestRealmName()`)
> You have the request realm and the auth realm.

It is correctly available indeed. However in case of an superuser, `getRequestRealmName()` and `getAuthenticatedRealmName()` return `master`, in whatever realm you are present. (since the request was on `/api/master/dashboard/{id}`)
This is what was preventing dashboards to be deleted in custom realms, since the dashboard wasn't present in that realm.

As mentioned, I thought of a `isSuperUser()` check and applying an asterisk `*` for example,
but thought the easiest option was to just add the realm to the URL.

All suggestions are welcome though 😄

richturner:
OK I see the problem now; I thought everything was using the `Realm` header which is extracted from the start of the API path but like you say this is always the user's auth realm...should look at cleaning this up in the future but not for this task.

richturner (approved):

PR #1080, opened 2023-06-28
Introduced metrics instrumentation and work on load testing

richturner:


richturner:
@MartinaeyNL don't expect you to go through this in detail but a quick look over is always appreciated.

Main things are:

- Added metrics collection using Prometheus and Micrometer
- Restructured camel routes
- Removed global lock
- Improve test container re-use

manager/src/main/java/org/openremote/manager/system/StatusResourceImpl.java
```diff
@@ -20,6 +20,8 @@
 package org.openremote.manager.system;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.prometheus.client.CollectorRegistry;
+import io.prometheus.client.Counter;
```
MartinaeyNL, line 24: Might be unused imports? Don't see any usages in the file.

MartinaeyNL (commented):
Well I was actually able to understand most of it 👍 
Not much to comment on really. Only spotted the classic unused code haha

Great seeing you've been able to get rid of the global lock 😄

MartinaeyNL (approved):

PR #1082, opened 2023-06-29
Migrated to camel undertow component for websockets

richturner:


richturner:
Deleted files are our old custom websocket camel component which is now replaced with camel undertow component

Issue #1086, opened 2023-07-07
App: 'Web page not available' when app runs in background/phone closed for a while.


DonWillems:
Reproduce:
1. Open customer K app
2. Close your phone and put it away for 10 minutes
3. Open your phone. Android will show the following page. iOS shows an error dialog.
![image](https://github.com/openremote/openremote/assets/14177035/26065118-a4fe-40af-85bb-d844b1e454f5)

Expected:
3. Open your phone. The app continues as it was.

Is this resolved with an offline token?

MartinaeyNL:
Put a fix in place in PR #1089.

Basically what was happening was:
When running the app in the background, and opening it again, it would prompt a login screen since the token was expired.

With this PR, it does a check beforehand whether keycloak is accessible, to make sure the user can
safely redirect to that login page. (preventing a "no internet available" error from appearing)
If that is not the case, it will fallback to the default mechanism of requesting a token every 30 seconds.
So, when launching the app, it takes around 30 seconds to show all data correctly.

This is where we should take next steps.
- Maybe lower the interval during a disconnect, similar to websocket?
- To prevent incorrect data being shown, show a "No internet connection" placeholder page instead?
(as @richturner mentioned in calls earlier)

MartinaeyNL:
Correction: it requests the token every 10 seconds. Don't think it's necessary to lower that interval even more.

PR #1089, opened 2023-07-11
Fix for redirect to keycloak when offline or no connection is available.

MartinaeyNL:
### Summary
This PR includes a first fix for #1086 where native android/iOS errors were shown when requesting a new token.
Basically:
- When no internet is available, after around 30 seconds it would prompt a login screen since the token was expired.
- When running the app in the background, and opening it again, it would prompt a login screen since the token was expired.

.
### The fix in place
Currently we listen for any errors that happen during the `updateKeycloakAccessToken()`,
and when one appeared we redirect the user to login again.

Now, it does a http check beforehand by checking whether keycloak is accessible (by making a http request to keycloak),
to make sure the user can safely redirect to that login page. (preventing a "no internet available" error from appearing)

Since we have an interval for requesting a token, it will automatically reauthenticate when the connection is back online.
Same counts for the token expiring in the background. It might take ~~30~~ 10 seconds, which needs tweaking in the future.
See the issue #1086 for more details / the next steps on this.

.


MartinaeyNL:
Correction: it requests the token every 10 seconds. Don't think it's necessary to lower that interval even more.

ui/component/core/src/index.ts
```diff
@@ -40,6 +40,8 @@ export declare type Keycloak = {
     resourceAccess: any;
     onAuthSuccess: () => void;
     onAuthError: () => void;
+    onAuthRefreshSuccess: () => void;
+    onAuthRefreshError: () => void;
```
richturner, line 44: What are these for?

MartinaeyNL: It's a callback event of the Keycloak js adapter, which I added to the `Keycloak` type we declare in this file.
(since the adapter doesn't support types out of the box)

FYI: The `_keycloak` object is `protected` here, so these methods are only used in the core module.

richturner: Ah yes I should have expanded the whole file :)

richturner (approved):

PR #1090, opened 2023-07-12
Added offline fallback page, with configurable override per app

MartinaeyNL:
### Changelog:
- Replaced `EVENTS_CONNECTED` and `EVENTS_DISCONNECTED` events with more generic `ONLINE` and `OFFLINE` events.
- `offline` field in `AppState` is now used, and is updated when one of above events trigger.
- Made generic `_attemptReconnect()` reconnect logic, to include other disconnect causes in the future. When all of them succeed to connect, the connection status will be set back to `ONLINE`.
- Changed page loading logic to support overlay/error pages like the offline page. It will keep the regular page loaded in the background, so input fields won't get cleared when an error appears. Navigation is still accessible, and will keep track of whether pages are loaded during an 'offline period', so they can get recreated when you are back online.
.
### Offline page specific:
- Created offline placeholder page (`page-offline`) that overrides the router when `offline` in `AppState` is true.
The page has a default message and a countdown on when a new reconnect attempt is made.
- Added `offlinePage` override to `AppConfig`, that can be customized by apps.
- Added `allowOffline` field to `PageProvider`, to allow pages to be available during offline/disconnect.
.


richturner:
`AUTH_REFRESH_FAILED` and `AUTH_REFRESH_SUCCESS` is too specific and `ONLINE/OFFLINE` should be used

`keycloak` and `manager` are separate services and either could be offline yet both result in the app being un-usable; capturing both (when websocket ping/pong implemented) as `ONLINE/OFFLINE` seems reasonable.

Only reason to use `AUTH_REFRESH_x` events would be if the app would do something different in that scenario to `MANAGER_OFFLINE` scenario

manager/src/web/shared/locales/nl/or.json
```diff
@@ -322,6 +322,9 @@
   "host": "Hostname",
   "errorOccurred": "Er is een fout opgetreden",
   "sendFailed": "Verzenden mislukt",
+  "youAreOffline": "Geen internet.",
```
richturner, line 325: Not sure about fullstop included in these translation strings

ui/component/core/src/index.ts
```diff
@@ -69,7 +69,9 @@ export enum OREvent {
     EVENTS_DISCONNECTED = "EVENTS_DISCONNECTED",
```
richturner, line 69: As mentioned in my comment `ONLINE/OFFLINE` events would be more appropriate and EVENTS_DISCONNECTED/EVENTS_CONNECTED could be replaced with `ONLINE/OFFLILNE` also

richturner: Can always console log what exactly fails (manager or keycloak)

ui/component/core/src/index.ts
```diff
@@ -891,13 +893,18 @@ export class Manager implements EventProviderFactory {
                             this._keycloak!.clearToken();
                             this._keycloak!.login();
                         } else {
-                            console.error("Something went wrong reaching the keycloak server. Not redirecting.")
+                            console.error("Something went wrong reaching the keycloak server. Not redirecting.");
+                            this._emitEvent(OREvent.AUTH_REFRESH_FAILED);
                         }
                     }).catch(() => {
                         console.error("Could not reach keycloak server. Not redirecting.")
+                        this._emitEvent(OREvent.AUTH_REFRESH_FAILED);
```
richturner, line 901: console errors should be consistent

ui/component/core/src/index.ts
```diff
@@ -891,13 +893,18 @@ export class Manager implements EventProviderFactory {
                             this._keycloak!.clearToken();
                             this._keycloak!.login();
                         } else {
-                            console.error("Something went wrong reaching the keycloak server. Not redirecting.")
+                            console.error("Something went wrong reaching the keycloak server. Not redirecting.");
+                            this._emitEvent(OREvent.AUTH_REFRESH_FAILED);
```
richturner, line 897: Event should be emitted as soon as `onAuthRefreshError` is called not wait for fetch to timeout or fail

ui/component/or-app/src/index.ts
```diff
@@ -241,6 +241,21 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
 
                 this._initialised = true;
 
+                // Register listeners to change global state based on certain events
+                this._eventCallback = (event) => {
```
richturner, line 245: Don't understand why this is a field rather than a regular method?

ui/component/or-app/src/index.ts
```diff
@@ -283,15 +298,27 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
...
                     const pageProvider = this.appConfig!.pages.find((page) => page.name === this._page);
                     if (pageProvider) {
-                        const pageElem = pageProvider.pageCreator();
+                        let pageElem;
+                        if(!this._offline || pageProvider.allowOffline) {
+                            pageElem = pageProvider.pageCreator();
+                        } else {
+                            // Check for override offline page
+                            if(this.appConfig?.offlinePage) {
+                                pageElem = this.appConfig.offlinePage;
```
richturner, line 317: Does custom offlinePage not need access to the `store`?

ui/component/or-app/src/or-offline.ts
```diff
@@ -0,0 +1,107 @@
...
+
+async function* countDown(count: number) {
+    while (count > 0) {
+        yield count--;
+        await new Promise((r) => setTimeout(r, 1000));
+    }
+}
+
+
+@customElement("or-offline")
```
richturner, line 52: Maybe put page in the name of this, given webcomponent names must be globally unique the more unique it is the better chance of avoiding conflict with a custom component

ui/component/or-app/src/index.ts
```diff
@@ -283,15 +298,27 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
...
             if (this._mainElem) {
                 if (this._mainElem.firstElementChild) {
                     this._mainElem.firstElementChild.remove();
                 }
                 if (this._page) {
                     const pageProvider = this.appConfig!.pages.find((page) => page.name === this._page);
                     if (pageProvider) {
-                        const pageElem = pageProvider.pageCreator();
+                        let pageElem;
+                        if(!this._offline || pageProvider.allowOffline) {
```
richturner, line 312: If already on a page that allows offline then it will be recreated when offline changes which is probably undesirable esp. if that page has input fields etc.

ui/component/or-app/src/index.ts
```diff
@@ -283,15 +298,27 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
             return;
         }
 
-        if (changedProps.has("_page")) {
+        // If either page or 'offline'-status is changed, it should update to the correct page,
+        // by appending the page to the HTML content
+        if (changedProps.has("_page") || changedProps.has("_offline")) {
```
richturner, line 303: The logic looks sound but I haven't tried it out and not sure if there are some edge cases that might cause problems

ui/component/or-app/src/index.ts
MartinaeyNL: No, it only listens to manager events (`OrEvent`).
But might be in the future, so I changed `page-offline` to a `PageProvider`.

richturner (approved):

PR #1105, opened 2023-08-21
Enhancement/data table

NinaBianca:
Addition of optional checkmarks and sorting header in or-mwc-table.ts


ui/component/or-mwc-components/src/or-mwc-table.ts
```diff
@@ -279,19 +312,30 @@ export class OrMwcTable extends LitElement {
                                                 const content: (string | number | TemplateResult)[] | undefined = (Array.isArray(item) ? item : (item as TableRow).content);
                                                 return html`
                                                     <tr class="mdc-data-table__row"
-                                                        @click="${(ev: MouseEvent) => this.dispatchEvent(new OrMwcTableRowClickEvent((this.rows as any[]).indexOf(item)))}">
+                                                        ${this.config.multiSelect ? '' : html` @click="${(ev: MouseEvent) => this.dispatchEvent(new OrMwcTableRowClickEvent((this.rows as any[]).indexOf(item)))}"`}>
```
MartinaeyNL, line 315: Should the click be removed when `multiSelect` is true?

MartinaeyNL (changes requested):

MartinaeyNL (dismissed):

DonWillems:
- Sort arrow should have app color: --or-icon-fill: var(--or-app-color3);
- Header is clickable: sort class changes, but sorting is not applied. It does work when the text is clicked. Make it so that sorting happens on full header cell click.
- I didn't know how to use the selectedRows, so I could test that with e.g. a button on the page that logs the selected rows. Help!
- What is "columnFilter" in the TableConfig? Doesn't seem to do anything? @MartinaeyNL 
- There is a lost console.log in there
- With multiSelect enabled, the padding on the header cells changes and it doesn't align with the content of the cells anymore.

MartinaeyNL:
> * What is "columnFilter" in the TableConfig? Doesn't seem to do anything? @MartinaeyNL 

It indeed is unused currently. Was originally added in 2020, so might be removed functionality since then.
Also don't know what it was for unfortunately.

DonWillems (approved):

MartinaeyNL:
Merged. Code also looks fine enough to me.

PR #1108, opened 2023-08-29
Bugfix for appearance logo upload

MartinaeyNL:
Fix for a bug where you couldn't upload any images on the appearance page.
It did work for the `default` realm config, but not for any other realm.

After some updates on the appearance page several weeks ago,
I forgot to add a `forEach()` loop for every individual realm panel. (since I changed the HTML structure of it)
Fixed it now 👍 

*Accidentally included commits of a different branch. Reverted those in the "Reset or-mwc-table to current ..." commit.*


MartinaeyNL:
Will make it a PR after testing on a deployment by @pierrekil

richturner (approved):

PR #1118, opened 2023-10-09
Allow identity provider to set email as username

DonWillems:
Enhancement of users page to handle email as username when set by identity providers.
Exception made for service users, as they won't have an email address.

### DIFF
--- a/ui/app/manager/src/pages/page-users.ts
+++ b/ui/app/manager/src/pages/page-users.ts
@@ -207,6 +207,8 @@ export class PageUsers extends Page<AppStateKeyed> {
     protected _roles: Role[] = [];
     @state()
     protected _realmRoles: Role[] = [];
+    @state()
+    protected _registrationEmailAsUsername: boolean = false;
 
     @state()
     protected _compositeRoles: Role[] = [];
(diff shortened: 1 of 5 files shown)


richturner:
Made a couple of minor updates to avoid expensive (slow) calls in `ManagerKeycloakIdentityProvider`

ui/app/manager/src/pages/page-users.ts
```diff
@@ -314,20 +316,22 @@ export class PageUsers extends Page<AppStateKeyed> {
...
-            return;
+        if ((this._registrationEmailAsUsername && !user.serviceAccount) ? !user.email : !user.username) {
+            showSnackbar(undefined, i18next.t((this._registrationEmailAsUsername && !user.serviceAccount) ? "noEmailSet" : "noUsernameSet"), i18next.t("dismiss"));
+            return false;
         }
 
         if (user.password === "") {
             // Means a validation failure shouldn't get here
-            return;
+            return false;
```
richturner, line 334: Not sure why you're returning a value as the function doesn't declare a return value

richturner (changes requested):

ui/app/manager/src/pages/page-users.ts
DonWillems: Will add a declaration. The returned value is used to show a snackbar and return to the main users page.

richturner (approved):

PR #1147, opened 2023-10-27
Fleet Management Integration, Teltonika Telematics Device Support

pankalog:
This is still missing tests, which I plan to implement this week. 

The UI folder will be removed from the PR before merging, it is just useful for debugging purposes.


pankalog:
There are various notices that need to be given to all that peer into this pull request;

First, there is no tests here because I am not sure what the testing strategy is. Once I recognize exactly what is expected I will of course follow up with a proper testing strategy and structure. 

Second, the current state of this PR is not going to be mergeable for the above reasons. I am not expecting an all-clear, but I am expecting comments about what has already been done, and what needs to yet be done. 

OpenRemote is a very vast project with a lot of design choices and I am sure I have not made every right choice. I feel that a lot of code might be misplaced, re-written, or incorrectly used. I would like to mostly focus on issues like that.

I am also not sure if FMC003.json should or should not be included within the OpenRemote package. 

Can't wait for your comments!

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+
+    /**
+     * Checks if the Subscribing client should be allowed to subscribe to the topic that is handled by this Handler.
+     * For Teltonika device endpoints, we need the fourth token (Index 3) to be a valid IMEI number.
+     * We do that by checking using IMEIValidator.
+     */
+    // To be removed when auto-provisioning works
+    @Override
+    public boolean canSubscribe(RemotingConnection connection, KeycloakSecurityContext securityContext, Topic topic) {
+        if(topic.getTokens().size() < 5){
```
MartinaeyNL, line 213: In other MQTT implementations I see the use of `topicMatches()` override.
Maybe worth practicing here too?

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+                        TOKEN_TELTONIKA_DEVICE + "/" + TOKEN_SINGLE_LEVEL_WILDCARD + "/" + TELTONIKA_DEVICE_RECEIVE_TOPIC,
+                TOKEN_SINGLE_LEVEL_WILDCARD + "/" + TOKEN_SINGLE_LEVEL_WILDCARD + "/" +
+                        TOKEN_TELTONIKA_DEVICE + "/" + TOKEN_SINGLE_LEVEL_WILDCARD + "/" + TELTONIKA_DEVICE_SEND_TOPIC
+        );
+    }
+
+    @Override
+    public void onPublish(RemotingConnection connection, Topic topic, ByteBuf body) {
+        String payloadContent = body.toString(StandardCharsets.UTF_8);
+        getLogger().info(payloadContent);
```
MartinaeyNL, line 281: Remove these logging statements before merging.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+        return Set.of(
+                TOKEN_SINGLE_LEVEL_WILDCARD + "/" + TOKEN_SINGLE_LEVEL_WILDCARD + "/" +
+                        TOKEN_TELTONIKA_DEVICE + "/" + TOKEN_SINGLE_LEVEL_WILDCARD + "/" + TELTONIKA_DEVICE_RECEIVE_TOPIC,
+                TOKEN_SINGLE_LEVEL_WILDCARD + "/" + TOKEN_SINGLE_LEVEL_WILDCARD + "/" +
+                        TOKEN_TELTONIKA_DEVICE + "/" + TOKEN_SINGLE_LEVEL_WILDCARD + "/" + TELTONIKA_DEVICE_SEND_TOPIC
+        );
+    }
+
+    @Override
+    public void onPublish(RemotingConnection connection, Topic topic, ByteBuf body) {
```
MartinaeyNL, line 279: Please add some comments / context to this method.
I'm not fully known with the use case, but for me it's weird to create assets during `onPublish()` ?
Especially since no actual payload is published?

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+     * //TODO: Add {@value DEVICE_SEND_COMMAND_ATTRIBUTE_NAME }, {@value DEVICE_RECEIVE_COMMAND_ATTRIBUTE_NAME} Attributes when creating Asset
+     *
+     * @param newDeviceId The ID of the device's Asset.
+     * @param newDeviceImei The IMEI of the device. If passed to
+     *                      {@link UniqueIdentifierGenerator#generateId(String)},
+     *                      it should always return {@code newDeviceId}.
+     * @param realm The realm to create the Asset in.
+     * @param attributes The attributes to insert in the Asset.
+     */
+    private void CreateNewAsset(String newDeviceId, String newDeviceImei, String realm, AttributeMap attributes) {
```
MartinaeyNL, line 372: We practice lowerCamelCase methods ;)

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+        //TODO: Maybe move these to getAttributes?
+        attributes.add(new Attribute<>("IMEI", ValueType.TEXT, newDeviceImei));
+        attributes.add(new Attribute<>("lastContact", ValueType.DATE_AND_TIME,
+                new Date(timerService.getCurrentTimeMillis())));
+
+        testAsset.addOrReplaceAttributes(attributes.stream().toArray(Attribute[]::new));
+        Asset<?> mergedAsset = assetStorageService.merge(testAsset);
+
+        if(mergedAsset != null){
+            getLogger().info("Created Asset through Teltonika: " + testAsset);
```
MartinaeyNL, line 388: Be hesitant when adding log statements. They shouldn't be overdone let's say

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+
+            getLogger().info("Parsed "+paramArray.length+" Teltonika Parameters");
+
+            // Add each element to the HashMap, with the key being the unique parameter ID and the parameter
+            // being the value
+            for (TeltonikaParameter param : paramArray) {
+                params.put(param.getPropertyId(), param);
+            }
+        } catch (Exception e) {
+            getLogger().info(e.toString());
```
MartinaeyNL, line 417: `info()` instead of `severe()` ?

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+        try {
+            // Parse file with Parameter details
+            TeltonikaParameter[] paramArray = mapper.readValue(getParameterFileString(), TeltonikaParameter[].class);
+
+            getLogger().info("Parsed "+paramArray.length+" Teltonika Parameters");
+
+            // Add each element to the HashMap, with the key being the unique parameter ID and the parameter
+            // being the value
+            for (TeltonikaParameter param : paramArray) {
+                params.put(param.getPropertyId(), param);
```
MartinaeyNL, line 414: There is probably an more efficient / elegant solution to this. Java streams?

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+                throw e;
+            }
+            return attributeMap;
+        } catch (Exception e) {
+            //If the payload wasn't parsed, then it means that it is either a response to a command,
+            //or it's genuinely a wrong payload.
+
+            mapper = new ObjectMapper();
+            TeltonikaResponsePayload response = mapper.readValue(payloadContent, TeltonikaResponsePayload.class);
+            getLogger().info(response.rsp);
```
MartinaeyNL, line 440: Remove this before merging.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+        ArrayList<AssetDatapoint> list = new ArrayList<>(valueDatapoints);
+
+        // If there are no historical data found, add some first
+        if(list.isEmpty()) return Optional.empty();
+
+        //What we do now is, we will try to figure out the latest datapoint where the predicate fails,
+        //before the newValue.
+        //This means that, the state change took place between the datapoint we just found and its next one.
+
+        //Find the first datapoint that passes the negated predicate
```
MartinaeyNL, line 495: Just as a practice; I would be hesitant in creating lots of multiline comments. Just for readability you know haha

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+
+        AssetDatapoint StateChangeAssetDatapoint = null;
```
MartinaeyNL, line 497: We practice lowerCamelCase variables ;)

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+                    break;
+                }
+            }
+
+            if (StateChangeAssetDatapoint != null){
+                if (!pred.value.asPredicate(timerService::getCurrentTimeMillis).test(StateChangeAssetDatapoint.getValue())){
+                    throw new Exception("Found state change datapoint failed predicate");
+                }
+            }else{
+                throw new Exception("Couldn't find asset state change value");
```
MartinaeyNL, line 526: I know it's being catched by line 529, but throwing a more specific exception type helps readability / future cases.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+            }
+
+        }catch (Exception e){
+            getLogger().warning(e.getMessage());
+            return Optional.empty();
+        }
+
+        //BUG: The Timestamp of an asset datapoint is NOT attribute.getTimestamp()!
+        //This makes it much harder to relate datapoints to relate a duration to the time-distance between two datapoints.
+        //To make this easier, we are going to use some time-padding of 1 second from the gap. It's going to be sufficient to cover some edge-cases.
```
MartinaeyNL, line 536: Hmmmm interesting finding! Don't know if we can change this in the datapoint code.
You can create a GitHub issue for this. I guess you can also assign it to me.

Don't know how large the gap is between an datapoint timestamp and an attributeChanged timestamp?

I would hesitate to practice a "margin of error" of 1 second.
Imagine the system being slow, taking for example 500+ ms to handle an attribute change, it could hurt this.
Maybe something like querying the datapoint with the closest timestamp and the same value would work?

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+
+        //The AssetStateDuration will have a startTime of the found state-change variable, and the endDate will be the previousValue's timestamp, that being the end of the state duration
+        Attribute<?> tripAttr = new Attribute<>("LastTripStartedAndEndedAt", ValueType.ASSET_STATE_DURATION, new AssetStateDuration(
+                new Timestamp(StateChangeAssetDatapoint.getTimestamp()),
+                new Timestamp(previousValue.getTimestamp().get())
+        ));
+
+        tripAttr.setTimestamp(StateChangeAssetDatapoint.getTimestamp());
```
MartinaeyNL, line 544: ... I guess this works too. (see previous comment)

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+     * @param attributes The attributes to be upserted to the Attribute.
+     * @param topic The topic to which the MQTT payload was sent.
+     * @param connection The connection on which the payload was sent.
+     */
+    private void UpdateAsset(Asset<?> asset, AttributeMap attributes, Topic topic, RemotingConnection connection) {
+
+        String imei = asset.getAttribute("IMEI").toString();
+
+        getLogger().info("Updating CarAsset with IMEI "+imei);
+        //DONE: If the parameters do not exist, then the update fails,
```
MartinaeyNL, line 572: Remove these notes when merging I guess? Don't know how important they are?

It's also a kind of general practice to not use too many multiline comments in methods.
Placing a large comment above the function describing it is indeed a good habit,
but in the method itself it should only shortly describe its meaning and possible behavior.

test/src/test/groovy/org/openremote/test/protocol/mqtt/TeltonikaMQTTProtocolTest.groovy
```diff
@@ -0,0 +1,284 @@
...
+
+    def setup(){
+//        I think that it is required to provide usernamePassword to connect the client to the broker. Look at AbstractMQTT_IOClient:L100.
+        client.removeAllMessageConsumers();
+        client.disconnect();
+
+        client = new MQTT_IOClient(mqttClientId, mqttHost, mqttPort, false, false, null, null, null)
+    }
+
+    def cleanupSpec(){
```
MartinaeyNL, line 78: I think cleanup methods shouldn't be necessary. See `MqttBrokerTest` for inspiration.

MartinaeyNL:
For context;

@pankalog asked me for a review of his draft.
I myself don't really have much knowledge of the MQTTHandler, neither much of MQTT (other than the basics)
Will need additional review when it's in a ready state.

MartinaeyNL:
To respond on your comment @pankalog 

In general I believe architecture-wise you did a great job in implementing your own MQTTHandler.
And you did a good effort in providing tests to cover your additional additions.

Of course I can't fully agree nor disagree with architecture things, since @richturner is more experienced with the backend architecture than I am. But in general it looks good!

Let me know if you need anything 😉

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+
+
+    //Sending a message to master/teltonikaDevice1/teltonika/864636060373301/commands
+    //with payload {"CMD":"getstatus"} seems to work. Maybe the quotes are the issue?
+    private void handleAttributeMessage(AttributeEvent event) {
+        // If this is not an AttributeEvent that updates a sendToDevice field, ignore
+        if (!Objects.equals(event.getAttributeName(), DEVICE_SEND_COMMAND_ATTRIBUTE_NAME)) return;
+        getLogger().info(event.getEventType());
+        //Find the asset in question
+        Asset<?> asset = assetStorageService.find(new AssetQuery().ids(event.getAssetId()).types(CarAsset.class));
```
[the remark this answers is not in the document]
pankalog: That's a great method, I need to refactor everything to use it in a lot of places now

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+
+    /**
+     * Overrides MQTTHandler.checkCanPublish for this specific Handler,
+     * until secure Authentication and Auto-provisioning
+     * of Teltonika Devices is created.
+     * To be removed after implementation is complete.
+     */
+    @Override
+    public boolean checkCanPublish(RemotingConnection connection, KeycloakSecurityContext securityContext, Topic topic) {
+        return true;
```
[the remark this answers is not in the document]
pankalog: The reason I'm currently overriding this is because of avoiding Keycloak authentication, something currently not possible for Teltonika devices. This will be removed when authentication for the devices is properly implemented

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,639 @@
...
+    }
+
+    @Override
+    public boolean canPublish(RemotingConnection connection, KeycloakSecurityContext securityContext, Topic topic) {
+        getLogger().info("Teltonika device will publish to Topic "+topic.toString()+" to transmit payload");
+        return true;
```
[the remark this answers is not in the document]
pankalog: Removed all Keycloak usages, especially in TeltonikaMQTTHandler#start, since Keycloak is currently not required for any function of this Handler.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Fixed everywhere in TeltonikaMQTTHandler

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Added better verbosity

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Implemented using Java streams

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Removed

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Removed a lot of logs from TeltonikaMQTTHandler

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Maximum of 2 lines per comment, from now on, I hope nobody minds!

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Renamed all functions

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Of course the margin of error was added at least so that I can get my data in, because without it I lost a lot of datapoints. It's not optimal in any way. 

Lets continue this in #1156.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Still does not solve the issue, it's not a great way to solve this because I also get some "false" trips

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Removed a lot of them, but I am keeping the TODOs until I have a very stable (or merge-able) version. When I do, this will be converted to proper entries in a Kanban board

model/src/main/java/org/openremote/model/asset/impl/CityAsset.java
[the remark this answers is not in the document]
pankalog: Reverted

model/src/main/java/org/openremote/model/teltonika/IMEIValidator.java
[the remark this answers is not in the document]
pankalog: I couldn't find a validator for IMEIs within OpenRemote, and I wanted to use a fully-implemented validator with checksums.

test/src/testFixtures/groovy/org/openremote/test/ContainerTrait.groovy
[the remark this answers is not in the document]
pankalog: Reverted

model/src/main/java/org/openremote/model/value/ValueType.java
```diff
@@ -130,6 +131,8 @@ public static class MultivaluedStringMap extends HashMap<String, List<String>> {
         new ValueConstraint.Pattern(Constants.ISO8601_DURATION_REGEXP)
     );
 
+    public static final ValueDescriptor<AssetStateDuration> ASSET_STATE_DURATION = new ValueDescriptor<>("AssetStateDuration", AssetStateDuration.class);
+
```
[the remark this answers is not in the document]
pankalog: We can talk about renaming at #1146

test/src/test/groovy/org/openremote/test/protocol/mqtt/TeltonikaMQTTProtocolTest.groovy
pankalog: I think they are because I use multiple tests for my groovy tests, and they just help with keeping things neat I believe.

pankalog:
I just went through @MartinaeyNL 's review comments and I resolved as many as I can. The rest of them cannot really be resolved right now without further approval/definition of what needs to do. Waiting for further reviews!

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,627 @@
...
+
+import static org.openremote.manager.event.ClientEventService.CLIENT_INBOUND_QUEUE;
+import static org.openremote.manager.mqtt.DefaultMQTTHandler.prepareHeaders;
+import static org.openremote.model.syslog.SyslogCategory.API;
+import static org.openremote.model.value.MetaItemType.*;
+
+public class TeltonikaMQTTHandler extends MQTTHandler {
+
+    //TODO: Maybe move this to Models?
+    private static class TeltonikaDevice {
```
richturner, line 53: Prefer `protected` as it easily allows others to subclass your code and make changes.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,627 @@
...
+
+    @Override
+    public void start(Container container) throws Exception {
+        super.start(container);
+        getLogger().info("Starting Teltonika MQTT Handler");
+        ManagerIdentityService identityService = container.getService(ManagerIdentityService.class);
+        assetStorageService = container.getService(AssetStorageService.class);
+        AssetDatapointService = container.getService(AssetDatapointService.class);
+        timerService = container.getService(TimerService.class);
+        DeviceParameterPath = Paths.get("deployment/manager/fleet/FMC003.json");
```
richturner, line 100: This could be baked into the resources of the module unless a user needs to change it.

This relative path won't work when image is packaged into docker.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,627 @@
...
+        } else {
+            isKeycloak = true;
+            identityProvider = (ManagerKeycloakIdentityProvider) identityService.getIdentityProvider();
+        }
+        //TODO: Figure out a way to update the AssetFilter when a new asset is a candidate to receive a response message
+        // Does it even automatically refresh the Asset list?
+
+        clientEventService.addInternalSubscription(
+                AttributeEvent.class,
+                buildAssetFilter(),
```
richturner, line 113: `AssetFilter` is a bit limited in its' capability I would suggest using `null` and then filtering the incoming `AttributeEvents` for asset type car etc.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,627 @@
...
+
+        AssetFilter<AttributeEvent> event = new AssetFilter<>();
+        event.setAssetIds(listOfCarAssetIds.toArray(new String[0]));
+        event.setAttributeNames(TELTONIKA_DEVICE_SEND_COMMAND_ATTRIBUTE_NAME);
+        return event;
+    }
+
+    private void handleAttributeMessage(AttributeEvent event) {
+        // If this is not an AttributeEvent that updates a TELTONIKA_DEVICE_SEND_COMMAND_ATTRIBUTE_NAME field, ignore
+        if (!Objects.equals(event.getAttributeName(), TELTONIKA_DEVICE_SEND_COMMAND_ATTRIBUTE_NAME)) return;
```
richturner, line 139: Check asset type of event also

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,627 @@
...
+        Map<String, Object> headers = prepareHeaders(topicRealm(topic), connection);
+
+        AttributeMap nonExistingAttributes = new AttributeMap();
+
+        attributes.forEach( attribute ->  {
+            //Attribute exists, needs to be updated
+            if(asset.getAttributes().containsKey(attribute.getName())){
+                AttributeEvent attributeEvent = new AttributeEvent(asset.getId(), attribute.getName(), attribute.getValue(), packetTimestamp);
+                LOG.finer("Publishing to client inbound queue: " + attributeEvent);
+                messageBrokerService.getFluentProducerTemplate()
```
richturner, line 612: You can use `AssetProcessingService.sendAttributeEvent()`

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,627 @@
...
+    /**
+     * Sends a Command to the {@link TeltonikaDevice} in the correct format.
+     *
+     * @param command string of the command, without preformatting.
+     * List of valid commands can be found in Teltonika's website.
+     * @param device A {@link TeltonikaDevice} that is currently subscribed, to which to send the message to.
+     */
+    private void sendCommandToTeltonikaDevice(String command, TeltonikaDevice device) {
+        HashMap<String, String> cmdtest = new HashMap<>();
+        cmdtest.put("CMD", command);
```
richturner, line 190: `Map.of` is a convenient shorthand

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,627 @@
...
+     * To be removed after implementation is complete.
+     */
+    @Override
+    public boolean checkCanPublish(RemotingConnection connection, KeycloakSecurityContext securityContext, Topic topic) {
+        return canPublish(connection,securityContext, topic);
+    }
+
+    @Override
+    public boolean canPublish(RemotingConnection connection, KeycloakSecurityContext securityContext, Topic topic) {
+        getLogger().info("Teltonika device will publish to Topic "+topic.toString()+" to transmit payload");
```
richturner, line 258: Log level should be lower as this will be hit on every device publish

model/src/main/java/org/openremote/model/asset/impl/CarAsset.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Date;
+import java.util.Optional;
+@Entity
+public class CarAsset extends Asset<CarAsset> {
+    public static final AttributeDescriptor<String> IMEI = new AttributeDescriptor<>("IMEI", ValueType.TEXT);
+    public static final AttributeDescriptor<Date> LAST_CONTACT = new AttributeDescriptor<>("lastContact", ValueType.DATE_AND_TIME);
+    public static final AttributeDescriptor<String> MAKE_AND_MODEL = new AttributeDescriptor<>("makeAndModel", ValueType.TEXT).withOptional(true);
+    public static final AttributeDescriptor<Integer> MODEL_YEAR = new AttributeDescriptor<>("modelYear", ValueType.INTEGER).withOptional(true);
+    /**
+     * If Rich sees this, sorry for the American spelling!
```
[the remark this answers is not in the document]
richturner: That's it no more assistance from me

model/src/main/java/org/openremote/model/teltonika/IMEIValidator.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Class that validates an IMEI value.
+ * Employs checksum as outlined by the GSM Association.
+ * Class retrieved from <a href="https://www.geeksforgeeks.org/program-check-valid-imei-number/">GeeksForGeeks</a>.
+ *
+ * @see <a href="https://www.gsma.com/newsroom/wp-content/uploads//TS.06-v16.0.pdf">the GSMA IMEI Allocation Guidelines</a>.
+ *
+ */
+public class IMEIValidator {
```
richturner, line 11: A protected constructor will prevent instances of this helper class

model/src/main/java/org/openremote/model/teltonika/PayloadJsonObject.java
```diff
@@ -0,0 +1,49 @@
...
+        return this.additionalProperties;
+    }
+
+    @JsonAnySetter
+    public void setAdditionalProperty(String name, Object value) {
+        this.additionalProperties.put(name, value);
+    }
+
+    @Override
+    public String toString() {
```
richturner, line 30: Is this for logging purposes?

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.*;
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.openremote.model.value.MetaItemType.*;
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
richturner, line 27: This is default for our Jackson `ObjectMapper`

model/src/main/java/org/openremote/model/teltonika/PayloadJsonObject.java
```diff
@@ -0,0 +1,49 @@
+package org.openremote.model.teltonika;
+
+import com.fasterxml.jackson.annotation.*;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({
```
richturner, line 9: key ordering in JSON is not guaranteed plus you only have one property being serialised

model/src/main/java/org/openremote/model/teltonika/PayloadJsonObject.java
```diff
@@ -0,0 +1,49 @@
+package org.openremote.model.teltonika;
+
+import com.fasterxml.jackson.annotation.*;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
richturner, line 8: This is default behaviour for our Jackson `ObjectMapper`

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+public class State {
+
+//    public ReportedState reportedState;
+    @JsonProperty("reported")
+    private Map<String, Object> reported;
+
+
+
+    @Override
+    public String toString() {
```
richturner, line 40: Is this just for use in logging?

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+        return sb.toString();
+    }
+
+
+    //TODO: Add real timestamp reported by device, check issue #1156
+    //TODO: Add description of each attribute (No idea where to put that yet)
+    //TODO: Move any function from here; it's a POJO.
+    //TODO: Figure out what the parameters: pr, alt, ang, sat, sp, and evt do
+    // and implement their functionality (I can guess but I want to know exactly).
+    public AttributeMap GetAttributes(Map<Integer, TeltonikaParameter> params, AttributeMap additionalAttributes, Logger logger){
```
richturner, line 61: camel case is normal in java not pascal case

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+                parsedEntryId = Integer.parseInt(entryId);
+            }catch (Exception e){
+                continue;
+            }
+            //If the Teltonika Parameter HashMap doesn't contain the requested AVL ID, continue:
+
+            if (!params.containsKey(parsedEntryId)) continue;
+
+            //Retrieve the parameter data
+            TeltonikaParameter parameter = params.get(Integer.valueOf(entry.getKey()));
```
richturner, line 77: use `parsedEntryId` here no need to parse the `int` again

richturner, line 77: Can just check if `parameter == null`

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+                        logger.severe(parameter.propertyName + "Failed multiplier");
+                        logger.severe(e.toString());
+                        throw e;
+                    }
+                }
+
+                //TODO: Fix frontend to be able to display even custom units, not just the predefined ones
+                //possibly prepend the unit with the string "custom."? So that it matches the predefined format
+                //Add on its units
+                if(!Objects.equals(parameter.units, "-")){
```
richturner, line 127: `parameter.units` could be null or empty?

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+                logger.severe(e.toString());
+                throw e;
+            }
+            //Use the MetaMap to create an AttributeDescriptor
+            AttributeDescriptor<?> attributeDescriptor = new AttributeDescriptor<>(parameter.propertyId.toString(), attributeType, metaMap);
+
+
+
+            //Use the AttributeDescriptor and the Value to create a new Attribute
+            Attribute<?> generatedAttribute = new Attribute(attributeDescriptor, value.get());
```
richturner, line 208: `AttributeDescriptor` are for static use not runtime use; you should just use the other constructor(s):

```
Attribute(String name, ValueDescriptor<T> valueDescriptor, T value)
```

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+            // Add it to the AttributeMap
+            attributes.addOrReplace(generatedAttribute);
+        }
+        //Special parameter definitions without being defined in the AVL Parameter List, thanks Teltonika
+
+        //latlng are the latitude-longitude coordinates, also check if it's 0,0, if it is, don't update.
+        if (reported.containsKey("latlng") && !Objects.equals(reported.get("latlng"), "0.000000,0.000000")){
+            try{
```
richturner, line 216: unlikely but what if the device is at this actual location

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+            // Figure out its attributeType
+            ValueDescriptor<?> attributeType = GetAttributeType(parameter);
+
+            //Retrieve its coerced value
+            Optional<?> value;
+            try {
+                //Inner method returns Optional.empty, but still throws and prints exception. A lot of clutter, but the exception is handled.
+                value = ValueUtil.getValueCoerced(entry.getValue(), attributeType.getType());
+                if (value.isEmpty()){
+                    attributeType = ValueType.TEXT;
```
richturner, line 91: This could try to change an existing attributes type which seems odd

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+                logger.severe("Failed timestamps");
+                logger.severe(e.toString());
+                throw e;
+            }
+
+        }
+
+        // Store data points, allow use for rules, and don't allow user parameter modification, for every attribute parsed
+        try{
+            attributes.forEach(attribute -> attribute.addOrReplaceMeta(
```
richturner, line 249: Maybe this should only be done for new attributes so user can potentially decide to not store data points for an attribute by removing the configuration item

manager/src/main/java/org/openremote/manager/asset/AssetStorageService.java
```diff
@@ -604,17 +604,18 @@ public List<String> findNames(String... ids) {
...
     /**
      * @param overrideVersion If <code>true</code>, the merge will override the data in the database, independent of
      *                        version.
+     * @param preserveTimestamps    If true, the attributes being merged will be stored with their stored timestamp, given those timestamps are newer than the current asset's attributes.
      * @return The current stored asset state.
      * @throws IllegalArgumentException if the realm or parent is illegal, or other asset constraint is violated.
      */
-    public <T extends Asset<?>> T merge(T asset, boolean overrideVersion) throws IllegalStateException, ConstraintViolationException {
-        return merge(asset, overrideVersion, false, null);
+    public <T extends Asset<?>> T merge(T asset, boolean overrideVersion, boolean preserveTimestamps) throws IllegalStateException, ConstraintViolationException {
```
richturner, line 617: Don't want to do this unless really necessary would rather solve this problem without a configurable option for consistency; an overload method to maintain the original method signature should have been added

manager/src/main/java/org/openremote/manager/system/HealthService.java
```diff
@@ -63,7 +63,7 @@ public void onInit(Route route) {
 
     public static final System.Logger LOG = System.getLogger(HealthService.class.getName());
     public static final String OR_METRICS_PORT = "OR_METRICS_PORT";
-    public static final int OR_METRICS_PORT_DEFAULT = 8404;
+    public static final int OR_METRICS_PORT_DEFAULT = 8403;
```
richturner, line 66: Don't want the default changed here, you can override within your own Run configuration as needed

model/src/main/java/org/openremote/model/teltonika/State.java
richturner, line 61: additionalAttributes doesn't seem to be used?

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+                throw e;
+            }
+        }
+        //Timestamp grabbed from the device.
+        //TODO: It doesn't look like the timestamp is being parsed correctly...
+        if (reported.containsKey("ts")){
+            try{
+                long unixTimestampMillis = Long.parseLong(reported.get("ts").toString());
+                Timestamp deviceTimestamp = Timestamp.from(Instant.ofEpochMilli(unixTimestampMillis));
+                attributes.add(new Attribute<>("lastContact", ValueType.DATE_AND_TIME, deviceTimestamp).setTimestamp(deviceTimestamp.getTime()));
```
richturner, line 234: There is an `Attribute` constructor that takes timestamp

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+            logger.severe(e.toString());
+            throw e;
+        }
+
+        return attributes;
+    }
+
+
+
+    private ValueDescriptor<?> GetAttributeType(TeltonikaParameter parameter) {
```
richturner, line 266: May be better using `ValueUtil.parse()` to convert the string to the correct type and then you can pass the parsed value into here for value type extraction (to prevent parsing the value multiple times which is computationally expensive)

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+            return ValueType.NUMBER;
+        }catch (NumberFormatException e){
+            return switch (parameter.type) {
+                case "Unsigned", "Signed", "unsigned", "UNSIGNED LONG INT" -> ValueType.NUMBER;
+                default -> ValueType.TEXT;
+            };
+        }
+    }
+
+    private GeoJSONPoint ParseLatLngToGeoJSONObject(String latlngString) {
```
richturner, line 279: Make a note of the format of the data being parses

model/src/main/java/org/openremote/model/teltonika/State.java
```diff
@@ -0,0 +1,298 @@
...
+        String regexPattern = "^([-+]?[0-8]?\\d(\\.\\d+)?|90(\\.0+)?),([-+]?(1?[0-7]?[0-9](\\.\\d+)?|180(\\.0+)?))$";
```
richturner, line 280: Would a simple string split be more effecient here...regex is orders of magnitude slower than a string split

model/src/main/java/org/openremote/model/teltonika/TeltonikaDataPayload.java
```diff
@@ -0,0 +1,16 @@
+package org.openremote.model.teltonika;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({
```
richturner, line 8: Same as previous comments

model/src/main/java/org/openremote/model/teltonika/TeltonikaDataPayload.java
```diff
@@ -0,0 +1,16 @@
...
+        "payload"
+})
+public class TeltonikaDataPayload {
+
+    @JsonProperty("state")
+    public State state;
```
richturner, line 14: Fields are by default used as json properties so this isn't needed unless you want to change the key name

model/src/main/java/org/openremote/model/teltonika/TeltonikaParameter.java
```diff
@@ -0,0 +1,141 @@
...
+    @JsonProperty("multiplier")
+    public String multiplier;
+    @JsonProperty("units")
+    public String units;
+    @JsonProperty("description")
+    public String description;
+    @JsonProperty("hwSupport")
+    public String hwSupport;
+    @JsonProperty("parameterGroup")
+    public String parameterGroup;
```
richturner, line 52: Same as other comments

model/src/main/java/org/openremote/model/teltonika/TeltonikaResponsePayload.java
```diff
@@ -0,0 +1,32 @@
...
+package org.openremote.model.teltonika;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({
+        "RSP"
+})
```
richturner, line 11: Same as previous comments

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,654 @@
...
+//            Attribute<String> payloadAttribute = new Attribute<>("payload", ValueType.TEXT, payloadContent);
+//            payloadAttribute.addMeta(new MetaItem<>(MetaItemType.STORE_DATA_POINTS, true));
+//            attributes.add(payloadAttribute);
+
+
+
+
+            if (asset == null) {
+                try{
+                    createNewAsset(deviceUuid, deviceImei, realm, attributes, topic, connection);
```
richturner, line 318: Should be setting asset to the returned asset from the merge

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,654 @@
...
+        boolean previousValueTest = pred.value.asPredicate(timerService::getCurrentTimeMillis).test(previousValue   .getValue().get());
+        //If the predicate fails, then no changes need to happen.
+
+        // newValue is not 1, previousValue == 1
+        if(!(!newValueTest && previousValueTest)) {
+            return Optional.empty();
+        }
+
+        // Grab all datapoints (To be replaced by AssetDatapointValueQuery)
+        List<AssetDatapoint> valueDatapoints = AssetDatapointService.getDatapoints(ref);
```
richturner, line 517: Need a limit on this as this could be a large amount of data

richturner, line 517: It would be more efficient to keep in memory the attributes that have an active session rather than searching back through historical data on each publish

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,654 @@
...
+     */
+    private AttributeMap getAttributesFromPayload(String payloadContent) throws JsonProcessingException {
+        HashMap<Integer, TeltonikaParameter> params = new HashMap<>();
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            // Parse file with Parameter details
+
+            // Add each element to the HashMap, with the key being the unique parameter ID and the parameter
+            // being the value
+            params = Arrays.stream(mapper.readValue(getParameterFileString(), TeltonikaParameter[].class)).collect(Collectors.toMap(
```
richturner, line 435: You're reading the parameter file in for every payload

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,654 @@
...
+            if (asset.getAttribute(attribute.getName()).isPresent()) {
+                existingAttributes.add(attribute);
+            } else {
+                nonExistingAttributes.add(attribute);
+            }
+        });
+
+        //First merge, then update existing attributes
+        asset.addAttributes(nonExistingAttributes.stream().toArray(Attribute[]::new));
+        if(nonExistingAttributes.size() > 0){
```
richturner, line 633: isEmpty() method is preferred

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Could I still use the buildAssetFilter() method's returned AssetFilter using apply() to check if the event applies? This way we retain the same filter functionality but dynamically refresh the applicable asset list on every message.

pankalog: Here is what I am trying to do when entering handleAttributeMessage(): 

`AssetFilter<AttributeEvent> eventFilter = buildAssetFilter();

        if(eventFilter.apply(event) != null) {
            getLogger().finer("eventFilter applied and the event is not for me");
            return;
        }`

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: https://github.com/openremote/openremote/pull/1147#discussion_r1401988178

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Should I use `AttributeEvent.Source#SENSOR` as the `Source` in the overload of that function? Since it sort of is a sensor update, just handled by the MQTT Handler.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: set to `FINER`

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: Yes, should I remove it?

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: https://github.com/openremote/openremote/pull/1147#discussion_r1402023477

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: Since these are the values parsed from the Teltonika wiki, all of the no-values have that dash as a value. Should I do some pre-processing when importing the values to set them to proper values, and then check those values in this method?

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: The only reason I do this is because the device will send those coordinates if it isn't able to get a satellite fix. Not to mention, this coordinate is quite far out at sea, but at that point, we can think of something else. I am only doing this because I would like the timestamps of attributes to be when the actual last data of the attribute was sent, instead of when dummy data was sent (like 0,0), for other reasons. Also, this is configurable on the device-side, meaning that the user can configure if the device will send packets if there is a satellite fix or not

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: How do you think this should be handled? Should I not process this attribute or change its type?

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: Actually, by relating this to https://github.com/openremote/openremote/pull/1147#discussion_r1402049387, should I create some sort of mediary function that checks the existing attribute's attributes against the parsed ones, and then try to merge all of the metadata of the attribute? I have considered this before, but this requires a conversation of what we consider to be more important, the current attributes of the asset or the incoming ones.

manager/src/main/java/org/openremote/manager/asset/AssetStorageService.java
pankalog: This is for-sure not going to remain like this in the future, it is part of the debugging attempts at

manager/src/main/java/org/openremote/manager/system/HealthService.java
pankalog: Accidentally committed the file, but for some reason something is already hosted on 8404 at least on my PC, and I'm not exactly sure what

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: It is part of my rethinking of how I should deal with #1156 , meaning if I should pass all attributes I would like to eventually add/update to the Asset or if I should use the returned AttributeMap to append any attributes needed. How should I continue?

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: I'm not sure I understand, are you talking about checking parameter.type?

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: Added javadoc

model/src/main/java/org/openremote/model/teltonika/State.java
pankalog: I agree, but in this way, we are completely sure that parsing the two numbers is safe and will not throw an exception. Would you want me to convert this to using String.split() and check if the numbers are parsed?

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: I have actually been wanting to ask you, this would be much easier and more efficient if I could apply the predicate on values of the database. In fact, I tried implementing a new AssetDatapointQuery that would do that, but I am kinda lost on how that would look like. How do you think I should manage this?

pankalog: Would this be stored inside some sort of instance variable in TeltonikaDevice? Hence we could easily link the incoming data/state change to the device producing them. Thanks for the idea!

pankalog: But would it be okay if the state was deleted, in case OpenRemote is shut down? If it was stored in the database, we could simply check the last time the value was equal to the current value.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: We had a conversation about how we should manage the parameter files. Would we want to read it once on `TeltonikaMQTTHandler.start()`?

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
```diff
@@ -0,0 +1,654 @@
...
+     * TODO: I have a very big feeling that the way this is done is wrong.
+     *
+     * @param asset The asset to be updated.
+     * @param attributes The attributes to be upserted to the Attribute.
+     * @param topic The topic to which the MQTT payload was sent.
+     * @param connection The connection on which the payload was sent.
+     */
+    private void updateAsset(Asset<?> asset, AttributeMap attributes, Topic topic, RemotingConnection connection) {
+
+        Attribute<Date> packetTime = attributes.get(new AttributeDescriptor<Date>("lastContact", ValueType.DATE_AND_TIME)).get();
```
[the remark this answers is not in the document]
pankalog: Remnants of a lot of testing runs, there will be no Optional.get() calls without prior checking when ready.

manager/src/main/java/org/openremote/manager/mqtt/TeltonikaMQTTHandler.java
pankalog: Can you show me an example for this so that I can implement it?

model/src/main/java/org/openremote/model/util/ValueUtil.java
```diff
@@ -216,7 +216,7 @@ public static <T> Optional<T> parse(String jsonString, Type type) {
         try {
             return Optional.of(JSON.readValue(jsonString, JSON.constructType(type)));
         } catch (Exception e) {
-            LOG.log(Level.WARNING, "Failed to parse JSON", e);
+//            LOG.log(Level.WARNING, "Failed to parse JSON", e);
```
richturner, line 219: Wouldn't merge this

ui/app/manager/src/pages/page-map.ts
```diff
@@ -341,33 +370,66 @@ export class PageMap extends Page<MapStateKeyed> {
...
+                                    }
+                                })
+                                .map(asset => {
+                                    return html`
+                                        <or-map-marker-asset
+                                                ?active="${this._currentAsset && this._currentAsset.id === asset.id}"
+                                                .asset="${asset}"
+                                                .config="${this.config.markers}"></or-map-marker-asset>
+                                    `;
+                                })
```
richturner, line 427: Reformatting here makes it difficult to follow what has actually changed...we do need to enforce some linting and coding standards though

ui/app/manager/src/pages/page-map.ts
```diff
@@ -341,33 +370,66 @@ export class PageMap extends Page<MapStateKeyed> {
...
+        }
+        // currentAssetSelected? console.log("render") : console.log("asset unselected");
+            return html`
+<!--                
+                <or-map-location-history-markers .assetId="${this._currentAsset?.id}" .map="${this._map}"></or-map-location-history-markers>
+            -->
+                <or-map-location-history-overlay .assetId="${this._currentAsset?.id}" .map="${this._map}" ></or-map-location-history-overlay>
+${currentAssetSelected ? html`
+                <or-map-asset-card .config="${this.config?.card}" .assetId="${this._currentAsset.id}"
+                                   .markerconfig="${this.config?.markers}" .map="${this._map}"></or-map-asset-card>
```
richturner, line 386: This newly added map property doesn't appear to be used

ui/app/manager/src/pages/page-map.ts
```diff
@@ -341,33 +370,66 @@ export class PageMap extends Page<MapStateKeyed> {
...
+        let isMouseAPointer = false;
+        if(this._map?._map?._mapGl != undefined){
+            isMouseAPointer = this._map._map._mapGl.getCanvas().style.cursor == '';
+        }
+        // currentAssetSelected? console.log("render") : console.log("asset unselected");
+            return html`
+<!--                
+                <or-map-location-history-markers .assetId="${this._currentAsset?.id}" .map="${this._map}"></or-map-location-history-markers>
+            -->
+                <or-map-location-history-overlay .assetId="${this._currentAsset?.id}" .map="${this._map}" ></or-map-location-history-overlay>
```
richturner, line 383: If this location history is to be integrated into the generic manager app then it will need to be configurable

ui/component/or-attribute-card/src/index.ts
```diff
@@ -744,11 +744,11 @@ export class OrAttributeCard extends LitElement {
             assetId,
             attributeName,
             {
-                type: "lttb",
+                type: "all",
                 fromTimestamp: this._startOfPeriod,
                 toTimestamp: this._endOfPeriod,
                 amountOfPoints: 20,
-            } as AssetDatapointLTTBQuery
+            } as AssetDatapointAllQuery
```
richturner, line 751: Is this intentional?

ui/component/or-dashboard-builder/src/panels/attributes-panel.ts
```diff
@@ -193,7 +193,7 @@ export class AttributesPanel extends LitElement {
     }
 
     protected async loadAssets(): Promise<Asset[]> {
-        if(this.attributeRefs.filter(ar => !this.getLoadedAsset(ar)).length > 0) {
+        if(true) {
```
richturner, line 196: Remove or fix

ui/component/or-attribute-card/src/index.ts
MartinaeyNL: To additionally comment on this; we're intentionally using `LTTB` in the `or-attribute-card` here.

ui/component/or-dashboard-builder/src/util/nominatim-browser.ts
richturner, line 1: I think we can just update the forward/reverse geocoder functions in mapwidget rather than this, as it doesn't add enough functionality to warrant the copy paste and source code isn't maintained. We should definitely default to nominatim openstreetmap API though.

ui/component/or-dashboard-builder/src/widgets/session-widget.ts
richturner, line 1: Need a way of being able to plugin custom widgets at runtime from a custom project

ui/component/or-map/src/mapwidget.ts
```diff
@@ -873,6 +911,15 @@ export class MapWidget {
         }
     }
 
+    public async getGeoJsonFeature(query: {query: string}){
+        // let gc: MaplibreGeocoder. = this._geocoder;
+    }
```
richturner, line 916: Does this not work? Can it be fixed or removed or at least commented out with reason

ui/component/or-map/src/mapwidget.ts
```diff
@@ -873,6 +911,15 @@ export class MapWidget {
...
+    public async getAddress(query: {lat: number, lon: number}){
+        // let gc: MaplibreGeocoder. = this._geocoder;
+        return this._reverseGeocode(query);
+    }
+
+
```
richturner, line 922: Maybe should be exposed from the `or-map` so no dependency on the map for geocoding

ui/component/or-map/src/markers/or-map-asset-card-trip-section.ts
richturner, line 1: This needs to be a pluggable mechanism from a custom project

ui/component/or-map/src/or-map-location-history-markers.ts
```diff
@@ -0,0 +1,112 @@
...
+
+            const el = document.createElement('historymapmarker')
+            // el.style.position = "absolute"
+            // el.style.zIndex = "2";
+            // el.id = ""+this.assetId+historyElement.x;
+            // el.className = "historyMarker"
+
+            const time: string = moment(historyElement.x).toString();
+            const popup = new maplibregl.Popup().setText(
+                'Construction on the Washington Monument began in 1848.'
```
richturner, line 73: Cleanup required

ui/component/or-map/src/or-map-location-history-markers.ts
richturner, line 1: I like this as a marker type but needs to be configurable at the map level.

Can this marker not extend the OrMapMarkerAsset or even just add this functionality to that marker.

History duration to retrieve in map config (if set then uses this marker instead of asset marker)

ui/component/or-map/src/orMapLocationHistoryOverlay.ts
richturner, line 1: Not clear to me why this is needed rather than encapsulating everything in the history marker

ui/component/or-attribute-card/src/index.ts
richturner: Thanks but is it a change that makes sense for everything or just this fleet management application?

MartinaeyNL:
@pankalog Can this be closed in favor of your PR in #2571 ?

pankalog:
@MartinaeyNL I would like to preserve the branch, and this PR serves that, so nobody deletes the branch without realizing. Do we want to delete the branch? That's a SU/meeting decision that you can discuss with everyone.

PR #1149, opened 2023-10-31
Reverse Geocoding functionality for MaplibreGeocoder

pankalog:
Implementation is basically the exact same forwardGeocode function, with the function input prop and subsequently the request URL changed. Indicated as public for use from other classes, until this is moved to a utility class for use throughout or-map.


richturner:
Hi @pankalog this looks like a good piece of functionality to have, I'm going to pass this to @jvervier so he can merge this into a bigger piece of map work so we can keep the master branch history a little cleaner.

@jvervier can you please merge this into your work on the map and then we can remove this standalone PR

richturner:
Just merging this instead

PR #1151, opened 2023-11-01
Column header fix

DonWillems:
'column' was shown in the header instead of the column name

**Please review carefully**


MartinaeyNL (approved):

MartinaeyNL:
Don't know why I didn't comment on it, because I remember this error from a PR I reviewed earlier haha

PR #1153, opened 2023-11-01
Visual improvements with offline behavior for frontend apps

MartinaeyNL:
- Offline page now disabled by default.
- Increased reconnect timer to every 5 seconds.
- Added auto logout when refresh token expires (since it was present before the offline UI was implemented).
- The default offline page is now optional by boolean.


MartinaeyNL:
Put back to draft since some additional internal discussion is required.

MartinaeyNL:
Deleted this outdated PR. All functionality and visual changes got merged in #1172 .

PR #1172, opened 2023-11-21
Improvements offline consistency

MartinaeyNL:
Several improvements on the consistency of detecting the "offline state",
after doing research on the matter (internally) and discussions with other contributors.

Changelog to be written..


MartinaeyNL:
Currently in a testing phase on deployments, might need further improvements.
Not ready for review / merge yet.

MartinaeyNL:
The improvements are now stable.

Other than removing the debug comments, and doing visual tweaks on the offline page,
the code ready for review.

Can you take a look @richturner ?

ui/component/or-app/src/index.ts
```diff
@@ -316,13 +322,15 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
...
-                        this._mainElem.replaceChild(pageProvider.pageCreator(), elem); // recreate page
-                    } else {
+                    // If custom onRefresh() is set by the page, run that function.
+                    // Otherwise, just force recreate the page.
+                    const elem = this._mainElem.firstElementChild as Page<any>;
+                    if(elem?.onRefresh) {
                         elem?.style.removeProperty('display'); // show the current page again (back to the foreground)
+                        elem.onRefresh();
+                    } else if(pageProvider) {
+                        this._mainElem.replaceChild(pageProvider.pageCreator(), elem); // recreate page
```
richturner, line 333: Nice mechanism here.

ui/component/or-app/src/index.ts
```diff
@@ -419,9 +437,71 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
...
+        if(force) {
+            this._completeOfflineFallbackTimer(true);
+        } else if(this._offlineFallbackDeferred || this._showOfflineFallback) {
+            return;
+        }
+
+        const deferred = new Util.Deferred<void>();
+        let finished = false;
+        deferred.promise.then(() => {
+            if(this._showOfflineFallback !== this._offline) {
```
richturner, line 481: Is this check needed?

ui/component/or-app/src/index.ts
```diff
@@ -419,9 +437,71 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
...
+        }
+    }
+
+    // Offline timer logic
+    //
+    // This will start a Deferred promise that keeps track of the 'wait before showing offline page' timer.
+    // - Resolving the promise updates the 'show offline fallback' variable based on OFFLINE state.
+    // - Rejecting the promise 'aborts the process' and skips that logic and does nothing.
+    //
+    // To explain; when the Manager reports "We're offline!" it will wait 10+ seconds before visually reporting the user that he/she is offline.
```
richturner, line 469: Why do we need to 10s delay in showing offline state in the UI?

ui/component/core/src/index.ts
```diff
@@ -863,6 +966,16 @@ export class Manager implements EventProviderFactory {
         return this._basicIdentity ? this._basicIdentity.token : undefined;
     }
 
+    // Checking (manually) if token is invalid by verifying expiry date.
+    // By default, using a margin / minimum validity of 2 seconds
+    public isTokenExpired(margin = 2000): boolean {
```
richturner, line 971: Just a point to note that server and client times may be different but not much you can do about that

ui/component/core/src/index.ts
```diff
@@ -981,11 +1096,22 @@ export class Manager implements EventProviderFactory {
     }
 
     protected async updateKeycloakAccessToken(): Promise<boolean | void> {
+
```
richturner, line 1099: We should set token refresh based on keycloak settings, maybe create an issue for this

ui/component/or-app/src/index.ts
```diff
@@ -163,14 +168,30 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
...
+            this._onEvent(OREvent.CONSOLE_VISIBLE);
+        } else {
+            this._onEvent(OREvent.CONSOLE_HIDDEN);
+        }
+    }
+
     connectedCallback() {
         super.connectedCallback();
         this._storeUnsubscribe = this._store.subscribe(() => this.stateChanged(this.getState()));
+        this._onVisibilityBind = this.onVisibilityChange.bind(this);
```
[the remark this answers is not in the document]
MartinaeyNL: Arrow function didn't work when testing.
I have to store the listener function to be able to remove it during the `disconnectedCallback`.
So I decided to simply bind `this` to the function, so I'm able to dispatch events.

ui/component/or-app/src/index.ts
```diff
@@ -419,9 +437,71 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
...
+        } else if(event === OREvent.RECONNECT_FAILED) {
+            this._startOfflineFallbackTimer(); // start fallback timer (if not done yet)
+
+        } else if(event === OREvent.CONSOLE_VISIBLE) {
+            this._store.dispatch((setVisibility(true)));
+
+            // When the manager appears on Mobile devices, but the connection is OFFLINE,
+            // we reset the timer to the {appConfig.offlineTimeout} seconds. This is because we saw issues with reopening the app,
+            // and seeing a connection interval of 30+ seconds. We now give the user the benefit of the doubt, by resetting the timer.
+            if(manager.console?.isMobile && this._offline) {
```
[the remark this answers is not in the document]
MartinaeyNL: I would assume that the javascript threads are active again once it dispatches events.
Until now I haven't seen any issues with it. However, I get your point here.

Changed it a bit now where I try reconnecting on every foreground change,
no matter what state you are in, on both mobile and desktop.
Will do some testing now 👍

ui/component/or-app/src/index.ts
MartinaeyNL: For UI / UX purposes we didn't want to lock the UI when a user is only offline for 0.5 seconds.
So I implemented a minimum of 10 seconds, before it is shown to the user.
In practice this works fine for us.

ui/component/or-app/src/index.ts
MartinaeyNL: Good spot! Previously had a different mechanism for this, but now Lit will simply handle it.
Not needed anymore indeed.

MartinaeyNL:
Applied the requested changes. Did some testing with several devices and several network states, and it works great!

richturner (approved):

PR #1174, opened 2023-11-22
Two axis support on charts

### LINKED ISSUE #1123: Insights: Set the axes of a chart widget
As a insights page designer I want to set to which y-axis an asset attribute belongs so that the selected asset attributes are more clearly visualized.

In the asset attribute list there is an icon at the end of every line to select if this attribute is shown on y-axis 1 or 2. Click to toggle.

If the unit of the attributes set for a certain axis are the same, it can be shown on the axis

In the legend "(right)" will be shown to indicate which asset attributes belong to the right axis.

The asset attributes should be grouped by side (so all belonging to left axis first, then right)

image

MartinaeyNL:
Adds support for use of two axes on `or-chart`, and with configurable options in the chart widget.

Full changelog:
- Support for multi axis on the ChartJS config of `or-chart`.
- `or-chart` now has an `rightAxisAttributes` field, where an AttributeRef array lists what entries should appear on the right. Visually it shows a '(right)' text in the chart legend.
- Added optional actions callback to the attribute panel on Insights, to supply other actions than 'delete'.
Now used by the chart widget, where users can toggle if an entry should appear on the left or right.
- Added optional label callback to the attribute panel on Insights, to customize the HTML shown.
Now used by the chart widget, where '(right)' is shown when an entry is aligned to the right axis.
- Right axis min/max options on the chart widget settings.

Unrelated to the two axes support;
I also updated 'refresh options' dropdown in the UI, where 'OFF' now is a pause button.

Closes #1123


MartinaeyNL:
Ready to be merged after UI approval by @DonWillems.
@richturner we would like to have this merged before the end of the week.

manager/src/web/shared/locales/ro/or.json
richturner, line 1: Not all values have been added here

ui/component/or-chart/src/index.ts
```diff
@@ -370,6 +371,9 @@ export class OrChart extends translate(i18next)(LitElement) {
     @property({type: Object})
     public assetAttributes: [number, Attribute<any>][] = [];
 
+    @property({type: Array}) // The indexes of the assetAttributes array that appear on the right Y axis.
+    public readonly rightAxisAttributes: AttributeRef[] = [];
+
```
richturner, line 376: Maybe more appropriate to use a dedicated data type for an attribute on a chart which could contain all config options:

```
{
  "attributeRef": ...,
  "secondaryAxis": true,
 ...
}
```

Can see the benefit of having a separate property to avoid breaking existing usages.

Why is this property readonly?

richturner (approved):

manager/src/web/shared/locales/ro/or.json
MartinaeyNL: Did a minor change in the header from "Chart configuration" to "Axis configuration".
Simply applied some google translate magic so it doesn't break for romanian haha

ui/component/or-chart/src/index.ts
MartinaeyNL: Yeah indeed had a dilemma on how to approach this.
Might be benefitial for future proofing, so I will update this to a separate type as you described.

Made it readonly since I thought it was odd to change this value within `or-chart` itself.
However, now thinking about it, I think it doesn't do any harm at all. I will update this as well.

richturner (approved):

PR #1184, opened 2023-11-30
UI update: Fix input field for JSON Object value types (Fixes #1175)

### LINKED ISSUE #1175: JSON Object
In the latest docker image of the manager (3a31b8f4cca4) I cannot see the content of a JSON object added to an asset to retrieve the information of an HTTPLink agent, it does not show it in the field defined for this purpose as the previous image did (ff1c7342ee00); I don't know if this is the case in the new version, because it actually recovers the data so it can be used with an AttributeLink but it doesn't show anything in the attribute used to retrieve the request information.

richturner:


PepijnMuskens (approved):
Looked hard and long at this one but cant find any issues ;)

PR #1201, opened 2023-12-13
Fix OpenAPI configuration to generate error-free OpenAPI spec

pankalog:
This PR is for fixing the OpenAPI configuration, because it currently generates errors when parsed. To fix those errors, I introduce a new mixin and fix another one, fix the OAuth flow for service users, add mock scopes for the authorization, add a version, add a description to the writeAttributeValue request's response. 

To be discussed:
- How is the versioning going to work? How can I get the current version of OpenRemote there?
- Do we have scopes that can be defined there? They're not required for authentication anyway.


pankalog:
Closed by #1320

PR #1204, opened 2023-12-13
Add typed call for getOldValue consisent to getValue in AssetState class

peterdewitte:
In the AssetState class the next call is there
`public <U> Optional<U> getValue(Class<U> valueType)`

The equivalent call for the oldValue is not available. I added it
`public <U> Optional<U> getOldValue(Class<U> valueType)`

It increases the consistency in writing groovy scripts.


richturner:
This is obsolete due to a refactor of this file in https://github.com/openremote/openremote/pull/1205

PR #1208, opened 2023-12-19
Added endpoint to upsert predicted datapoints (Closes #1129)

### LINKED ISSUE #1129: API: post and get predicted datapoints
As a data analyst I want to retrieve and send predicted datapoints so I can feed my ML tool and write the generated forecast/future data to control the system.

Details of what kind of data sets will be used will be collected to add info to this issue

- Review usability of existing endpoints.
- Add missing endpoints
- Document use of endpoints

richturner:


model/src/main/java/org/openremote/model/datapoint/AssetDatapointResource.java
```diff
@@ -42,7 +40,7 @@ public interface AssetDatapointResource {
      * not have datapoint storage enabled.
      */
     @POST
-    @Path("{assetId}/attribute/{attributeName}")
+    @Path("{assetId}/{attributeName}")
```
MartinaeyNL, line 43: For consistency in endpoint paths I would either;
- rename the `POST` and `PUT` endpoints in `AssetPredictedDatapointResource` to this one.
- or revert this to `{assetId}/attribute/{attributeName}`

richturner: Good point I've updated `AssetPredictedDatapointResource` to match

MartinaeyNL (approved):

PR #1227, opened 2024-01-21
Enhancement/rememberme

bitmorse:
Enabling setting of remember me flag in realms. This allows users to use the checkbox upon login.


richturner (approved):
Thanks for the contribution

PR #1244, opened 2024-02-09
Feature/alarms

NinaBianca:
To do before merge:
- Code cleanup 

- [X] Update testcases


model/src/main/java/org/openremote/model/alarm/AlarmResource.java
MartinaeyNL, line 1: Before merging, I think it's worth updating the endpoint paths to a more standard REST API specification. For example:
- `GET` requests don't need an additional `/get` to the path.
- `POST`/`PUT` requests should have the IDs and other content specified in the body, and not in the path.
- Capital letters should be removed from the paths; use lower case.

model/src/main/java/org/openremote/model/rules/Alarms.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2017, OpenRemote Inc.
```
MartinaeyNL, line 2: You can by the way update these copyright notes to 2024.

test/src/test/groovy/org/openremote/test/alarm/AlarmTest.groovy
MartinaeyNL, line 1: Just to comment in general; We try to practice using end-to-end testing instead of writing "unit tests" for single use cases. This is because it simply takes a lot of time to write all these tests. It's fine as it is now, but worth keeping in mind for later haha

MartinaeyNL, line 1: The tests here are built around the REST API, and whether it functionally works. However, I'm missing tests about triggering alarms, if rules are correctly executed, etc.

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+import {getAlarmsRoute} from "../routes";
+import {when} from "lit/directives/when.js";
+import {until} from "lit/directives/until.js";
+import {InputType, OrInputChangedEvent, OrMwcInput} from "@openremote/or-mwc-components/or-mwc-input";
+import {OrMwcDialog, showDialog, showOkCancelDialog} from "@openremote/or-mwc-components/or-mwc-dialog";
+import {OrAssetTreeRequestSelectionEvent, OrAssetTreeSelectionEvent} from "@openremote/or-asset-tree";
+import {OrMwcTable, OrMwcTableRowClickEvent, TableColumn, TableRow} from "@openremote/or-mwc-components/or-mwc-table";
+import {MDCDataTable} from "@material/data-table";
+
+const tableStyle = require("@material/data-table/dist/mdc.data-table.css");
```
MartinaeyNL, line 29: Are these imports necessary for `page-alarms`? I would expect it in `or-mwc-table` but not here. If it's regards in styling, try adding `import "@openremote/or-mwc-components/or-mwc-table";` alongside of the current import.

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+    previousAssetLinks?: AlarmAssetLink[];
+    alarmUserLinks?: AlarmUserLink[];
+}
+
+@customElement("page-alarms")
+export class PageAlarms extends Page<AppStateKeyed> {
+    static get styles() {
+        // language=CSS
+        return [
+            unsafeCSS(tableStyle),
```
MartinaeyNL, line 63: Should not be necessary here.

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+
+    get name(): string {
+        return "alarm.alarm_plural";
+    }
+
+    constructor(store: Store<AppStateKeyed>) {
+        super(store);
+        window.addEventListener('selectChanged',  (event : CustomEvent)=>  {
+            this._selected = event.detail;
+        });
```
MartinaeyNL, line 289: You shouldn't listen and dispatch event for logic that is happening within the same HTMLElement. You can simply call a function in this case.

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+    public disconnectedCallback() {
+        super.disconnectedCallback();
+        this.reset();
+        window.removeEventListener('selectChanged', (event : CustomEvent)=>  {
+            this._selected = event.detail;
+        });
+    }
+
+    protected updated(changedProperties: Map<string, any>) {
+        setTimeout(() => {
```
MartinaeyNL, line 306: `setTimeout()` should not be necessary here, and wouldn't behalve much differently since the `updated()` lifecycle method is executed after rendering the UI.

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+            if(!this.alarm && !this.creationState && this.shadowRoot) {
+                const elem = this.shadowRoot.querySelector('or-mwc-table') as OrMwcTable;
```
MartinaeyNL, line 308: You can use `_tableElem` that you defined earlier here.

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+                                        if(columnIndex == 2 && span.parentElement?.style){
+                                            span.innerHTML = '<span style="border: 1px grey solid;' +
+                                                'padding: 4px;' +
+                                                'border-radius: 5px;' +
+                                                'color: grey;">Closed</span>';
+                                            span.parentElement.style.width = '125px';
+                                        }
+                                        break;
+                                }
+                            });
```
MartinaeyNL, line 380: This code could be way more efficient haha. Using a Lit directive might help with applying styling dynamically; see [classMap](https://lit.dev/docs/templates/directives/#classmap) and [styleMap](https://lit.dev/docs/templates/directives/#stylemap) docs. Either way I would move this logic to a separate function for easier reusability.

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+            return false;
+        }
+
+        return true;
+    }
+
+    protected async _createUpdateAlarm(alarm: AlarmModel, action: "update" | "create") {
+        console.log(alarm);
+        if (!alarm.title! || !alarm.content!) {
+            console.log("First if");
```
MartinaeyNL, line 435: Should remove these temporary logging statements before merging

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+        const assignOnly = !readAlarms && !writeAlarms;
+        this.config = {assignOnly: assignOnly};
+        return html`
+            <div id="wrapper">
+                <!-- Alarm Specific page -->
+
+                <div id="title" class="${this.creationState || this.alarm ? "hidden" : ""}" style="justify-content: space-between; margin-top: 10px; margin-bottom: 10px">
+                    <div class="${this.creationState || this.alarm ? "hidden" : ""}">
+                        <or-icon icon="bell-outline" style="margin-left: 20px"></or-icon>
+                        <span> ${this.alarm != undefined ? this.alarm.title : i18next.t("alarm.alarm_plural")} </span>
```
MartinaeyNL, line 510: You can use the `<or-translate>` component instead. We try to practice it nowadays as it's easier to maintain. Scan through this file and apply where necessary.

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+                                        style="margin: 0;"
+                                        type="${InputType.BUTTON}"
+                                        icon="delete"
+                                        @or-mwc-input-changed="${() => (this._deleteAlarms())}"
+                                ></or-mwc-input>
+                            </div>
+                            <hr class="${this._selected?.length > 0 ? "" : "hidden"}">
+                            <div class="${this.creationState || this.alarm ? "hidden" : "controls-left"}">
+                                <or-mwc-input .type="${InputType.CHECKBOX}" id="assign-check"
+                                              ?disabled="${disabled}" .label="${i18next.t("alarm.assignedToMe")}"
```
MartinaeyNL, line 528: With recent updates you don't have to do `i18next.t()` for the labels anymore, they're automatically translated now by or-mwc-input. (please double check)

ui/app/manager/src/pages/page-alarms.ts
```diff
@@ -0,0 +1,1133 @@
...
+
+                            </div>
+                            <div class="${this.creationState || this.alarm || assignOnly || readonly ? "hidden" : "panel-title"}"
+                                 style="justify-content: flex-end;">
+                                <or-mwc-input
+                                        raised
+                                        style="padding: 0px 10px; margin: 0;"
+                                        type="${InputType.BUTTON}"
+                                        icon="plus"
+                                        label="${i18next.t("add")} ${i18next.t("alarm.")}"
```
MartinaeyNL, line 550: Would recommend to just add a new entry in the translation file with `alarm.addAlarm`, as different languages use these combinations of words differently.

MartinaeyNL:
As a general note @NinaBianca, please update this branch with latest master branch.
There is something weird going on, making it tough to review, as I saw changes not related to your alarm branch haha

manager/src/main/java/org/openremote/manager/alarm/AlarmResourceImpl.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Override
+    public List<SentAlarm> getOpenAlarms(RequestParams requestParams) {
+        try{
+            return alarmService.getOpenAlarms();
+        } catch (IllegalArgumentException e) {
+            throw new WebApplicationException("Invalid criteria set", Status.BAD_REQUEST);
+        }
+    }
```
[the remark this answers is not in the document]
NinaBianca: Something similar to the user service?

NinaBianca:
one unrelated test seems to be failing

ui/app/manager/src/pages/page-alarms.ts
MartinaeyNL, line 1: As a minor (optional) note;
I recently started applying JSDoc / TSDoc for documentation within the code.
This automatically gets picked up by our docs tooling, and it's useful for when editing the code. 👍

ui/component/or-mwc-components/src/or-mwc-table.ts
```diff
@@ -1,14 +1,14 @@
...
 import {until} from 'lit/directives/until.js';
 import {MDCDataTable} from "@material/data-table";
-import {MDCCheckbox} from "@material/checkbox";
 import {when} from 'lit/directives/when.js';
-import {DefaultColor3, DefaultColor2, DefaultColor1, manager} from "@openremote/core";
+import {DefaultColor3, DefaultColor2, DefaultColor1} from "@openremote/core";
 import {i18next} from "@openremote/or-translate";
 import {InputType, OrInputChangedEvent} from "./or-mwc-input";
 import { styleMap } from "lit/directives/style-map.js";
+import moment from "moment";
```
MartinaeyNL, line 11: Not a fan importing `moment` here as it's a big library, and should not be a dependency of the table on itself.
Feel free to use a (temporarily) different strategy for adjusting / formatting dates in `or-mwc-table`.

ui/component/or-mwc-components/src/or-mwc-table.ts
```diff
@@ -180,15 +180,21 @@ export interface TableColumn {
     isSortable?: boolean
 }
 
+export type TableContent = string | number | Date | TemplateResult
```
MartinaeyNL, line 183: Nice touch of this `type`. Worth applying to many functions here. 👍

ui/component/or-mwc-components/src/or-mwc-table.ts
```diff
@@ -595,4 +704,12 @@ export class OrMwcTable extends LitElement {
         return this._dataTable ? (this._dataTable.root as HTMLElement) : (this.shadowRoot!.querySelector('.mdc-data-table') as HTMLElement);
     }
 
+    protected getColumnWidth(index: number, columns?: string[] | TableColumn[], tableWidthPx?: number): string | undefined {
+        return undefined;
```
MartinaeyNL, line 708: Note for myself; maybe make this a function a bit more useful compared to `undefined`. (if possible)

MartinaeyNL:
Just in general; there is still quite some work left before this can be merged.
From API improvements, to wiriting tests for the notifications and other services, to code quality standards.

Will chat on this later 👍

manager/src/main/java/org/openremote/manager/alarm/AlarmResourceImpl.java
```diff
@@ -0,0 +1,190 @@
...
+        if (isSuperUser()) {
+            LOG.finest("ALLOWED: Request from super user");
+            return;
+        }
+
+        if (!isAuthenticated()) {
+            LOG.fine("DENIED: Anonymous request are forbidden");
+            throw new WebApplicationException(Status.FORBIDDEN);
+        }
+    }
```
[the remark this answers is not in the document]
wborn: Each endpoint in `AlarmResource` now has roles assigned using `@RolesAllowed`.

manager/src/main/java/org/openremote/manager/alarm/AlarmService.java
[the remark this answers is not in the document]
wborn: Maybe create a separate issue for this as it is an issue in all other code as well?

There is the JPA Criteria API that can be used to prevent this, see: https://www.baeldung.com/sql-injection

model/src/main/java/org/openremote/model/alarm/AlarmResource.java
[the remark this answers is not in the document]
wborn: The API has been made more consistent. Let me know what you think of it now!

manager/src/main/java/org/openremote/manager/alarm/AlarmService.java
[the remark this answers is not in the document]
wborn: It's been reworked into a single GET operation with several optional filter query parameters.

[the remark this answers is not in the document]
wborn: There's mostly trivial code in this file now so it might be a bit unnecessary.

ui/app/manager/src/pages/page-alarms.ts
wborn: Is there still something to do he?
If so please link to an example.

ui/component/or-app/src/or-header.ts
```diff
@@ -357,9 +367,41 @@ export class OrHeader extends LitElement {
         if (changedProperties.has("config")) {
             this.activeMenu = getCurrentMenuItemRef(this.config && this.config.mainMenu && this.config.mainMenu.length > 0 ? this.config.mainMenu[0].href : undefined);
         }
+        if (changedProperties.has("realm")) {
+            this._getAlarmButton().then();
```
MartinaeyNL, line 371: `.then()` is not necessary here. Same for everywhere else in this class.

ui/component/or-mwc-components/src/or-mwc-table.ts
```diff
@@ -306,24 +321,43 @@ export class OrMwcTable extends LitElement {
...
+                                            const indeterminate = !allSelected && this.selectedRows && this.selectedRows.length > 0;
+                                            return html`
+                                                <th class="mdc-data-table__header-cell mdc-data-table__header-cell--checkbox"
+                                                    role="columnheader" scope="col">
+                                                    <div class="">
+                                                        <or-mwc-input type="${InputType.CHECKBOX}" id="checkbox"
+                                                                      .indeterminate="${indeterminate}" .value="${allSelected}"
+                                                                      @or-mwc-input-changed="${(ev: OrInputChangedEvent) => this.onCheckChanged(ev, ev.detail.value, "all")}">
+                                                        </or-mwc-input>
+                                                    </div>
```
MartinaeyNL, line 338: I think the `<div>` wrapper is unnecessary here.

ui/component/or-mwc-components/src/or-mwc-table.ts
```diff
@@ -306,24 +321,43 @@ export class OrMwcTable extends LitElement {
...
-                                                                  class="${classMap({'mdi-checkbox-intermediate': this.indeterminate!})}"
-                                                                  .indeterminate="${this.indeterminate}"
-                                                                  @or-mwc-input-changed="${(ev: OrInputChangedEvent) => this.onCheckChanged(ev.detail.value, "all")}" .value="${this.allSelected}">
-                                                    </or-mwc-input>
-                                                </div>
-                                            </th>
-                                            `
+                                        if(index === 0 && this.config.multiSelect){
+                                            const allSelected = this.rows && this.rows.length > 0 && this.selectedRows?.length === this.rows?.length;
+                                            const indeterminate = !allSelected && this.selectedRows && this.selectedRows.length > 0;
```
MartinaeyNL, line 329: Good improvement compared to the `state` it was previously!

test/src/test/groovy/org/openremote/test/alarm/AlarmTest.groovy
```diff
@@ -0,0 +1,263 @@
...
+
+    // Delete alarms as admin
+    def "should delete alarms as admin"() {
+        when:
+        def delete = adminResource.getAlarms(null, MASTER_REALM, null, null, null)
+        adminResource.removeAlarms(null, (List<Long>) delete.collect { it.id })
+
+        then:
+        adminResource.getAlarms(null, MASTER_REALM, null, null, null).length == 0
+    }
```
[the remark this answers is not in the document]
wborn: One tests removing a single alarm while the other tests removing all alarms. I've made this more clear and improved the tests.

manager/src/main/java/org/openremote/manager/alarm/AlarmService.java
```diff
@@ -0,0 +1,359 @@
...
+
+    public void removeAlarms(List<Long> alarmIds, Set<String> realms) throws IllegalArgumentException {
+        persistenceService.doTransaction(entityManager ->
+                entityManager.createQuery("delete from SentAlarm sa where sa.id in :ids")
+                        .setParameter("ids", alarmIds)
+                        .executeUpdate()
+        );
+
+        realms.forEach(realm -> clientEventService.publishEvent(new AlarmEvent(realm, PersistenceEvent.Cause.DELETE)));
+    }
```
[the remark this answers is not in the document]
wborn: The checks are there now. If you are a master realm admin you should be able to delete alarms in every realm. The validatity check in the AlarmService now checks for this.

manager/src/main/java/org/openremote/manager/alarm/AlarmResourceImpl.java
```diff
@@ -0,0 +1,183 @@
...
+    public void removeAlarms(RequestParams requestParams, List<Long> alarmIds) {
+        Set<String> realms = validateExistingAlarmIds(alarmIds);
+        alarmService.removeAlarms(alarmIds, realms);
+    }
+
+    @Override
+    public SentAlarm getAlarm(RequestParams requestParams, Long alarmId) {
+        validateExistingAlarmId(alarmId);
+        return alarmService.getAlarm(alarmId);
+    }
```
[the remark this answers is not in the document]
wborn: Moved the validation into the `AlarmService` now so it can perform better and always validates input parameters.

manager/src/main/java/org/openremote/manager/alarm/AlarmResourceImpl.java
```diff
@@ -0,0 +1,183 @@
...
+        try {
+            if (getUserId() != null) {
+                alarm.setSource(MANUAL);
+                alarm.setSourceId(getUserId());
+            }
+            else if (getClientId() != null) {
+                alarm.setSource(CLIENT);
+                alarm.setSourceId(getClientId());
+            }
+            return alarmService.sendAlarm(alarm);
```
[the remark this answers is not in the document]
wborn: I don't think it will ever return `null`. `persistenceService.doReturningTransaction` can throw an exception which would result in a "500 Internal Server Error".

manager/src/main/java/org/openremote/manager/alarm/AlarmResourceImpl.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    private void validateAlarmId(Long alarmId) {
+        if (alarmId == null) {
+            throw new WebApplicationException("Missing alarm ID", Status.BAD_REQUEST);
+        }
+        if (alarmId < 0) {
+            throw new WebApplicationException("Alarm ID cannot be negative", Status.BAD_REQUEST);
+        }
+    }
```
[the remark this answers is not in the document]
wborn: I've moved this into the `AlarmService` and added some exception mapping. This makes it easier to validate constraints whenever it is possible/required. This also allows the code to be more decoupled as well as provide information like messages, stacktraces and root causes why exceptions occur.

manager/src/main/java/org/openremote/manager/alarm/AlarmService.java
```diff
@@ -0,0 +1,359 @@
...
+        clientEventService.publishEvent(new AlarmEvent(alarm.getRealm(), PersistenceEvent.Cause.UPDATE));
+
+        if (alarm.getSeverity() == Alarm.Severity.HIGH) {
+            Set<String> excludeUserIds = Stream.of(userId, oldAssigneeId).filter(Objects::nonNull).collect(Collectors.toSet());
+            sendAssigneeNotification(alarm, excludeUserIds);
+        }
+    }
+
+    public void linkAssets(List<String> assetIds, String realm, Long alarmId) {
+        linkAssets(assetIds.stream().map(assetId -> new AlarmAssetLink(realm, alarmId, assetId)).toList());
```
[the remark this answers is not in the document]
wborn: There's now a `linkAsset` method for this flexibility.
You can only have one varargs argument and it can only be the last param.

ui/app/manager/src/pages/page-alarms.ts
wborn: That does not seem to work.

manager/src/main/java/org/openremote/manager/security/ManagerIdentityProvider.java
```diff
@@ -185,7 +185,7 @@ static User[] getUsersFromDb(PersistenceService persistenceService, UserQuery qu
                     // exists otherwise it will just match any other row
                     sb.append("NOT ");
                 }
-                sb.append("EXISTS (SELECT urm.user_id from public.user_role_mapping urm join public.keycloak_role kr on urm.role_id = kr.id where urm.user_id = u.id and not kr.client_role and");
+                sb.append("EXISTS (SELECT urm.user_id from public.user_role_mapping urm join public.keycloak_role kr on urm.role_id = kr.id where urm.user_id = u.id and");
```
wborn, line 188: Apparently the `alarms:write` role is a client role and would previously be filtered out. Without this change it would not be possible to query users having this role which is used to determine who should be send an e-mail in case an alarm has severity HIGH and no assignee. Querying users based on roles only seems to be done by the `UserResourceTest`.

richturner: This was to query `realmRoles`, user query doesn't seem to support client role searching so this should be added separately if needed.

I don't know why you would need `alarms:write` to recieve a notification but will look at that separately.

manager/src/main/java/org/openremote/manager/alarm/AlarmService.java
```diff
@@ -0,0 +1,504 @@
...
+     * the {@link Constants#READ_ADMIN_ROLE} or{@link Constants#WRITE_ALARMS_ROLE} are notified.
+     *
+     * @param alarm the alarm to send e-mail and push notifications for
+     * @param excludeUserIds users matching these user IDs will are excluded from the notifications
+     */
+    protected void sendAssigneeNotification(SentAlarm alarm, Set<String> excludeUserIds) {
+        ManagerIdentityProvider identityProvider = identityService.getIdentityProvider();
+
+        List<User> users = new ArrayList<>();
+        if (alarm.getAssigneeId() == null) {
```
richturner, line 264: Looks like `clientRoles` will need to be added to `UserQuery` if this behaviour is required

richturner (changes requested):

PR #1272, opened 2024-03-25
Added extensive /dashboard/query endpoint

MartinaeyNL:
Revamped dashboard querying and added advanced filtering such as
'only get dashboards where the user has access to all assets present in their widgets'

Full changelog;
- Added `/dashboard/query` endpoint with DashboardQuery object to filter dashboards. Uses a similar approach to `AssetQuery` where a JSON object provides details on what conditions to apply on the database query. Existing endpoints like `GET /all` and `POST /update` now also use the same querying function in `DashboardStorageService`.
.
- Dashboards can now be filtered by ids, displayName and user ids using the new `/query` endpoint.
- API for dashboards now supports pagination using the new `/query` endpoint.
- Now in the Manager, by default, restricted users will only see dashboards in the list that include linked assets.
- Bugfix for UI showing "error occurred" when user without `WRITE_INSIGHTS` was in edit mode;
now automatically forcing readonly use (view mode only).
- Added additional dashboard tests.


MartinaeyNL:
Currently a draft PR, since I need to add Groovy integration tests for this.

manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
```diff
@@ -28,59 +54,47 @@ public DashboardResourceImpl(TimerService timerService, ManagerIdentityService i
...
-        } catch (IllegalStateException ex) {
-            ex.printStackTrace();
-            throw new WebApplicationException(ex, INTERNAL_SERVER_ERROR);
+        return dashboards[0];
+    }
+
+    @Override
+    public Dashboard[] query(RequestParams requestParams, DashboardQuery query) {
+        if(query == null) {
+            query = this.createDashboardQuery(getRequestRealmName());
```
richturner, line 92: `getRequestRealmName()` only looks at the request but this may not be the user's realm and should only be allowed for super users; normal users should just look at authenticated realm.

We should probably change the `getRequestRealmName()` so this mistake cannot be made as it is too easy to make currently.

richturner (changes requested):
Great work on this one, just see my comment about the requested realm vs authenticated realm, might be worth covering cross realm requests in your test as well.

manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
MartinaeyNL: I was aware of differences between  `getRequestRealmName()` and `getAuthenticatedRealmName()` ;)
This is using the request realm on purpose, as the `/query` endpoint is also used for querying public dashboard that don't require authorization.

Will double check whether the cross-realm access is covered in tests right now.

manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
```diff
@@ -178,7 +190,7 @@ protected DashboardQuery createDashboardQuery(String realm) {
         assetAccess.add(DashboardQuery.AssetAccess.REALM);
 
         // Adjust query object based on user roles/permissions
-        if (isAuthenticated()) {
+        if (isSuperUser() || (isAuthenticated() && realm.equals(getAuthenticatedRealmName()))) {
```
richturner, line 193: I think this logic should be reversed and return a `FORBIDDEN` error if authenticated non super user is requesting another realm then return the 403.

Not sure just accepting it and giving them public dashboards of that realm is good. If they're doing something bad then let them know.

test/src/test/groovy/org/openremote/test/dashboard/DashboardTest.groovy
```diff
@@ -204,6 +205,27 @@ class DashboardTest extends Specification implements ManagerContainerTrait {
...
+        then: "it returns the building realm dashboard"
+        assert dashboardsInOtherRealm.length == 1
+        assert dashboardsInOtherRealm[0].displayName == "Dashboard with assets"
+
+        when: "the private user in master realm tries requesting a dashboard in the building realm"
+        def sameDashboardOtherRealmQuery = new DashboardQuery().realm(new RealmPredicate(keycloakTestSetup.realmBuilding.name))
+        def sameDashboardsInOtherRealm = privateUser1DashboardResource.query(null, sameDashboardOtherRealmQuery)
+
+        then: "no dashboard should be returned"
+        assert sameDashboardsInOtherRealm.length == 0
```
richturner, line 225: Then: throws a 403 exception

richturner (changes requested):

manager/src/main/java/org/openremote/manager/dashboard/DashboardResourceImpl.java
MartinaeyNL: Good suggestion! I've updated it in the newest changes. 👍

MartinaeyNL:
Good suggestion on returning `403` when a user tries accessing dashboards 'cross-realm'.👍
I've updated it in the newest changes.

richturner (approved):

PR #1273, opened 2024-03-27
Updated map-libre

Miggets7:


MartinaeyNL:
Don't have enough context / reasoning of this map-libre update.
For example, I'm really curious on;
- whether all breaking changes since v1 have been resolved
- compatibility with our own deployments and community projects (breaking changes?)
- compatibility with OR custom projects and their map configuration
- legacy compatibility with mapbox (whether that is necessary at all)
- in general whether it is possible for us, for example in regards to licensing.

So I'd rather avoid reviewing before discussed internally / rich is back from holidays.

Miggets7:
@DonWillems Was it tested on a project?

DonWillems:
@Miggets7 It's tested with the OR manager UI demo (and generic app) and in the K project. For OR everything seems fine, in the K project app we have immediate crashes on iOS. Android runs fine.

DonWillems (approved):

Issue #1277, opened 2024-03-31
Implement proper Keycloak Healthcheck before connecting


pankalog:
https://github.com/openremote/openremote/blob/622375d4e5c5c4985ba7679c8c23e73e51a0402e/container/src/main/java/org/openremote/container/security/keycloak/KeycloakIdentityProvider.java#L323

Using the welcome page as a tester for if keycloak is running is not the best. Keycloak now has a healthcheck API: https://www.keycloak.org/server/health

Maybe use this to check if keycloak is healthy or not before performing any actions. 

The healthcheck is already enabled in OpenRemote so no Keycloak configuration changes.

PR #1278, opened 2024-03-31
Update KeycloakIdentityProvider to use official healthcheck endpoint

pankalog:
Solves #1277: 

Using the welcome page as a tester for if keycloak is running is not the best. Keycloak now has a healthcheck API: https://www.keycloak.org/server/health

Maybe use this to check if keycloak is healthy or not before performing any actions.

The healthcheck is already enabled in OpenRemote's Keycloak instance so no Keycloak configuration changes.

### DIFF
--- a/container/src/main/java/org/openremote/container/security/keycloak/KeycloakIdentityProvider.java
+++ b/container/src/main/java/org/openremote/container/security/keycloak/KeycloakIdentityProvider.java
@@ -324,7 +324,7 @@ public abstract class KeycloakIdentityProvider implements IdentityProvider {
         Response response = null;
 
         try {
-            response = resource.getWelcomePage();
+            response = resource.getInstanceHealth();
             if (response != null &&
                 (response.getStatusInfo().getFamily() == SUCCESSFUL
                     || response.getStatusInfo().getFamily() == REDIRECTION)) {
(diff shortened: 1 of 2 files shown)


MartinaeyNL:
Looks OK to me.

If we would merge this, can we add the use of the `KeycloakIdentityProvider` here to the `or-core` frontend component as well?
I used an hardcoded path here, but would be useful to update.

```JavaScript
    // Checks whether keycloak is reachable using a simple HTTP request.
    // Since the keycloak JS adapter doesn't give us details of the HTTP responses they get, we test it manually using this.
    // Resolves or throws exception.
    protected async isKeycloakReachable(): Promise<void> {
        const controller = new AbortController();
        const timeout = setTimeout(() => controller.abort(), 2000);
        try {
            await fetch(this._config.keycloakUrl! + "/health/ready", {method: 'HEAD', mode: 'no-cors', signal: controller.signal});
            return Promise.resolve();
        } catch (e) {
            return Promise.reject(e);
        } finally {
            clearTimeout(timeout);
        }
    }
```

richturner:
The health endpoints on keycloak can be enabled/disabled so cannot be reliably used unless we hardcode this into our keycloak images so the health endpoints are always available.

Issue #1274, opened 2024-03-28
Users: add fields to filter users by username or email


DonWillems:
As an admin I want to easily find a user, so that I don't have to waste time clicking through the paginated list.

- Add filter field left of the '+ add user' button in both panels
- Filter the users shown below based on the match in username or email address
- Match is case insensitive, and can be at the start/middle/end of the username or email
- Decide whether filtering on front or backend is best
- Filter on input change, add a debounce of 250ms

PR #1280, opened 2024-04-04
Frontend updates: Search functionality to users page

MartinaeyNL:
In regards to issue #1274, I added a search field to both "regular user" and "service user" tables,
that filters the shown items in the table. So, this filtering done locally on the client after fetching all users.

Only includes frontend changes.

Changelog;
- Added search field to tables on the users page, that locally filters the `_users` or `_serviceUsers` array based on
username, email, first name, last name, et cetera. Using `@input` on the search field for live updates of the tables.
- Minor cleanup / refactors
- Small styling changes


ui/component/or-mwc-components/src/or-mwc-input.ts
```diff
@@ -678,6 +678,10 @@ export class OrMwcInput extends LitElement {
     @property({type: Boolean})
     public required: boolean = false;
 
+    /* Triggers OrInputChangedEvent based on @input instead of @change. Only supported for text-field input. */
+    @property({type: Boolean})
+    public live: boolean = false;
+
```
richturner, line 684: Can't any consumer just attach an input listener from outside of the `or-mwc-input` component to do something when input occurs?

richturner (commented):
See my comments

richturner (commented):
See my code comments

ui/component/or-mwc-components/src/or-mwc-input.ts
MartinaeyNL: You are indeed correct. I thought that event bubbling wasn't working for `or-mwc-input`, but it apparently is.
I removed it from `or-mwc-input` and added a `@input` listener on the consumer component instead.

MartinaeyNL:
Put it back to a draft as it is currently WIP.

ui/component/or-mwc-components/src/or-mwc-input.ts
MartinaeyNL: Event bubbling was not working for `@input` listener unfortunately.

Since `@input` only dispatches the keypress, for example `H`, we have no idea what its current value is.
Normally we would use `event.currentTarget` for this, but with the shadow DOM it won't return the correct element.
Directly referencing `OrMwcInput.value` also won't work, as this value isn't updated before a user submit.

I think adding a `live` property, where `@input` behaves exactly the same as `@change` does now,
is the most useful solution here. (instead of simply redispatching the input event)

MartinaeyNL:
Went through the requested changes, and tried out to use event bubbling instead of re-dispatching an event.
However, event bubbling was not working for `@input` listener unfortunately.
Comments from me;

> Since @input only dispatches the keypress, for example H, we have no idea what its current value is.
> Normally we would use event.currentTarget for this, but with the shadow DOM it won't return the correct element.
> Directly referencing OrMwcInput.value also won't work, as this value isn't updated before a user submit.
> 
> I think adding a live property, where @input behaves exactly the same as @change does now,
> is the most useful solution here. (instead of simply redispatching the input event)

Feel free to PM me if an additional chat is necessary 😄

richturner:
I still don't understand the need for this `live` option; for example in other parts of the code base the `@input` event handler is attached to the `or-mwc-input` control and the current value can be accessed:

```
@input=${(ev: Event) => inputChanged((ev.target as OrMwcInput).nativeValue)}
```

Happy to have a call to clarify why this alternative approach is required

MartinaeyNL:
Thanks for referring to `OrMwcInput.nativeValue`, as that was the fix I was looking for.
Previously `.value` was not working, as it didn't have the actual input value, only the submitted one.
This is now resolved.

Removed the `live` property from `or-mwc-input` as there is no use for it anymore.

@DonWillems you can either deploy this first or simply approve this.

DonWillems (approved):

Issue #1211, opened 2023-12-20
Assets: Linked users panel: all roles in the system are shown, not the roles assigned to the user


DonWillems:
Reproduce:

1. link user with no roles to an asset
2. View asset and look at linked users panel
3. See all roles per user (e.g. read, write) while that user does not have those roles.

Expected:
3. No roles listed

PR #1284, opened 2024-04-05
Frontend bugfixes: linked users table on or-asset-viewer

MartinaeyNL:
Fix for #1263 where only 10 items were present in the table due to a pagination issue.
Also fixed the roles being incorrect in the linked users table. (#1211)

Full changelog;
- Fix for `or-mwc-table` where a pagination filter was applied without pagination being enabled.
- Fix for `or-mwc-table` where pagination controls were shown without pagination being enabled.
- Issue where the linked users table in `or-asset-viewer` showed all roles instead of the assigned roles.


MartinaeyNL:
Sorry for this late response on the PR;

I agree that an exposed `paginationFilter` property, where the consumer / parent component decides what rows to show,
would be a nice addition. It would enhance flexibility and control. However, I would keep the additional `enable` property,
as there are more configuration options, such as setting the amount of rows per page.

Instead of implementing this functionality now, I suggest to create a new issue for this.
I'd rather have this simple fix merged first, and work on that later 😉

richturner (approved):

PR #1289, opened 2024-04-12
Added applyConfigToAdmin option to manager_config.json

### LINKED ISSUE #1287: Manager UI: Add manager_config item that lets superuser ignore the config so that they can see everything
Currently if logged in as SuperUser all (except basic colours) manager UI configurations are ignored. This is confusing/annoying to a lot of users.

Add an item to the highest level op de manager config that makes this optional. Boolean "ApplyConfigForSuperUser"

If true (default), the manager_config is applied for the SuperUser too.

This does not have to be shown on the appearance page yet.

MartinaeyNL:
Fixes #1287 

Added `applyConfigToAdmin` option to `manager_config.json`.
It controls if custom page configurations are shown to the admin user.
Defaults to true.


richturner:
I am not a fan of config that affects the admin user as quickly the admin user cannot perform actions in the manager UI (the admin tool)

DonWillems:
As someone who edits the manager config using the appearance page, its a bit awkward to do:
`"manager": {
    "applyConfigToAdmin": false
}`

instead of directly on the same level as 'realms' and 'pages':
`"applyConfigToAdmin": false`

But if that code-wise makes the most sense, it's ok for me

PR #1296, opened 2024-04-29
OAuth2 client credentials support for Email notifications

MartinaeyNL:
Currently a very basic, untested implementation of OAuth2 client credentials support for email notifications.
Haven't been able to test this at all, as I've been having a tough time setting up a SMTP service / connection using oAuth2.

Will need additional work and testing, as I have no clue if this behaves.


manager/src/main/java/org/openremote/manager/notification/EmailNotificationHandler.java
```diff
@@ -85,32 +92,59 @@ public void init(Container container) throws Exception {
...
-            props.put("mail." + protocol + ".port", port);
+            MailClientBuilder mailClientBuilder = new MailClientBuilder(
+                    container.getExecutorService(), protocol, host, port
+            );
+
+            // Add authentication
+            if(useOAuth) {
+                String oAuthUrl = container.getConfig().getOrDefault(OR_EMAIL_OAUTH2_URL, null);
+                String oAuthScopes = container.getConfig().getOrDefault(OR_EMAIL_OAUTH2_SCOPES, "");
+                if(!TextUtil.isNullOrEmpty(oAuthUrl)) {
```
richturner, line 108: Should also check clientId and secret are provided

model/src/main/java/org/openremote/model/Constants.java
```diff
@@ -59,6 +59,10 @@ public interface Constants {
     String OR_SSL_PORT = "OR_SSL_PORT";
     String OR_EMAIL_USER = "OR_EMAIL_USER";
     String OR_EMAIL_HOST = "OR_EMAIL_HOST";
+    String OR_EMAIL_OAUTH2_URL = "OR_EMAIL_OAUTH2_URL";
+    String OR_EMAIL_OAUTH2_CLIENT_ID = "OR_EMAIL_OAUTH2_CLIENT_ID";
+    String OR_EMAIL_OAUTH2_CLIENT_SECRET = "OR_EMAIL_OAUTH2_CLIENT_SECRET";
+    String OR_EMAIL_OAUTH2_SCOPES = "OR_EMAIL_OAUTH2_SCOPES";
```
richturner, line 65: Add these to `deploy.yml` with some info

richturner (changes requested):

richturner:
Discussed with @DonWillems to deploy this in k* dev environment to test there

manager/src/main/java/org/openremote/manager/notification/EmailNotificationHandler.java
MartinaeyNL: It's not an amazing solution, but the `useOAuth` variable is linked to whether both client_id and client secret are set.
I wanted to prevent a new `OAUTH_ENABLE` environment variable, so I simply check whether both exist,
as stated on line 80 of this file.

MartinaeyNL: Decided to add an additional check anyway. 👍

MartinaeyNL:
Made the changes as requested @richturner 👍

manager/src/main/java/org/openremote/manager/notification/EmailNotificationHandler.java
richturner: Sorry I missed that

richturner (approved):

PR #1299, opened 2024-05-14
Fix for value 0 not being shown in dropdown

MartinaeyNL:
Fixes an issue where the value `0` was not showing up as 'selected text' within the `or-mwc-input` dropdown.


MartinaeyNL:
Give me a second to double check a few things in the UI.

MartinaeyNL:
Yes this can be merged. I'm not expecting any UI to break, so skipping manual testing for now.

richturner (approved):

PR #1302, opened 2024-05-15
UI fixes: Attribute input not responding to language change

### LINKED ISSUE #1301: or-attribute-input: translate label when language is changed
If you change language some items are translated and other aren't. (some are listening to i18next LanguageChanged)

Make it so that the label in or-attribute-input also changes as soon as the user changes language.

MartinaeyNL:
Fixed an issue where the label in `or-attribute-input` did not update when a user changes the manager language.
Made a change in the `getLabel()` function, so it does not simply return if a label already exists anymore,
as a parent component might want to update the label. Don't think this change will have a performance impact.

Fixes #1301.

### DIFF
--- a/manager/src/web/shared/locales/nl/or.json
+++ b/manager/src/web/shared/locales/nl/or.json
@@ -460,7 +460,14 @@
       "ElectricityProducerWindAsset": "Windturbine asset"
     },
     "meta": {},
-    "attribute": {},
+    "attribute": {
+      "LightAsset": {
+        "brightness": "Helderheid",
+        "colourRGB": "Kleur RGB",
+        "colourTemperature": "Kleur temperatuur",
+        "groupNumber": "Groep nummer"
+      }
+    },
     "value": {
       "unknown": "Onbekend"
     }
--- a/manager/src/web/shared/locales/nl/or.json
+++ b/manager/src/web/shared/locales/nl/or.json
@@ -460,14 +460,7 @@
       "ElectricityProducerWindAsset": "Windturbine asset"
     },
     "meta": {},
-    "attribute": {
-      "LightAsset": {
-        "brightness": "Helderheid",
-        "colourRGB": "Kleur RGB",
-        "colourTemperature": "Kleur temperatuur",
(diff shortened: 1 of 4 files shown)


DonWillems:
The component does not respond to the different size of the label, resulting in possible unreadable labels.
image

richturner:
@MartinaeyNL as discussed on Skype, I cannot see a reason why this simple change should not work:
```
langChangedCallback = () => {
 this.requestUpdate(""label");
}
```

@DonWillems Looks like a rendering issue with material design component so we might need to call some update function on the MwcComponent itself

MartinaeyNL:
@richturner I just pushed another approach to fixing the issue.
I apparently misunderstood the cause, where the new label was correctly updated, but not applied,
as the cached template in `or-asset-viewer` was used, causing the HTML attributes not to update.

Now simply clearing the cache when the language changes.

This should also fix the issue @DonWillems mentioned, as the template is freshly rendered now.
Haven't seen the issue in the manager anymore after the fix.

MartinaeyNL:
Put back to draft, to double check whether all uses of `or-attribute-input` are fixed.
For example on the Rules page, and on several projects of ours.

MartinaeyNL:
Fixed several issues that occurred in regards to `or-attribute-input`.
The issues we were seeing in projects of ours like K**** were luckily reproducable within the OR manager.

All of these issues have all been tested locally on the Asset page, Rules page, Insights page, and in the K**** project.
Here is a changelog;

- The Material styling issue @DonWillems mentioned is fixed;
now calling the material UI `layout()` function upon `label` change.
- Use of `or-attribute-input` on other places than `or-asset-viewer`, were also not corresponding to language changes.
It was solved by calling `_updateTemplate()`, and requesting a UI update after `_templateProvider` was updated.
- For the `or-asset-viewer`, clearing the cache of `TemplateResult`s is still necessary for it to update the label properly.

.
Feel free to take a look @richturner, and to comment what you think.

DonWillems:
Seems to be all good. Could we have this merged this morning?

richturner (approved):

PR #1306, opened 2024-05-21
Set IdentityService to start before PersistenceService

richturner:


richturner:
This should solve the service startup issues seen on Balena, feel free to give it a try @pankalog

If ok then it can be approved and merged, would suggest removing the keycloak healthcheck PR then also due to reason mentioned there

richturner:
@pankalog Did you get a chance to try this branch?

pankalog:
> @pankalog Did you get a chance to try this branch?

No unfortunately, I wanted to but it's a quite big change and I'd like to do more extensive testing on the devices. Will try to do it this week!

pankalog (approved):
Looks good, testing showed good results! Maybe down the line we should adjust this service to only be for out-of-container setup tasks for both Keycloak and Postgres. 

Maybe to make this a bit more "official", an option is to add a "preinit" task for all services, which check if out-of-container services that are used are ready to be used before the actual init script.

PR #1310, opened 2024-06-02
Docker OpenRemote-wide logging to Manager Syslog port

pankalog:
This PR introduces the feature of OpenRemote-wide logging from all containers into the manager's `SyslogService`. This would allow for users to have an easier logging experience and allow for easier debugging/monitoring of OpenRemote as a whole. If I get this to work, I would optimally have that port serve as a Syslog server itself, allowing for sending all those logs outside of the container (to some other monitoring server). 

@richturner I'm having an issue within `UDPStringServer` specifically from `AbstractTCPServer`, starting somewhere in line 83. When debugging, I send a message using `echo "test" | nc -u -w1 localhost 1514 -v`, and I can see that I am hitting that line using a breakpoint, but the message does not seem to propagate to my messageConsumer. I'd appreciate any pointers, but I assume it's still my mistake, since it's dealing with low-level stuff. Thanks!

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,12 @@ volumes:
   manager-data:
   postgresql-data:
 
+x-logging: &default-logging
+  driver: syslog
+  options:
+    syslog-address: "tcp://manager:1514"
+    syslog-facility: "daemon"
+
 services:
 
   proxy:
--- a/manager/src/main/java/org/openremote/manager/syslog/SyslogService.java
+++ b/manager/src/main/java/org/openremote/manager/syslog/SyslogService.java
@@ -19,6 +19,16 @@
  */
 package org.openremote.manager.syslog;
 
+import io.netty.buffer.Unpooled;
+import io.netty.channel.socket.DatagramChannel;
+import io.netty.channel.socket.SocketChannel;
+import io.netty.handler.codec.DelimiterBasedFrameDecoder;
+import io.netty.handler.codec.string.StringDecoder;
+import io.netty.handler.codec.string.StringEncoder;
+import org.openremote.agent.protocol.io.IOServer;
+import org.openremote.agent.protocol.tcp.AbstractTCPServer;
+import org.openremote.agent.protocol.tcp.TCPStringServer;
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -12,6 +12,12 @@ volumes:
   manager-data:
   postgresql-data:
 
+x-logging: &default-logging
+  driver: syslog
+  options:
+    syslog-address: "tcp://manager:1514"
```
richturner, line 18: Does this driver use the host network rather than the docker network created for the stack? If so would need to use localhost and map port `127.0.01:1514` of the manager container

pankalog: The driver does use the host network, but there is a new mode called "non-blocking", which means that the docker stack will start regardless of the driver. This means that we could boot up the stack and then receive the syslogs. We could probably use `localhost` and only bind that port of the manager to `127.0.0.1`, so that it's not available outside the host. Again, eventually, it would be cool if OpenRemote could handle outside syslog requests (maybe let's say a remote device?) but that's out of the scope of this PR for now.

richturner:
I like the idea of being able to view all container logs within the manager but I have concerns about turning the manager into a syslog server as the manager is already doing too much.

The syslog service is also a considerable burden on the manager and the DB and I'd like to see a more standalone solution to logging and ability to read logs from the UI.

pankalog:
> I like the idea of being able to view all container logs within the manager but I have concerns about turning the manager into a syslog server as the manager is already doing too much.
> 
> The syslog service is also a considerable burden on the manager and the DB and I'd like to see a more standalone solution to logging and ability to read logs from the UI.

The main use-case for this would be to allow users (mainly the superuser) to view all container logs through the UI/maybe programmatically. This would allow users to debug issues without having to SSH into the instance and view the container logs. I'd personally love to have this with my fleet endeavours, since I have logging set up in a way that I can understand the payloads that come in from the devices. It would be an interim solution without adding new docker containers/services like elastisearch and Kibana to do this sort of thing, and a lot of the infrastructure for it is already there.

PR #1320, opened 2024-06-11
Fix OpenAPI configuration to generate error-free OpenAPI spec

pankalog:
Had to migrate to a new branch instead of using my fork. Close the other issue if you want to.

@richturner the password OAuth flow has been also added, let me know if anything else is needed!

### DIFF
--- a/manager/src/main/java/org/openremote/manager/web/ManagerWebService.java
+++ b/manager/src/main/java/org/openremote/manager/web/ManagerWebService.java
@@ -75,6 +75,11 @@ public class ManagerWebService extends WebService {
         List<String> _default;
     }
 
+    private static abstract class StringSchemaMixin {
+        @JsonProperty("enum")
+        protected List<String> _enum;
+    }
+
     public static final int PRIORITY = LOW_PRIORITY + 100;
     public static final String OR_APP_DOCROOT = "OR_APP_DOCROOT";
     public static final String OR_APP_DOCROOT_DEFAULT = "manager/src/web";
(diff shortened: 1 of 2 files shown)


pankalog:
@richturner it seems that the last error remaining for an error-free OpenAPI schema is about the request body from the DELETE request in `org.openremote.model.gateway.GatewayServiceResource#stopTunnel`. 

OpenAPI is very much a stickler for stuff like this, and according to the [HTTP reference](https://www.rfc-editor.org/rfc/rfc9110#name-delete), a DELETE request has "No Generally Defined Semantics". So OpenAPI is showing an error because a responseBody is included in the JSON schema: 
```
Semantic error at paths./gateway/tunnel.delete.requestBody
DELETE operations cannot have a requestBody.
```

Not really sure if you'd want to spend any amount of time on this, but just letting you know.

MartinaeyNL:
Hey @pankalog, saw this PR of yours coming by haha
Just as a small note; you can now also use https://docs.openremote.io for referring to the OR documentation. 😉

pankalog:
@MartinaeyNL fixed!

richturner (approved):
I replaced the password OAuth flow with authorizationCode as password is only available in dev mode generally as it is unsafe. Authcode is the standard flow used by users.

pankalog:
> I replaced the password OAuth flow with authorizationCode as password is only available in dev mode generally as it is unsafe. Authcode is the standard flow used by users.

Thanks!

PR #1327, opened 2024-06-20
Feature/mqtt gateway api

### LINKED ISSUE #1499: MQTT Gateway API branch update

dominiquekleeven:
**Draft pull request for the MQTT Gateway API**

Should be mostly feature complete, the GatewayV2Asset / GatewayV2Service might need a better name for clarity purposes?

**Couple notes:**
- Auto provisioning for gateway assets is not in this pull request, I've elaborated the issue to Rich (related to the service user creation), I am still trying to solve/come up with good solution here. I don't want to make too many changes in the provisioning handler since I am aware that scaling improvements are being made to it.  
- Ive added an event authorizer for AssetEvents, not sure if this is desirable can be changed so that the MQTT API does these checks by itself, just like the AssetResourceImpl class does for example.  
- It is worth noting that the parentId can be set in the asset template for both the update and create topic method, this has additional authorization checks outside of the canPublish function (just like multi-attribute update). If this is not desirable the parentId can be ignored and we can potentially provide a 'update-parent' topic that takes a parentId in addition to the assetId so we can handle it in the canPublish.

Regarding the topic structure/naming, the handler class is quite flexible when it comes to the topic keywords, they all use constants so these can be updated very quickly if there is a desire to potentially make the topics less verbose. It is worth noting that the research I've done, does show that MQTT topics can be rather verbose. AWS's reserved topics are a good example.


dominiquekleeven:
Branch is up to date again and functional, I am still going over a couple areas that might need some improvement and additional testing. Mainly handling the handling of subscriptions and pending events.

dominiquekleeven:
I spoke with Rich today about the MQTT Gateway API. 

A key takeaway was that the Gateway MQTT Handler has some generic topics, such as the CRUD operations, which should be added to the DefaultHandler. This way, the Gateway MQTT Handler can reuse the logic while adding its own specific logic, functionality, and authorization checks.

Essentially, by ensuring the logic is properly separated, between non gateway service users and regular ones.

Requires a bit of a refactor, will work on this as time permits.

Hackerberg43:
Just a suggestion: I spoke with Pierre last week about this topic and that the name gateway in this context is confusing, because the term gateway is also used in the manager interconnect feature. In IoT in the industrial sector, there are Edge gateways (like the manager interconnect achieves) and RTUs (remote terminal unit). RTUs read raw sensor data and translate it into a digital data stream to send it to a central unit, which seems like the exact goal of this feature :)

PR #1328, opened 2024-06-20
UI fixes; Gateway widget now shows active connections

MartinaeyNL:
Changelog:
- Gateway widget now shows active tunnels upon initialization of the UI;
letting the user know it is already active, and along with that, generally improving UX.

### DIFF
--- a/ui/component/or-dashboard-builder/src/widgets/gateway-widget.ts
+++ b/ui/component/or-dashboard-builder/src/widgets/gateway-widget.ts
@@ -127,9 +127,9 @@ export class GatewayWidget extends OrWidget {
                     ${when(this._loading, () => html`
                         <or-loading-indicator></or-loading-indicator>
                     `, () => {
-            if(this._activeTunnel) {
-                return html`
-                                
+                        if (this._activeTunnel) {
+                            return html`
+
                                 <or-mwc-input .type="${InputType.BUTTON}" icon="stop" label="${i18next.t('gatewayTunnels.stop')}" .disabled="${disabled}"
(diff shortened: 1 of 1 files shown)


ui/component/or-dashboard-builder/src/widgets/gateway-widget.ts
```diff
@@ -187,6 +204,17 @@ export class GatewayWidget extends OrWidget {
         }
     }
 
+    /**
+     * Internal function to set the active tunnel.
+     * Navigates the user to the tunnel after updating. This can be disabled using the {@link silent} parameter.
+     */
+    protected _setActiveTunnel(tunnelInfo?: GatewayTunnelInfo, silent = false) {
```
richturner, line 211: Don't see any usage where `silent=true` which should be the case for tcp tunnels

richturner (changes requested):

ui/component/or-dashboard-builder/src/widgets/gateway-widget.ts
MartinaeyNL: The protected `_navigateToTunnel()` function will automatically catch / ignore TCP tunnels.
(as there is no web URL to navigate to)

Currently I'm using the `silent` parameter during the `firstUpdated()` lifecycle function on line 112.
During first initialization of the page, the active tunnel is fetched and stored in cache using `_setActiveTunnel()`.
We only want to "open the tunnel in a new tab" when the tunnel is created using the "Start" button.
Not during page creation, so that's where the `silent` parameter comes in handy. 😉

richturner (approved):

PR #1329, opened 2024-06-21
UI Bugfix; landing page now based on header configuration

MartinaeyNL:
Previously, the default landing page for the Manager UI was always `/map`.
However, some realm configurations remove the `/map` from the header, causing a weird UX as "an invisible page" is used.
As discussed internally, in those cases we'd like to use a different header item as the default landing page.

Changelog:
- Pages not visible in the header will be put to a "lower priority", by moving them to the end of the `pages` array.
As the first item of that array acts as the landing page, this change makes sure an appropiate page will be shown.

### DIFF
--- a/ui/app/manager/src/index.ts
+++ b/ui/app/manager/src/index.ts
@@ -233,6 +233,18 @@ fetch(configURL).then(async (result) => {
             });
         }
 
+        // When the page is not present in the header, move the PageProvider to the end of the array.
+        // This is to prevent the landing page for the user not being visible in the header
+        const realmAppConfig = orAppConfig.realms[manager.displayRealm] || orAppConfig.realms.default;
+        if(realmAppConfig) {
+            const headerPaths = [...realmAppConfig.header.mainMenu, ...realmAppConfig.header.secondaryMenu].map(item => item.href);
+            pages.forEach(pageProvider => {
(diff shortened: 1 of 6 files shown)


ui/app/manager/src/index.ts
```diff
@@ -233,6 +233,18 @@ fetch(configURL).then(async (result) => {
...
+        // When the page is not present in the header, move the PageProvider to the end of the array.
+        // This is to prevent the landing page for the user not being visible in the header
+        const realmAppConfig = orAppConfig.realms[manager.displayRealm] || orAppConfig.realms.default;
+        if(realmAppConfig) {
+            const headerPaths = [...realmAppConfig.header.mainMenu, ...realmAppConfig.header.secondaryMenu].map(item => item.href);
+            pages.forEach(pageProvider => {
+                if(!headerPaths.includes(pageProvider.name)) {
+                    pages.push(pages.splice(pages.indexOf(pageProvider), 1)[0]);
+                }
+            })
```
richturner, line 245: My JS is getting rusty but modifying the collection whilst iterating...does that even work in JS? Maybe a quirk of how arrays are managed allows this.

MartinaeyNL: Good point. Adjusted it to use `.filter()` instead.

richturner (approved):

PR #1338, opened 2024-07-04
UI Bugfix; Insights navigation causing the browser tab to hang

MartinaeyNL:
Fix for browser crashes we were seeing on the Insights page, when navigating between dashboards.
It was very tricky to discover the cause of this, as it only occurred in production mode,
and the Insights overall is a very complex and intuitive piece of JS code.

I decreased the amount/occurrence of `widgetConfig` corrections, as that seems to be the cause.
Along with that, I've improved logging for the dashboard 'hasChanged' diff checks.

Changelog;
- While in view mode, the 'widget config correction' is now only applied within the preview,
and not exposed to a higher layer. As there is no need to 'save the changes that have been made' for example.
- Updated 'widget config correction' function, to only return a different JS object when changes have been made.
Originally it was returning the same object, but the fields were in a different order; causing unnecessary UI updates.
- Improved logging for `hasChanged` functionality, with the difference between dashboard objects being shared in the console.
- Added new util function using lodash; `difference(obj1, obj2)`
- Added `catch` clause for getting the initial tunnel connection in the `tunnel-widget`.

### DIFF
--- a/ui/component/core/src/util.ts
+++ b/ui/component/core/src/util.ts
@@ -29,7 +29,7 @@ import Qs from "qs";
 import {AssetModelUtil} from "@openremote/model";
 import moment from "moment";
 import DateTimeFormatOptions = Intl.DateTimeFormatOptions;
-import {isEqual, transform} from "lodash";
+import {transform} from "lodash";
 
 export class Deferred<T> {
 
@@ -284,7 +284,7 @@ export function objectsEqual(obj1?: any, obj2?: any, deep: boolean = true): bool
 export function difference(object?: any, base?: any): any {
     function changes(object?: any, base?: any) {
         return transform(object, function(result: any, value, key: string | number | symbol) {
-            if (!isEqual(value, base?.[key])) {
+            if (!objectsEqual(value, base?.[key])) {
                 result[key] = (isObject(value) && isObject(base?.[key])) ? changes(value, base?.[key]) : value;
             }
         });
--- a/ui/component/core/src/util.ts
+++ b/ui/component/core/src/util.ts
@@ -29,6 +29,7 @@ import Qs from "qs";
 import {AssetModelUtil} from "@openremote/model";
 import moment from "moment";
 import DateTimeFormatOptions = Intl.DateTimeFormatOptions;
+import {isEqual, transform} from "lodash";
 
 export class Deferred<T> {
 
@@ -273,6 +274,22 @@ export function objectsEqual(obj1?: any, obj2?: any, deep: boolean = true): bool
     }
 
     return false;
+}
+/**
+ * Deep diff between two object, using lodash
+ * @param  {Object} object Object compared
+ * @param  {Object} base   Object to compare with
+ * @return {Object}        Return a new object who represent the diff
+ */
+export function difference(object?: any, base?: any): any {
+    function changes(object?: any, base?: any) {
+        return transform(object, function(result: any, value, key: string | number | symbol) {
+            if (!isEqual(value, base?.[key])) {
+                result[key] = (isObject(value) && isObject(base?.[key])) ? changes(value, base?.[key]) : value;
+            }
+        });
+    }
+    return changes(object, base);
 }
 
 export function arrayRemove<T>(arr: T[], item: T) {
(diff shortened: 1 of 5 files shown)


MartinaeyNL:
Deployed it on a test server to confirm the issue is fixed 👍

MartinaeyNL:
Confirmed that Insights is working as expected, and nothing breaking is introduced.
Feel free to review.

ui/component/or-dashboard-builder/src/index.ts
```diff
@@ -329,9 +329,30 @@ export class OrDashboardBuilder extends LitElement {
...
-            const templateEqual = Util.objectsEqual(this.currentTemplate, this.initialTemplateJSON ? JSON.parse(this.initialTemplateJSON) : undefined);
-            this.hasChanged = (!dashboardEqual || !templateEqual);
+
+            // Check for dashboard changes
+            const initialDashboard = this.initialDashboardJSON ? JSON.parse(this.initialDashboardJSON) : undefined;
+            const dashboardEqual = Util.objectsEqual(this.selectedDashboard, initialDashboard);
+            if(!dashboardEqual) {
+                this.hasChanged = true;
+                const diff = Util.difference(this.selectedDashboard, initialDashboard);
+                console.debug("Dashboard has detected changes! Diff:", diff);
```
richturner, line 339: You're doing this diff just to log it to the console? Doesn't seem efficient considering the computation needed

ui/component/core/src/util.ts
```diff
@@ -274,6 +275,22 @@ export function objectsEqual(obj1?: any, obj2?: any, deep: boolean = true): bool
...
     return false;
 }
+/**
+ * Deep diff between two object, using lodash
+ * @param  {Object} object Object compared
+ * @param  {Object} base   Object to compare with
+ * @return {Object}        Return a new object who represent the diff
+ */
+export function difference(object?: any, base?: any): any {
+    function changes(object?: any, base?: any) {
```
richturner, line 285: Prefer arrow functions when inlining functions

ui/component/or-dashboard-builder/src/index.ts
```diff
@@ -329,9 +329,30 @@ export class OrDashboardBuilder extends LitElement {
...
+            }
+
+            // Check for template changes
+            else {
+                const initialTemplate = this.initialTemplateJSON ? JSON.parse(this.initialTemplateJSON) : undefined;
+                const templateEqual = Util.objectsEqual(this.currentTemplate, initialTemplate);
+                if(!templateEqual) {
+                    this.hasChanged = true;
+                    const diff = Util.difference(this.currentTemplate, initialTemplate);
+                    console.debug("Template has detected changes! Diff:", diff);
```
richturner, line 349: Same comment as above

ui/component/or-dashboard-builder/src/or-dashboard-widgetcontainer.ts
```diff
@@ -67,7 +73,7 @@ export class OrDashboardWidgetContainer extends LitElement {
             const oldVal = changedProps.get('widget') as DashboardWidget | undefined;
             const idChanged = oldVal?.id !== this.widget?.id;
             const nameChanged = oldVal?.displayName !== this.widget?.displayName;
-            const configChanged = JSON.stringify(oldVal?.widgetConfig) !== JSON.stringify(this.widget?.widgetConfig);
+            const configChanged = JSON.stringify(oldVal?.widgetConfig) !== JSON.stringify(this.getWidgetConfig());
```
richturner, line 76: stringify comparison is not reliable for objects as key order is not guaranteed, that's what `Util.objectsEqual()` method is for

richturner (changes requested):

ui/component/core/src/util.ts
MartinaeyNL: Sure! I adjusted it in the newest commit.

ui/component/or-dashboard-builder/src/or-dashboard-widgetcontainer.ts
MartinaeyNL: Agreed. I adjusted it in the newest commit.

MartinaeyNL:
Thanks for the review again! Made the adjustments you requested 😉

richturner (approved):

PR #1339, opened 2024-07-08
Tunnel is now stopped upon deleting the gateway asset

MartinaeyNL:
Fix for issue that the tunnel was not stopped when deleting a gateway asset.
I've added the `stopTunnel()` function within the `requestDisconnect` runnable, so it attempts to disconnect the tunnel.

Haven't done extensive testing (on a deployment for example), but it works fine for me locally.


manager/src/main/java/org/openremote/manager/gateway/GatewayService.java
```diff
@@ -684,7 +684,13 @@ protected void processGatewayConnected(String gatewayClientId, String sessionId)
...
         connector.connected(sessionId, createConnectorMessageConsumer(sessionId), () -> {
             clientEventService.closeSession(sessionId);
-            tunnelInfos.values().removeIf(tunnelInfo -> tunnelInfo.getGatewayId().equals(gatewayId));
+            tunnelInfos.values().removeIf(tunnelInfo -> {
+                if(tunnelInfo.getGatewayId().equals(gatewayId)) {
+                    this.stopTunnel(tunnelInfo);
+                    return true;
+                }
+                return false;
+            });
```
richturner, line 693: This is the wrong place to do this; we want to stop and remove tunnels in the `processGatewayChange` method specifically in the `DELETE` case.

richturner (changes requested):

MartinaeyNL:
Moved the logic to the `DELETE` clause in `processGatewayChange` as requested.
To be clear; I haven't tested this new logic on my machine. Please verify.

richturner (approved):

PR #1340, opened 2024-07-08
UI updates: Improve handling of keycloak offline token and reconnect logic

### LINKED ISSUE #1260: Keycloak: Offline Token: user has to log in after each redeploy
When the system is redeployed, the tokens in Keycloak memory are removed and the user needs to log in next time using the application.

An offline token should be stored on the users device, so that they stay logged in even if the application has been redeployed.

richturner:
Fixes #1260


.github/workflows/ci_cd.yml
richturner, line 1: This replaces continually failing github action

profile/dev-proxy.yml
richturner, line 1: This makes testing edge gateways easier and should have been added before

ui/component/core/src/index.ts
```diff
@@ -1105,76 +962,118 @@ export class Manager implements EventProviderFactory {
...
+                    this.login();
+                    return;
                 }
+                return true;
             }
+
+            // Check events
+            const eventsOffline = this.events && this.events.status === EventProviderStatus.CONNECTING;
+            // Force reconnect attempt now if needed
+            return !eventsOffline || await this.events?.connect();
```
MartinaeyNL, line 1037: So if I understand correctly;
If the websocket has disconnected, and this `reconnect()` function is triggered,
it will always request keycloak a new token, no matter if the token is still valid?

The other way around (token update fail, but websocket still alive) is catched with `eventsOffline`,
so it would think it's worth applying for the token update as well.

ui/component/core/src/index.ts
```diff
@@ -1105,76 +962,118 @@ export class Manager implements EventProviderFactory {
...
+                this._keycloak!.refreshToken = offlineToken;
+
+                try {
+                    await this._updateKeycloakAccessToken();
+                } catch (e) {
+                    this.login();
+                    return;
                 }
+                return true;
             }
```
MartinaeyNL, line 1032: Yes, I like this 👍 

May be worth seperating `tryReconnect` to class function instead, but that's up to you.

ui/component/core/src/index.ts
MartinaeyNL, line 1: In general I think some JSDoc comments would be nice to have here.

ui/component/core/src/index.ts
```diff
@@ -256,11 +263,10 @@ export class Manager implements EventProviderFactory {
         return this._config.mapType || MapType.VECTOR;
     }
 
+    protected static MAX_RECONNECT_DELAY = 45000;
```
MartinaeyNL, line 266: Just to double check; `WebsocketEventProvider` has a `MAX_RECONNECT_DELAY` field as well.
Maybe combining the default values, or reuse one of the variables?

ui/component/core/src/index.ts
```diff
@@ -1105,76 +962,118 @@ export class Manager implements EventProviderFactory {
...
+            return;
         }
-    }
 
-    /** Function that clears the `WebView` history of a console. It will not delete the history on regular browsers. */
-    protected _clearWebHistory(): void {
-        this.console?._doSendGenericMessage("CLEAR_WEB_HISTORY", undefined);
+        if (!connected) {
+            // Schedule reconnect again
+            reattemptDelayMillis = Math.min(Manager.MAX_RECONNECT_DELAY, reattemptDelayMillis + 3000)
```
MartinaeyNL, line 1049: Should we make those 3000ms a "global" value similar to `MAX_RECONNECT_DELAY` ?

ui/component/core/src/index.ts
```diff
@@ -848,8 +735,8 @@ export class Manager implements EventProviderFactory {
         }
 
         let result: BasicLoginResult = {
-            username: this._config.credentials?.username ? this._config.credentials?.username : "",
-            password: this._config.credentials?.password ? this._config.credentials?.password : "",
+            username: this.config.credentials?.username ? this.config.credentials?.username : "",
+            password: this.config.credentials?.password ? this.config.credentials?.password : "",
```
MartinaeyNL, line 739: Is there a reason to use `config` instead of `_config` here?
I don't exactly know the difference between them...

richturner: `this.config` is just using the getter which seems more appropriate but not that important

ui/component/core/src/index.ts
richturner: Could extract as a constant but not the highest priority unless it was used in multiple places

ui/component/core/src/index.ts
richturner: Happy to leave here as it is not used anywhere else and don't like creating spaghetti functions

ui/component/core/src/index.ts
```diff
@@ -784,10 +662,19 @@ export class Manager implements EventProviderFactory {
     }
 
     public logout(redirectUrl?: string) {
+        if (!this._authenticated) {
+            return;
+        }
+        this._authenticated = true;
```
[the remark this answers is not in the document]
richturner: Not sure how `keycloak.logout()` would behave if no user session but also no point executing this method multiple times

ui/component/core/src/index.ts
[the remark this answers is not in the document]
richturner: `this._connectEvents()` doesn't return the result and is not async

richturner: Not quite `updateKeycloakAccessToken()` will only get a new token if the existing token is going to expire in next 20s. 
Keycloak must be available and tokens valid for

ui/component/core/src/index.ts
richturner: There are quite a few comments around the changes I made. JSDoc isn't utilised at this time and lots of places in the code that could do with improved JSDoc comments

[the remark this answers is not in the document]
richturner: Done

MartinaeyNL (approved):

PR #1351, opened 2024-07-12
Full X.509 mTLS support for MQTT clients and MQTT Agent

pankalog:
This PR introduces support for mTLS for the MQTT Agent, and by extension, the MQTT client.

The `KeystoreService` is what deals with storage of Keystores, and has been made so that it's expandable when we eventually decide to implement mTLS on the ActiveMQ MQTT broker as well. 

To allow users to select their preferred keypair from the keystore, they now set their preferred alias as an agent attribute. This is passed to the `CustomKeyManagerFactorySpi`, which will select the certificate with the correct alias. 

Users will for now need to access the storage volume and navigate to the keystores directory to find 3 keystores per realm: a client truststore, a client keystore, and a server keystore. For now, only the first two are used. They are created/checked on start of `KeystoreService`. In bare-metal, this is in the tmp/keystores directory. 

Since there is no way for us to use the MQTT broker or any webserver that is running locally in the test suite, I had to use [test.mosquitto.org](test.mosquitto.org), which provides a mechanism to automate certificate signing and mTLS, which allows me to do a proper integration test.


pankalog:
@DonWillems after importing the certificates and keys to the keystore, I was able to connect to AWS IoT Core: 

`Client 'mqtts://***.amazonaws.com:8883/?clientId=***' connection status changed: CONNECTED`

pankalog:
Interesting way to trust all certificates:

richturner:
Having started looking through this some observations:

* PKCS12 is preferred keystore format since java 9 (thought I saw mention of `jks` in your code see: https://www.baeldung.com/java-keystore-truststore-difference)
* I'm thinking that a single keystore and truststore that is shared might be a better approach here, java is more setup for using a single file for each, we can then use realm prefix on alias
* Useful article on creating a custom truststore without losing the built in cacerts https://www.baeldung.com/java-custom-truststore
* `KeyStoreService` should be an abstraction away from keyStores as much as possible (i.e. allow MQTT client to retrieve the stores and allow key lookup by alias)

Will add some more comments later.

pankalog:
@richturner thanks for the review!

> * PKCS12 is preferred keystore format since java 9 (thought I saw mention of `jks` in your code see: https://www.baeldung.com/java-keystore-truststore-difference)

KeyStores are PKCS12 format, I only mention JKS in the shell file which will be removed

> * I'm thinking that a single keystore and truststore that is shared might be a better approach here, java is more setup for using a single file for each, we can then use realm prefix on alias

So you are suggesting that all realms share a single client KeyStore and a single client TrustStore? I'd be okay with that, but issue is that for now, users are required to add the keypairs/certificates by themselves to the keystores, and having the same one for all realms is quite unsecure. For now, SERVER_KEYSTORE can be removed though. 

Since everything client-related in OpenRemote (MQTT agent/client) is naturally segregated by realm, I don't see why we should do this. Under no circumstances would multiple realms have the need (or ability) to share keypairs or trusted certificates. If anything, with my current implementation, it feels that it'll be more of a pain, and be less secure, to use the same key/truststore in the entire manager, since as you said we'd have to pre/append the alias during insertion/retrieval of the keypair/certificate, etc. For when the time comes to implement mTLS on the broker, then we can discuss about that, maybe the more sensible thing would be to combine KeyManagers when creating the KeyManagerFactory.

> * `KeyStoreService` should be an abstraction away from keyStores as much as possible (i.e. allow MQTT client to retrieve the stores and allow key lookup by alias)

That was the plan for it, I'll abstract `KeyManagerFactory` and `TrustManagerFactory` into the KeyStoreService.

richturner:
> So you are suggesting that all realms share a single client KeyStore and a single client TrustStore? I'd be okay with that, but issue is that for now, users are required to add the keypairs/certificates by themselves to the keystores, and having the same one for all realms is quite unsecure. For now, SERVER_KEYSTORE can be removed though.
> 
> Since everything client-related in OpenRemote (MQTT agent/client) is naturally segregated by realm, I don't see why we should do this. Under no circumstances would multiple realms have the need (or ability) to share keypairs or trusted certificates. If anything, with my current implementation, it feels that it'll be more of a pain, and be less secure, to use the same key/truststore in the entire manager, since as you said we'd have to pre/append the alias during insertion/retrieval of the keypair/certificate, etc. For when the time comes to implement mTLS on the broker, then we can discuss about that, maybe the more sensible thing would be to combine KeyManagers when creating the KeyManagerFactory.

By the fact the user has to manager the key stores using the host file system and volume mappings already means they have full access to the system so the security argument doesn't seem relevant.

Yes indeed everything is segregated by realm for regular users and the implementation details of how this is done is not too relevant. For example we have one DB and all queries are locked to a specific realm, this is not too different to the key store concept with realm prefix on aliases. The KeyStoreService again should hide this complication but should take a realm parameter on any relevant methods we could then change the underlying implementation (one or multiple key stores) and consumers are not affected.

richturner:
Use technique in https://www.baeldung.com/java-custom-truststore to provide a custom X509TrustManager that combines a custom key store file (using path defined by `OR_SSL_TRUST_STORE` default should be a relative path and all values should be resolved using `PersistenceService.resolvePath()`)

Looking at the SPI for KeyManagerFactory then the same approach can be used there.

Anywhere in code that KeyManagerFactory is used `realm.` should be used as a prefix - This could be a static helper method in KeyStoreService.

TrustManagerFactory can just share cacerts between realms.

model/src/main/java/org/openremote/model/security/KeyStoreService.java
```diff
@@ -0,0 +1,85 @@
...
+
+		public Path getPath(){
+			return filePath;
+		}
+
+		@Override
+		public String toString() {
+			return String.format("%s (File Extension: %s, OpenRemote environment variable: %s)", this.name(), fileName, environmentVariableName);
+		}
+	}
```
richturner, line 73: This shouldn't be needed now

model/src/main/java/org/openremote/model/security/KeyStoreService.java
```diff
@@ -0,0 +1,85 @@
...
+public interface KeyStoreService extends ContainerService {
+	String OR_KEYSTORE_PASSWORD = "OR_KEYSTORE_PASSWORD";
+	Logger LOG = Logger.getLogger(KeyStoreService.class.getName());
+
+	KeyStore getKeyStore(String realm, KeyStoreType type);
+	void storeKeyStore(KeyStore keyStore, String realm, KeyStoreType type);
+	char[] getKeyStorePassword();
+
+	KeyManagerFactory getKeyManagerFactory(String realm, String preferredAlias) throws Exception;
+	TrustManagerFactory getTrustManagerFactory(String realm) throws Exception;
```
richturner, line 22: One trust manager should be sufficient, I don't know why you would want to duplicate all cacerts for each realm

model/src/main/java/org/openremote/model/security/KeyStoreService.java
```diff
@@ -0,0 +1,85 @@
...
+
+public interface KeyStoreService extends ContainerService {
+	String OR_KEYSTORE_PASSWORD = "OR_KEYSTORE_PASSWORD";
+	Logger LOG = Logger.getLogger(KeyStoreService.class.getName());
+
+	KeyStore getKeyStore(String realm, KeyStoreType type);
+	void storeKeyStore(KeyStore keyStore, String realm, KeyStoreType type);
+	char[] getKeyStorePassword();
+
+	KeyManagerFactory getKeyManagerFactory(String realm, String preferredAlias) throws Exception;
```
richturner, line 21: I thought we discussed not exposing the realm to the key factory but instead the key alias should be prefixed with the realm name (i.e. in MQTTProtocol where the client is created).

richturner (changes requested):

agent/src/main/resources/org/openremote/agent/protocol/mqtt/create_keystores.sh
```diff
@@ -0,0 +1,27 @@
...
+keytool -export -alias $ALIAS -file ${ALIAS}.crt -keystore $KEYSTORE_FILENAME -storepass $PASSWORD
+
+# Create a truststore and import the certificate
+keytool -import -alias $ALIAS -file ${ALIAS}.crt -keystore $TRUSTSTORE_FILENAME -storepass $PASSWORD -noprompt
+
+# Clean up the exported certificate file
+rm -f ${ALIAS}.crt
+
+echo "Keystore and Truststore have been created with the password: $PASSWORD"
+
```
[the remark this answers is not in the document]
pankalog: I wasn't going to include this in the final merge, but if you like it I can keep it. OpenRemote will generate the KeyStores when starting anyway, in the absence of the environment variables

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -144,6 +176,17 @@ public TrustManagerFactory getTrustManagerFactory() throws Exception {
...
+		KeyStore keystore = ((KeyStore)null);
+		// Save the newly created KeyStore
+		try (OutputStream os = new FileOutputStream(path.toFile())) {
+			keyStore.store(os, getKeyStorePassword());
+			keystore.load(new FileInputStream(path.toFile()), getKeyStorePassword());
+		} catch (Exception saveException) {
+			getLogger().severe("Couldn't store KeyStore to Storage! " + saveException.getMessage());
+		}
+		return keystore;
+	}
```
richturner: Don't think we should be creating key stores, user supplies them or doesn't. If this is just for tests then move it there. FYI: Groovy tests can access private/protected/package visible methods.

model/src/main/java/org/openremote/model/security/KeyStoreService.java
```diff
@@ -14,72 +14,14 @@ public interface KeyStoreService extends ContainerService {
...
 
-	KeyStore getKeyStore(KeyStoreType type);
-	void storeKeyStore(KeyStore keyStore, KeyStoreType type);
+	KeyStore getKeyStore();
+	void storeKeyStore(KeyStore keyStore);
+	KeyStore getTrustStore();
+	void storeTrustStore(KeyStore keyStore);
 	char[] getKeyStorePassword();
-
 	KeyManagerFactory getKeyManagerFactory(String preferredAlias) throws Exception;
```
richturner, line 22: Not sure the intention of `preferredAlias` rather than just `alias`, if it is expected that a realm prefixed aliase be supplied then name the parameter something like `realmPrefixedAlias`

model/src/main/java/org/openremote/model/security/KeyStoreService.java
```diff
@@ -14,72 +14,14 @@ public interface KeyStoreService extends ContainerService {
...
-	/**
-	 * This method is used by all ManagerFactories, that returns True if the provided string s
-	 * @param s The string to check
-	 * @param requestedAlias the alias that the user specified in the client creation. The alias prefixing happens before
-	 *                       this is called.
-	 * @return True, if all criteria are matched.
-	 */
 	static Boolean isRequestedAlias(String s, String requestedAlias){
 		return s.equals(requestedAlias);
 	}
```
richturner: Don't see why this is needed here

richturner (commented):
Added some comments but guess you're still working on this

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -55,66 +71,83 @@ public void init(Container container) throws Exception {
...
+		try {
+			return KeyStore.getInstance(this.keyStorePath.toFile(), getKeyStorePassword());
+		} catch (Exception e) {
+			throw new RuntimeException(e);
 		}
 	}
 	@Override
-	public KeyStore getKeyStore(KeyStoreType type) {
-		Path storePath = type.getPath();
+	public KeyStore getTrustStore() {
```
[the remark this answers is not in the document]
pankalog: That's because there's no actual merging of keystores happening, because the SPI I had to create just considers all keystores supplied to it, in `org.openremote.manager.security.KeyStoreServiceImpl#getTrustManagerFactory`.

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
pankalog: The issue with this is that, if we do so, then we will have to somehow fail fast and hard during configuration of an agent to use mTLS. I'm not exactly sure what the difference would be if we do create the keystore, as we can fail silently saying that the alias requested does not exist in the keystore. Most users will ignore the functionality, the files will be basically empty, and when they want to use them, they're available. A lot of users also may not know how to exactly create a keystore, and using a GUI for it makes it a lot easier, and letting us do the heavy lifting with creating the keystores may be something convenient for them. Everything is documented as well, and I'll update the documentation when this PR is done.

model/src/main/java/org/openremote/model/security/KeyStoreService.java
pankalog: I was thinking of using this as a way to allow expandability but at the end of the day I didn't need it. Removed.

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -35,6 +38,19 @@ public class KeyStoreServiceImpl implements KeyStoreService {
...
+	private final String OR_SSL_CLIENT_TRUSTSTORE_FILE = "OR_SSL_CLIENT_TRUSTSTORE_FILE";
+
+	private final String OR_SSL_CLIENT_KEYSTORE_PASSWORD = "OR_SSL_CLIENT_KEYSTORE_PASSWORD";
+	private final String OR_SSL_CLIENT_TRUSTSTORE_PASSWORD = "OR_SSL_CLIENT_TRUSTSTORE_PASSWORD";
+
+	private Path keyStorePath;
+	private Path trustStorePath;
+
+
 	public String OR_KEYSTORE_PASSWORD_DEFAULT;
```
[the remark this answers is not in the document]
pankalog: I'm not sure what you mean? The password env vars are used if the specified KeyStore in OR_SSL_CLIENT_KEY/TRUSTSTORE_PATH uses a password other than OR_KEYSTORE_PASSWORD

container/src/main/java/org/openremote/container/util/MapAccess.java
```diff
@@ -68,9 +70,10 @@ static public BigDecimal getDecimal(Map<String, String> map, String key, BigDeci
         return defaultValue;
     }
 
-	public static <T> Optional<T> getValue(Map<String, ?> map, String key) {
+	public static <T> Optional<T> getValue(Map<String, ?> map, String key, Class<T> type) {
 		if (map.containsKey(key)) {
-			return Optional.ofNullable((T) map.get(key));
+			// Assuming ValueUtil.getValueCoerced returns a value of type T
+			return Optional.ofNullable(type.cast(ValueUtil.getValueCoerced(map.get(key), type)));
```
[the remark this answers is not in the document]
pankalog: Removed

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -55,66 +71,83 @@ public void init(Container container) throws Exception {
...
-					keyStore.store(os, getKeyStorePassword());
-				} catch (Exception saveException) {
-					getLogger().severe("Couldn't store KeyStore to Storage! " + saveException.getMessage());
-				}
+		if(trustStorePath.isPresent()){
+			try{
+				keyStore = KeyStore.getInstance(trustStorePath.get().toFile(), trustStorePassword.toCharArray());
+			} catch (Exception e){
+				throw e;
+			}
```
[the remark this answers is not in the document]
pankalog: Yep, removed

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -55,66 +71,83 @@ public void init(Container container) throws Exception {
...
+			throw new RuntimeException(e);
 		}
 	}
 	@Override
-	public KeyStore getKeyStore(KeyStoreType type) {
-		Path storePath = type.getPath();
+	public KeyStore getTrustStore() {
 		try {
-			return KeyStore.getInstance(storePath.toFile(), getKeyStorePassword());
+			return KeyStore.getInstance(this.trustStorePath.toFile(), getKeyStorePassword());
```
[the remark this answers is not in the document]
pankalog: Fixed, now checks using `Optional.isPresent()`

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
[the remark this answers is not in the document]
richturner: I mean I don't see a need for a field for this as it can be inlined

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -0,0 +1,194 @@
...
+		CustomX509TrustManagerFactory tmf = new CustomX509TrustManagerFactory(this.trustStore, (KeyStore) null);
+		try {
+			tmf.init((KeyStore) null);
+		} catch (Exception e) {
+			throw new Exception("Could not retrieve KeyManagerFactory: "+e.getMessage());
+		}
+
+		return tmf;
+	}
+	private KeyStore createKeyStore(Path path) throws Exception {
```
richturner, line 170: We tend to use `protected` methods to allow custom implementations to be created more easily

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -0,0 +1,194 @@
...
+				this.keyStore = KeyStore.getInstance(new File(defaultKeyStorePath.toUri()), getKeyStorePassword());
+			}else{
+				this.keyStore = createKeyStore(defaultKeyStorePath);
+			}
+
+			this.keyStorePath = defaultKeyStorePath;
+		}
+
+		if(trustStorePath.isPresent()){
+			this.keyStore = KeyStore.getInstance(trustStorePath.get().toFile(), trustStorePassword.toCharArray());
```
richturner, line 99: Incorrect var assignment

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -0,0 +1,194 @@
...
+			Path defaultTrustStorePath = persistenceService.resolvePath(Paths.get("keystores").resolve("client_truststore.p12"));
+			if(new File(defaultTrustStorePath.toUri()).exists()) {
+				this.trustStorePath = defaultTrustStorePath;
+				this.trustStore = KeyStore.getInstance(new File(defaultTrustStorePath.toUri()), getKeyStorePassword());
+			}else{
+				this.trustStore = createKeyStore(defaultTrustStorePath);
+			}
+			this.trustStorePath = defaultTrustStorePath;
+		}
+	}
```
richturner, line 110: Any assignment of a custom truststore should probably be set as the default for the whole JVM

agent/src/main/java/org/openremote/agent/protocol/mqtt/MQTTProtocol.java
```diff
@@ -107,7 +118,15 @@ protected MQTT_IOClient doCreateIoClient() throws Exception {
             lastWill = new MQTTLastWill(topic, payload, retain);
         }
 
-        return new MQTT_IOClient(agent.getClientId().orElseGet(UniqueIdentifierGenerator::generateId), host, port, agent.isSecureMode().orElse(false), !agent.isResumeSession().orElse(false), agent.getUsernamePassword().orElse(null), websocketURI, lastWill);
+		//It's fine if they're null, they're not going to be used when creating the client
+	    TrustManagerFactory trustManagerFactory = null;
+		KeyManagerFactory keyManagerFactory = null;
+		if(agent.isSecureMode().orElse(false)){
+			trustManagerFactory = keyStoreService.getTrustManagerFactory();
+			keyManagerFactory = keyStoreService.getKeyManagerFactory(agent.getRealm()+"."+agent.getCertificateAlias().orElseThrow());
```
richturner, line 126: It could be possible to have a custom trust cert but not want client auth and maybe vice versa so throwing exceptions if not set implies it is required.

agent/src/main/java/org/openremote/agent/protocol/mqtt/AbstractMQTT_IOClient.java
```diff
@@ -123,8 +128,12 @@ protected AbstractMQTT_IOClient(String clientId, String host, int port, boolean
             .applyAutomaticReconnect();
 
         if (secure) {
-            builder = builder.sslWithDefaultConfig();
-        }
+			MqttClientSslConfigBuilder sslBuilder = MqttClientSslConfig.builder();
+			sslBuilder = sslBuilder.keyManagerFactory(keyManagerFactory);
```
richturner, line 132: As mentioned below; it may be that only a custom trust store is needed, don't know what this method does if null is supplied

richturner (changes requested):

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -0,0 +1,200 @@
...
+    private KeyStore keyStore = null;
+    private KeyStore trustStore = null;
+
+    private final String OR_SSL_CLIENT_KEYSTORE_FILE = "OR_SSL_CLIENT_KEYSTORE_FILE";
+    private final String OR_SSL_CLIENT_TRUSTSTORE_FILE = "OR_SSL_CLIENT_TRUSTSTORE_FILE";
+
+    private final String OR_SSL_CLIENT_KEYSTORE_PASSWORD = "OR_SSL_CLIENT_KEYSTORE_PASSWORD";
+    private final String OR_SSL_CLIENT_TRUSTSTORE_PASSWORD = "OR_SSL_CLIENT_TRUSTSTORE_PASSWORD";
+
+    private final String OR_KEYSTORE_PASSWORD = "OR_KEYSTORE_PASSWORD";
```
richturner, line 46: These should be static

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -0,0 +1,200 @@
...
+
+
+    private Path keyStorePath;
+    private Path trustStorePath;
+
```
richturner, line 51: un-necessary space

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -0,0 +1,200 @@
...
+    public String keyStorePassword;
```
richturner, line 52: Why is this public

manager/src/main/java/org/openremote/manager/security/KeyStoreServiceImpl.java
```diff
@@ -0,0 +1,200 @@
...
+    private final String OR_SSL_CLIENT_TRUSTSTORE_FILE = "OR_SSL_CLIENT_TRUSTSTORE_FILE";
+
+    private final String OR_SSL_CLIENT_KEYSTORE_PASSWORD = "OR_SSL_CLIENT_KEYSTORE_PASSWORD";
+    private final String OR_SSL_CLIENT_TRUSTSTORE_PASSWORD = "OR_SSL_CLIENT_TRUSTSTORE_PASSWORD";
+
+    private final String OR_KEYSTORE_PASSWORD = "OR_KEYSTORE_PASSWORD";
+
+
+    private Path keyStorePath;
+    private Path trustStorePath;
```
richturner, line 50: prefer protected as it allows extensions of the service to be created more easily

richturner (changes requested):
Please add standard copyright header to all new files

richturner (approved):

PR #1353, opened 2024-07-16
First implementation of edge gateway attribute event rate limiting

### LINKED ISSUE #1354: Add ability to configure the rate at which attribute events are sent from an edge gateway to a central instance
Currently all attribute events that occur on an edge gateway are forwarded to the central instance; it can be desirable to have high frequency attribute updates on the edge gateway but to independently control the frequency with which these attributes are forwarded to the central instance.

richturner:
Closes #1354


richturner:
Not had much time to go through but I don't see all the `const` values extracted from the filters being used.

Important to note that if `matcher` is `null` then it is a match all condition and should generally be the last item in the array (otherwise it would match before the others).

MartinaeyNL:
Finished a basic version of the new Interconnect UI.

I haven't tested the interconnect logic itself, so there might be mistakes in the `GatewayAttributeFilter` upon saving.
So I'm expecting some changes to be honest. The JSON editor however should work fine.

MartinaeyNL:
Thanks @richturner! Will pick this up;

- [ ] Add translations (forgot to commit them, my bad)
- [ ] Discussed with @DonWillems, and will add this using a 'cloud upload icon'.
*Created an issue for adding "select all" option;* https://github.com/openremote/openremote/issues/1357
- [ ] Improve translations to make attribute sharing more clear. Double check the JSON it produces.
- [x] *Created a new issue to pick this up later. See here;* https://github.com/openremote/openremote/issues/1359
- [ ] Check what is going on with the `negated` flag.

richturner:
Don't think this item:

Attribute picker doesn't highlight the asset types previously selected which makes it difficult to see the current selection

Is addressed with the new select all issue, maybe it's not clear what I mean?

richturner (commented):
UI seems OK (only scanned through code)

PR #1355, opened 2024-07-19
UI updates: Fix for scrollbar not working when switching dashboards

MartinaeyNL:
Small fix for scrollbar being grayed out upon navigating to a different dashboard.
This was caused by an HTML child container not being of the correct height; preventing an overflow.
Moved the "should always be 100% height" logic to CSS instead of on grid initialization, which fixes the issue.


MartinaeyNL:
As it is such a small change; verifying on a test deployment should be enough for this PR.

MartinaeyNL:
@DonWillems tested on a deployment, and discussed what additional changes to make.
In short; this change creates a new bug on "dragging widgets to below your screen while in edit mode".
Will look into this asap

PR #1364, opened 2024-07-26
UI Updates; fix for widget error flash on Insights

### LINKED ISSUE #1341: Insights: when loading widget data it should show a loader instead of 'No attributes connected'
Image

MartinaeyNL:
Fixes #1341 .

Previously, a flash of "no attributes connected" was shown when opening a dashboard.
This was caused by the attributes list being empty during the loading phase.

Changelog:
- Added loader for all asset-related widgets, using a `_loading` state and applying conditions to the `render()` function.
- Cleanup of some lifecycle logic, now using `willUpdate()` instead of the `updated()` hook.


MartinaeyNL:
Tested the UI together with @DonWillems, and confirmed its ready to review.

richturner (approved):

PR #1369, opened 2024-07-30
resizer of attribute-card was looking at height instead of width

DonWillems:
Fix bug where value font is changing size every 200ms.
More improvements are needed as the font is not properly scaled atm. Issue will be created.


MartinaeyNL:
(the issue Don created can be found in #1371)

PR #1372, opened 2024-07-30
Modify mode of assets and insights kept when switching visibility of page

DonWillems:
Reproduce:
- Navigate to assets or insights page
- Go into modify mode
- Switch tabs within the browser and come back
- Assets/Insights page has gone back to view mode.

This issue became apparent since the introduction of visible as app state. Because of the change in visibility the state params would be checked again. But the params were never set when entering editMode because _updateRoute was set to silent.

With this change the params will be set when toggling editMode and be correctly picked up on visibility change.


richturner (approved):

MartinaeyNL:
Don allowed me to press the merge button haha 👍

PR #1374, opened 2024-08-01
Dynamic injection of timestamps in HTTP agent

pankalog:
This PR introduces the feature of dynamically injecting time values relative to the current time and date in any HTTP request performed using an HTTP Agent. When a user inputs `${time:DURATION:FORMAT}` anywhere in a request, then that will be transformed into the correct timestamp and format as requested. For example, `${time:PT1H:yyyy-MM-dd HH:mm}` in a request would be translated to the year-month-day hour:minutes format, for the current time, just one hour further into the future. This is useful for users that may want to set a value to a remote device that will persist for a certain point in the future, maybe something like requesting the average of a value from a device starting from an hour in the past, etc. 

Implementation:
- The `DynamicTimeReplacementClientRequestFilter`, which intercepts all requests, and checks using RegEx for the string `${time:DURATION:FORMAT}`, where `DURATION` is an ISO8601 Duration, and FORMAT is a `SimpleDateFormat`.
- A Mechanism to allow custom `ClientRequestFilter`s to be injected into the HTTP client.

Tests for the parser and integration tests to be committed soon.


agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java
```diff
@@ -602,4 +613,68 @@ protected void cancelPolling(AttributeRef attributeRef) {
...
+
+					// Set the modified body back to the request context
+					requestContext.setEntity(modifiedBody);
+				}
+			}
+		}
+
+		private String replace(String str){
+			// Regex to match the pattern ${time:DURATION:FORMAT}
+			String regex = "\\$\\{time:([^:]+):([^}]+)\\}";
```
richturner, line 650: `DURATION` and `FORMAT` should be optional so a regex like: `\\$\\{time(?::([^:]+))?(?::([^}]+))?\\}`

Default for `DURATION` should just be `now()`
Default for `FORMAT` should just be epoch seconds

agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java
```diff
@@ -602,4 +613,68 @@ protected void cancelPolling(AttributeRef attributeRef) {
...
+		private DynamicTimeReplacementClientRequestFilter(TimerService timerService) {
+			this.timerService = timerService;
+		}
+		@Override
+		public void filter(ClientRequestContext requestContext) throws IOException {
+			if (requestContext.hasEntity()) {
+				Object entity = requestContext.getEntity();
+
+				if (entity instanceof InputStream) {
+					String body = new Scanner((InputStream) entity, StandardCharsets.UTF_8).useDelimiter("\\A").next();
```
richturner, line 629: injection of dynamic values should be done in the protocol as body content may not be text.

Also injection must be supported on the query params and maybe headers which is likely the more common requirement rather than in the body (think adding date range for a query)

richturner (changes requested):

agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java
pankalog: What exactly do you mean by "In the protocol"? I was not exactly sure where to inject the filter so I thought of doing it this way. What I can do is ensure that replacement happens for `requestContext.getEntity()`, `requestContext.getHeaders()`, and `requestContext.getProperty(QUERY_PARAMETERS_PROPERTY)`, to make sure that all of them are checked.

richturner: OK to clarify:

For body injection see `ProtocolUtil.doOutboundValueProcessing()` which is the generic protocol mechanism for formatting the body for outbound protocol requests.

For query parameter injection see `QueryParameterInjectorFilter.java`

For path injection see `HTTPProtocol.getRequestBuilder()`

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -549,4 +561,113 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
             HTTPProtocol.client.set(null)
         }
     }
+
+    def "Check HTTP client dynamic time feature"() {
```
richturner, line 565: Great to see you included tests 😃

container/src/main/java/org/openremote/container/web/WebTargetBuilder.java
```diff
@@ -199,7 +207,7 @@ public static ResteasyClient createClient(ExecutorService executorService, int c
             .connectionCheckoutTimeout(CONNECTION_CHECKOUT_TIMEOUT_MILLISECONDS, TimeUnit.MILLISECONDS)
             .readTimeout(overrideSocketTimeout, TimeUnit.MILLISECONDS)
             .connectTimeout(CONNECTION_TIMEOUT_MILLISECONDS, TimeUnit.MILLISECONDS)
-            .register(new JacksonConfig());
+		    .register(new JacksonConfig());
```
richturner, line 210: Indents

container/src/main/java/org/openremote/container/web/WebTargetBuilder.java
```diff
@@ -172,6 +176,10 @@ public ResteasyWebTarget build() {
             target.register(new FollowRedirectFilter());
         }
 
+		if(this.filters != null && !this.filters.isEmpty()){
+			this.filters.forEach(target::register);
+		}
+
```
richturner, line 182: Indents

container/src/main/java/org/openremote/container/web/WebTargetBuilder.java
```diff
@@ -68,6 +71,7 @@ public class WebTargetBuilder {
     protected Map<String, List<String>> injectHeaders;
     protected Map<String, List<String>> injectQueryParameters;
     protected boolean followRedirects = false;
+	protected List<ClientRequestFilter> filters = new ArrayList<>();
```
richturner, line 74: Indents

agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java
richturner: The logic should be reused in these places so a static public method will be needed

model/src/main/java/org/openremote/container/web/TimeParameterInjectorFilter.java
richturner, line 1: This isn't the right way to approach this. The `String replace(String val)` should just be a static method in `ProtocolUtil` called `doDynamicTimeReplace(String value)`.

Your CRF is then only in agent module and you just call the `doDynamicTimeReplace()` to do the processing on the path and query params.

`doOutboundValueProcessing()` should also use this static `doDynamicTimeReplace()` method for body substitution.

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -613,11 +616,10 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
         String[] dynamicTimeTests = ["\${time:PT1H:yyyy-MM-dd HH:mm}", "\${time:PT2H:yyyy-MM-dd HH:mm:ss}", "\${time:P1D:yyyy-MM-dd}", "\${time:P2DT3H:yyyy-MM-dd HH:mm}", "\${time:PT30M:yyyy-MM-dd HH:mm}", "\${time:P7D:yyyy-MM-dd}", "\${time:PT5H:yyyy-MM-dd HH:mm}", "\${time:P2D:yyyy-MM-dd}", "\${time:PT1M:yyyy-MM-dd}", "\${time:P3D:yyyy-MM-dd}"] as List<String>
 
         then: "The strings are parsed correctly. If there is a way to use a constant timestamp, it'd be much easier."
-        HTTPProtocol.DynamicTimeReplacementClientRequestFilter filter = new HTTPProtocol.DynamicTimeReplacementClientRequestFilter(timerService);
+        TimeParameterInjectorFilterImpl filter = new TimeParameterInjectorFilterImpl(timerService::getCurrentTimeMillis);
```
richturner, line 619: Missing assertions?

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -613,11 +616,10 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
         String[] dynamicTimeTests = ["\${time:PT1H:yyyy-MM-dd HH:mm}", "\${time:PT2H:yyyy-MM-dd HH:mm:ss}", "\${time:P1D:yyyy-MM-dd}", "\${time:P2DT3H:yyyy-MM-dd HH:mm}", "\${time:PT30M:yyyy-MM-dd HH:mm}", "\${time:P7D:yyyy-MM-dd}", "\${time:PT5H:yyyy-MM-dd HH:mm}", "\${time:P2D:yyyy-MM-dd}", "\${time:PT1M:yyyy-MM-dd}", "\${time:P3D:yyyy-MM-dd}"] as List<String>
 
         then: "The strings are parsed correctly. If there is a way to use a constant timestamp, it'd be much easier."
```
richturner, line 618: Not sure what is meant about a way to use a constant timestamp would be much easier

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -632,12 +634,19 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
                                                 .setMethod(HTTPMethod.POST)
                                                 .setWriteValue('{"prop1": "${time:PT1H:yyyy-MM-dd HH:mm}"}')
                                                 .setPollingMillis(200)
+//                                                .setHeaders(new MultivaluedStringMap(
+//                                                        [
+//                                                                ("header3") : ["'"+testReplacementString.toString()+"'"]
+//                                                        ]
+//                                                ))
```
richturner, line 641: Obsolete?

richturner (changes requested):

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
pankalog: Being able to somehow set a static time that is returned, so that I can generate the literal strings that should be returned instead of having to write code to adapt it to current time.

model/src/main/java/org/openremote/model/protocol/ProtocolUtil.java
```diff
@@ -63,12 +64,10 @@ protected ProtocolUtil() {
      * searched on every single write request (for performance reasons), instead this should be recorded when the
      * attribute is first linked.
      */
-    public static Pair<Boolean, Object> doOutboundValueProcessing(String assetId, AttributeInfo attribute, AgentLink<?> agentLink, Object value, boolean containsDynamicPlaceholder, TimeParameterInjectorFilter filter) {
+    public static Pair<Boolean, Object> doOutboundValueProcessing(String assetId, AttributeInfo attribute, AgentLink<?> agentLink, Object value, boolean containsDynamicPlaceholder) {
```
richturner, line 67: I don't see dynamic time replacement happening here....

model/src/main/java/org/openremote/model/protocol/ProtocolUtil.java
```diff
@@ -227,4 +217,55 @@ public static Consumer<String> createGenericAttributeMessageConsumer(String asse
...
         };
     }
+
+	public static String doDynamicTimeReplace(String value, Supplier<Long> currentMillisSupplier){
+		if (value == null) return value;
+
+		LOG.finest("doDynamicTimeReplace. Current millis: "+currentMillisSupplier.get()+ ". Parsed:" + Date.from(Instant.ofEpochSecond(currentMillisSupplier.get()/1000)));
+
+		// Regex to match the pattern ${time:DURATION:FORMAT}
+		String regex = "\\$\\{time(?::([^:]+))?(?::([^}]+))?\\}";
```
richturner, line 227: Move to constants next to dynamic value regexp

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -616,9 +617,8 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
         String[] dynamicTimeTests = ["\${time:PT1H:yyyy-MM-dd HH:mm}", "\${time:PT2H:yyyy-MM-dd HH:mm:ss}", "\${time:P1D:yyyy-MM-dd}", "\${time:P2DT3H:yyyy-MM-dd HH:mm}", "\${time:PT30M:yyyy-MM-dd HH:mm}", "\${time:P7D:yyyy-MM-dd}", "\${time:PT5H:yyyy-MM-dd HH:mm}", "\${time:P2D:yyyy-MM-dd}", "\${time:PT1M:yyyy-MM-dd}", "\${time:P3D:yyyy-MM-dd}"] as List<String>
 
         then: "The strings are parsed correctly. If there is a way to use a constant timestamp, it'd be much easier."
-        TimeParameterInjectorFilterImpl filter = new TimeParameterInjectorFilterImpl(timerService::getCurrentTimeMillis);
         for (final String e in dynamicTimeTests){
-            def y = filter.replace(e);
+            def y = ProtocolUtil.doDynamicTimeReplace(e, timerService::getCurrentTimeMillis);
 
         }
```
richturner, line 623: There is no assertion here and you mention constant timestamp, as you now know, the clock in tests can be stopped/started and set to any time

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -668,16 +673,27 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
...
+
+
         then: "The attribute has the correct value"
         conditions.eventually{
+            mockServer.latestDynamicTimeRequest
             asset = assetStorageService.find(asset.getId(), true)
             getLOG().warn("Polls: "+ mockServer.pollCountFast);
-            assert asset.getAttribute("testPostRequest").flatMap({it.value}).orElse(null) == '{"prop1": "'+expected+'"}';
+//            assert asset.getAttribute("testPostRequest").flatMap({it.value}).orElse(null) == '{"prop1": "'+expected+'"}';
+            assert true;
```
richturner, line 684: ?

richturner (changes requested):
Please see all my comments and and try and ensure indents follow our standard (i.e. what is used in 90% of the code) and don't have commits that just change indents makes more unneccessary work reviewing.

Don't want to see commented out patches of new code either.

container/src/main/java/org/openremote/container/web/DynamicTimeInjectionFilter.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Objects;
+import java.util.function.Supplier;
+import java.util.regex.Matcher;
+import java.util.stream.Collectors;
+
+import static org.openremote.container.web.QueryParameterInjectorFilter.DYNAMIC_TIME_PATTERN;
+import static org.openremote.container.web.QueryParameterInjectorFilter.DYNAMIC_VALUE_PROPERTY;
+
+@Provider
+@Priority(1)
```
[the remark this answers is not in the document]
pankalog: Because for some reason I am getting this error: `RESTEASY003720: path param time has not been provided by the parameter map` when the time filter is applied after the value filter. What I believe is causing this, is that the value filter actually adds the path parameters to the URI. When the value filter is ran before the time filter, then the path parameters that have dynamic time will be attempted to be inserted without injecting the dynamic time, causing that error (hence `time`, which is the text that we have to indicate dynamic time on the value).

container/src/main/java/org/openremote/container/web/QueryParameterInjectorFilter.java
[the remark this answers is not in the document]
pankalog: You mean I should remove the time-related parts in the QueryParameterInjectorFilter?

agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java
```diff
@@ -346,7 +377,7 @@ protected void doStart(Container container) throws Exception {
         webTargetBuilder.followRedirects(followRedirects);
 
         LOG.fine("Creating web target client '" + baseUri + "'");
-        webTarget = webTargetBuilder.build();
+        webTarget = webTargetBuilder.build(new DynamicTimeInjectionFilter(timerService::getCurrentTimeMillis), new QueryParameterInjectorFilter());
```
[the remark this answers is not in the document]
pankalog: Because the WebTargetBuilder instantiates the filters by itself and applies them to the target. Thing is, the WebTargetTest fails if we don't supply both filters. The DynamicTimeFilter requires a currentMillisSupplier, so just  giving it the static reference doesn't do. Not all WebTargetBuilders require the filters, so injecting them where needed is a good way. What I can do though is at least use a static reference to the queryParameterInjectorFilter, and then inject just the dynamic time filter.

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -549,4 +563,136 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
...
+            agent = assetStorageService.find(agent.id, HTTPAgent.class)
+            assert agent.getAgentStatus().orElse(ConnectionStatus.DISCONNECTED) == ConnectionStatus.CONNECTED
+        }
+
+        when: "An array of dynamic time strings is tested with the string replacement function"
+
+        String[] dynamicTimeTests = ["\${time:PT1H:yyyy-MM-dd HH:mm}", "\${time:PT2H:yyyy-MM-dd HH:mm:ss}", "\${time:P1D:yyyy-MM-dd}", "\${time:P2DT3H:yyyy-MM-dd HH:mm}", "\${time:PT30M:yyyy-MM-dd HH:mm}", "\${time:P7D:yyyy-MM-dd}", "\${time:PT5H:yyyy-MM-dd HH:mm}", "\${time:P2D:yyyy-MM-dd}", "\${time:PT1M:yyyy-MM-dd}", "\${time:P3D:yyyy-MM-dd}"] as List<String>
+
+        then: "The strings are parsed correctly. If there is a way to use a constant timestamp, it'd be much easier."
+        for (final String e in dynamicTimeTests){
```
richturner, line 620: Not a big fan of unit testing like this esp inside an integration test. As this is doing nothing then it should be removed.

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -549,4 +563,136 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
...
+
+
+        then: "The attribute has the correct value"
+        conditions.eventually{
+            mockServer.latestDynamicTimeRequest
+            asset = assetStorageService.find(asset.getId(), true)
+            getLOG().warn("Polls: "+ mockServer.pollCountFast);
+            assert asset.getAttribute("testPostRequest").flatMap({it.value}).orElse(null) == '{"prop1": "'+expected+'"}';
+        }
+        and: "The request that was sent contains the correct body, headers, and query parameters"
```
richturner, line 685: Pay attention to line breaks and un-necessary line breaks

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -549,4 +563,136 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
             HTTPProtocol.client.set(null)
         }
     }
+
+    def "Check HTTP client dynamic time feature"() {
```
richturner, line 567: This test doesn't cover the different scenarios:

* `{$time}` Only
* time with +/-offset
* time with no offset but a format request

agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java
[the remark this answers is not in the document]
richturner: This comment still needs answering

richturner (changes requested):

test/src/test/groovy/org/openremote/test/protocol/http/WebTargetTest.groovy
```diff
@@ -508,7 +510,7 @@ class WebTargetTest extends Specification {
         and: "a web target for a server"
         def target = new WebTargetBuilder(client, new URIBuilder("https://dynamicparamserver").build())
             .setInjectQueryParameters(params)
-            .build()
+            .build(new DynamicTimeInjectionFilter(System::currentTimeMillis), new QueryParameterInjectorFilter())
```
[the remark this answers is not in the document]
pankalog: I tried doing that, but this test is not made inside a container, so I don't have access to any service.

test/src/test/groovy/org/openremote/test/protocol/http/WebTargetTest.groovy
```diff
@@ -475,7 +477,7 @@ class WebTargetTest extends Specification {
         and: "a web target for a server"
         def target = new WebTargetBuilder(client, new URIBuilder("https://paramserver").build())
             .setInjectQueryParameters(params)
-            .build()
+            .build(new DynamicTimeInjectionFilter(System::currentTimeMillis), new QueryParameterInjectorFilter())
```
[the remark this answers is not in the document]
pankalog: Same as above

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
pankalog: Agreed. The comment about having the timer be stable was so that I can generate a list of test cases and their predefined 100% correct results, and then comparing each other. But I think that an integration test that covers the required test cases should be enough.

agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java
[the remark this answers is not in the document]
richturner: `DynamicTimeFilter` could use a no arg constructor and then inject the time into the request context properties like the dynamic value for the `QueryParameterInjectorFilter` only if the property is set do you then search for `{$time}` strings to replace

container/src/main/java/org/openremote/container/web/QueryParameterInjectorFilter.java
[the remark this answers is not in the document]
richturner: Yes as it is just a basic implementation of what you are working on

richturner, line 1: We discussed removing the old basic time injection from this filter

richturner (changes requested):

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -549,4 +568,159 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
...
+            assert agent.getAgentStatus().orElse(ConnectionStatus.DISCONNECTED) == ConnectionStatus.CONNECTED
+        }
+
+
+        when: "An asset is created"
+        def asset = new ThingAsset("Test Asset")
+                .setParent(agent)
+                .addOrReplaceAttributes(
+                        // attribute that polls the server using GET and uses regex filter on response
+                        new Attribute<>("testPostRequest", TEXT)
```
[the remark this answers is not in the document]
pankalog: I rewrote a bit of the `doDynamicTimeReplace` function to make it easier and to add this functionality. <default> and `EPOCH_SECONDS` will return `instant.getEpochSecond()`, `EPOCH_MILLIS` will return `instant.toEpochMilli()`.  Also made it a bit expandable down the line.

Also added a new Attribute on the testAsset to test the default, `EPOCH_SECONDS`, and `EPOCH_MILLIS` at the end of the test.

agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java
```diff
@@ -196,16 +205,16 @@ protected Invocation.Builder getRequestBuilder(String value) {
...
-
             return requestBuilder;
         }
 
         protected Invocation buildInvocation(Invocation.Builder requestBuilder, String value) {
             Invocation invocation;
 
+            // Dynamic time Injection
+            value = ProtocolUtil.doDynamicTimeReplace(value, this.currentMillisSupplier);
+            requestBuilder.property(DynamicTimeInjectionFilter.CURRENT_MILLIS_SUPPLIER_PROPERTY, this.currentMillisSupplier);
```
richturner, line 216: Add a `dynamicTime` flag (like `dynamicQueryParameters`) and do a super quick string.contains(`{$time`) on the headers and query params and then only set `DynamicTimeInjectionFilter.CURRENT_MILLIS_SUPPLIER_PROPERTY`, this will avoid the filter doing lots of regex work for no reason (as most uses of the protocol won't be using dynamic time).

container/src/main/java/org/openremote/container/web/QueryParameterInjectorFilter.java
richturner, line 1: Rename this file `DynamicValueInjectionFilter` and add support for headers like you have done in `DynamicTimeInjectionFilter`

richturner (changes requested):

agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java
pankalog: Fixed, though I was kind of saddened by how the Jakarta API doesn't quite work well with requestContexts. I was expecting something that is a bit more developer-friendly. I would be happy to write some convoluted Predicate to make stuff a bit easier to work with for both headers and query parameters, but I assume that there are other, more pertinent matters to deal with 😅

richturner:
@pankalog Sorry to hijack your commit but I saw a couple of edge case issues (more general) and I wasn't happy with the way the filters worked and interdependency so I have cleaned up the logic but to avoid the JAX-RS issues around curly braces I have changed the dynamic syntax and will update docs accordingly.

richturner (approved):

PR #1382, opened 2024-08-12
Add support updating response headers to `WebTargetBuilder`

wborn:
This allows for removing response headers or adding new ones.


container/src/main/java/org/openremote/container/web/WebTargetBuilder.java
```diff
@@ -103,6 +104,11 @@ public WebTargetBuilder setInjectQueryParameters(Map<String, List<String>> injec
         return this;
     }
 
+    public WebTargetBuilder setUpdateResponseHeaders(Map<String, List<String>> updateResponseHeaders) {
```
richturner, line 107: Just rename to `setOverrideResponseHeaders()`

richturner (changes requested):

richturner (approved):

PR #1397, opened 2024-09-04
Use parallel Gradle builds

wborn:
I gave [parallel Gradle executution](https://docs.gradle.org/current/userguide/performance.html#parallel_execution) a test and it seems to work fine so far.
It puts my abundance of CPU cores to use making builds 45% faster (45s instead of 1m 17s).

The gain comes mostly from running the NPM tasks in parallel.

Sequential execution:

Screenshot from 2024-09-04 10-39-32

Parallel execution:

Screenshot from 2024-09-04 10-39-21

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,5 @@
 org.gradle.jvmargs=-Xms256m -Xmx1024m
+org.gradle.parallel=true
 
 jaxrsSpecVersion = 3.1.0
 javaxAnnotationsVersion = 2.1.1


richturner (commented):
Does this have an impact on the tests? They cannot run in parallel

wborn:
The parallel builds also seem to work fine after testing them on several custom projects. The same option needs to be added to the `gradle.properties` at the root level before it is considered.

Running the tests also keeps working because it runs tasks sequentially in projects:

Screenshot from 2024-09-05 10-06-35

richturner (approved):

PR #1414, opened 2024-09-12
Persist manager_config.json across deployments

pankalog:
To do so, we use the `PersistenceService`'s `getStorageDir()` to store it there.

The mindset behind the changes is that we initially look into the storageDir, and if it's not present there, then fall back into the deployment path. 

Unfortunately, due to the way that things are set up at the moment, usage of the resource through the rest API client is not possible. 

I'm not sure what and how to document this, open to any recommendations!


manager/src/main/java/org/openremote/manager/web/ManagerWebService.java
```diff
@@ -214,6 +225,7 @@ public void init(Container container) throws Exception {
         }
 
         PathHandler deploymentHandler = defaultHandler != null ? new PathHandler(defaultHandler) : new PathHandler();
+        HttpHandler managerConfigFileHandler = createFileHandler(container, storageDir, null);
```
richturner, line 228: Not needed anymore

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -113,4 +115,13 @@ public void saveImageFile(String path, FileInfo fileInfo) {
...
     }
+
+    public Optional<File> getManagerConfig(){
+        File file = persistenceService.getStorageDir().resolve("manager").resolve("manager_config.json").toFile();
+
+        if(file.exists() && !file.isDirectory()){
+           return Optional.of(file);
+        }
+        return Optional.empty();
+    }
```
richturner, line 126: Should fallback to look in `OR_CUSTOM_APP_DOCROOT` if doesn't exist in storageDir so that it can be provided in the deployment image if desired.

manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
```diff
@@ -43,13 +52,71 @@ public ConfigurationResourceImpl(TimerService timerService, ManagerIdentityServi
...
+            if (jsonNode.isObject()) {
+                ObjectNode objectNode = (ObjectNode) jsonNode;
+                return objectNode;
+            } else {
+                throw new IOException("The provided JSON is not an object.");
+            }
+        } catch (IOException e) {
+            LOG.severe("Error reading manager config file: " + e.getMessage());
+        }
+        return null;
```
richturner, line 92: FYI: We always use the existing ObjectMapper available as `ValueUtil.JSON`

I would cache the ObjectNode to avoid deserialisation on every request

manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
```diff
@@ -43,13 +52,71 @@ public ConfigurationResourceImpl(TimerService timerService, ManagerIdentityServi
...
+
+            String mimeType = Files.probeContentType(imageFile.toPath());
+            if (mimeType == null) {
+                mimeType = "application/octet-stream";
+            }
+
+            Response.ResponseBuilder response = Response.ok(imageFile, mimeType);
+            response.header("Access-Control-Allow-Origin", "*");
+            response.header("Access-Control-Allow-Methods", "GET, POST, DELETE, PUT");
+            response.header("Access-Control-Allow-Headers", "Content-Type, api_key, Authorization");
```
richturner, line 112: CORS headers should be automatically added and are configured with environment variables, we don't just let any origin make calls to our API

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -82,9 +108,9 @@ public void stop(Container container) throws Exception {
     }
 
 
-    public void saveMangerConfig(Object managerConfiguration) {
+    public void saveManagerConfigFile(Object managerConfiguration) {
```
richturner, line 111: I know it wasn't there before but this should throw an exception back to the reponse (500 server error)

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -109,8 +135,68 @@ public void saveImageFile(String path, FileInfo fileInfo) {
         try (OutputStream out = new FileOutputStream(file)) {
             out.write(CodecUtil.decodeBase64(fileInfo.getContents()));
         } catch (IOException | SecurityException exception) {
-            LOG.log(Level.WARNING, "Error when saving image in manger_config.json", exception);
+            LOG.log(Level.WARNING, "Error when saving image in manager_config.json", exception);
```
richturner, line 138: I know it wasn't there before but this should throw an exception back to the reponse (500 server error)

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -82,9 +108,9 @@ public void stop(Container container) throws Exception {
     }
 
 
-    public void saveMangerConfig(Object managerConfiguration) {
+    public void saveManagerConfigFile(Object managerConfiguration) {
         LOG.log(Level.INFO, "Saving manager_config.json..");
-        try (OutputStream out = new FileOutputStream(pathPublicRoot + "/manager_config.json")) {
+        try (OutputStream out = new FileOutputStream(persistenceService.getStorageDir().resolve("manager").resolve("manager_config.json").toFile())) {
```
richturner, line 113: Be consistent use a method `getManagerConfigPath()` which is then called by `getManagerConfiguImagePath()` and then both are used in all path references

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -109,8 +135,68 @@ public void saveImageFile(String path, FileInfo fileInfo) {
...
+        }
+
+    }
+
+    public Optional<File> getManagerConfigImage(String filename){
+        File file = getManagerConfigImagePath().resolve(filename).toFile();
+
+        // fallback to OR_CUSTOM_APP_DOCROOT
+        if(!file.isFile()){
+            file = pathPublicRoot.resolve("images").resolve(filename).toFile();
```
richturner, line 148: Need to make sure someone can't use a back reference to navigate around the storageDir e.g. filename `../../otherfile.txt`

deployment/manager/app/manager_config.json
```diff
@@ -6,13 +6,13 @@
     },
     "smartcity": {
       "styles": ":host > * {--or-app-color2: #F6F7FA; --or-app-color3: #1B1464; --or-app-color4: #662D91;}",
-      "logo": "/images/logo_smartcity.svg",
-      "logoMobile": "/images/logo-mobile_smartcity.svg"
+      "logo": "api/master/configuration/manager/image/logo_smartcity.svg",
```
richturner, line 9: It's ok to change the paths for new config but we have configs already deployed on instances which will be using `/images/...` paths. When the manager config is saved the images should be copied to the storageDir() also otherwise the images could still be lost on redeploy.

pankalog: @richturner I was thinking of performing a test when updating the manager_config file. Basically, check that the files referenced as images in the config are retrievable through the storageDir, and if they're not, check if they are retrievable from the deployment path. If that happens, copy them into the storageDir. I can't think of a more efficient way, any other ideas?

pankalog: @richturner refer to `ConfigurationService#checkAndFixImageReferences`

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
pankalog: L241 does that now, if you find that okay

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -140,63 +228,80 @@ public void saveImageFile(String path, FileInfo fileInfo) {
...
+                        File deploymentImageFile = path.toAbsolutePath().toFile();
+
+                        // This file is in the deployment folder, copy the file to the persistenceImageFile.
+                        if (deploymentImageFile.isFile()) {
+
+                            deploymentImageFile.mkdirs();
+                            deploymentImageFile.createNewFile();
+
+                            // Copy the file
+                            try (InputStream in = new FileInputStream(deploymentImageFile);
```
richturner, line 285: `Files.copy` is much simpler in java8+

deployment/manager/app/manager_config.json
richturner, line 9: These paths should probably be left as before as it is clearer that they relate to the images dir in the deployment also

manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
```diff
@@ -42,14 +48,59 @@ public ConfigurationResourceImpl(TimerService timerService, ManagerIdentityServi
...
 
     @Override
-    public Object update(RequestParams requestParams, Object managerConfiguration) {
-        this.configurationService.saveMangerConfig(managerConfiguration);
+    public Object update(RequestParams requestParams, ObjectNode managerConfiguration) {
+        try {
+            this.configurationService.saveManagerConfigFile(managerConfiguration);
+        } catch (Exception e) {
+            LOG.warning("Couldn't store image:" +e.getMessage());
+            return Response.status(Response.Status.INTERNAL_SERVER_ERROR).entity("Error storing image").build();
```
richturner, line 56: Wrong error message

manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
```diff
@@ -42,14 +48,59 @@ public ConfigurationResourceImpl(TimerService timerService, ManagerIdentityServi
...
     }
 
     @Override
     public String fileUpload(RequestParams requestParams, String path, FileInfo fileInfo) {
-        this.configurationService.saveImageFile(path, fileInfo);
-        return path;
+        try {
+            this.configurationService.saveConfigImageFile(path, fileInfo);
+        } catch (Exception e) {
+            throw new WebApplicationException(e);
```
richturner, line 66: Use same error and logging as `update` method

manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
```diff
@@ -42,14 +48,59 @@ public ConfigurationResourceImpl(TimerService timerService, ManagerIdentityServi
...
+        return managerConfigPath.toString();
+    }
+
+    @Override
+    public ObjectNode getManagerConfig() {
+        return configurationService.getManagerConfig();
+    }
+
+    @Override
+    public Object getManagerConfigImages(String fileName) {
```
richturner, line 84: Should be `getManagerConfigImage`

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -69,33 +87,129 @@ public void init(Container container) throws Exception {
...
+
+
+    public void saveManagerConfigFile(ObjectNode managerConfiguration) throws Exception {
         LOG.log(Level.INFO, "Saving manager_config.json..");
-        try (OutputStream out = new FileOutputStream(pathPublicRoot + "/manager_config.json")) {
+        try (OutputStream out = new FileOutputStream(getManagerConfigFile().orElseThrow())) {
             out.write(ValueUtil.JSON.writeValueAsString(managerConfiguration).getBytes());
+
+            // Check references to images
+
```
richturner, line 201: `checkAndFixImageReferences` not actually called

richturner (changes requested):

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -92,34 +100,114 @@ public void init(Container container) throws Exception {
...
+        loadManagerConfigJson();
     }
 
     @Override
     public void start(Container container) throws Exception {
-        /* code not overridden yet */
+
+        // Check if the configuration contains references to files that are located in the deployment, and if they are,
+        // move them to the storageDir.
+        ObjectNode changedConfig = this.checkAndFixImageReferences(this.managerConfig);
```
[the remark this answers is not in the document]
pankalog: This is mostly there as a "migration" script, as an automatic way to move the images when this code is first ran (e.g. when old projects get this feature). Because when running the manager with these changes for the first time, we are required to move the references so that next migrations don't delete the files.

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -69,33 +81,144 @@ public void init(Container container) throws Exception {
...
+    }
+
+    public Path getManagerConfigPath(){
+        return persistenceService.getStorageDir()
+                .resolve("manager")
+                .resolve("manager_config.json");
+    }
+
+
+    protected void loadManagerConfigJson() throws IOException {
```
richturner, line 165: It is perfectly valid to not have a manager config so exceptions shouldn't be thrown or severe log messages, info is appropriate.

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
richturner: This is still outstanding

richturner (changes requested):
Lots of my previous comments are still outstanding and no questions on the comments so not clear what you want me to do here.

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
pankalog: This has been changed as we now cache the manager, and when we do get the manager, and I do this in `getManagerConfigFile`.

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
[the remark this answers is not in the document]
pankalog: If the `changedConfig` variable is `null`, both the images and the manager_config are copied over to the storageDir right after that, so the migration of both the manager_config and its images occur at startup.

deployment/manager/app/manager_config.json
pankalog: After thinking about it, since I had to edit most of the frontend references to images, I am actually thinking of changing the frontend so that we don't actually require a change in the manager_config. So, the manager_config references would remain the same.

manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
pankalog: Forgot to push that commit, oops!

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
pankalog: Fixed, logging set to `INFO`

deployment/manager/app/manager_config.json
pankalog: It's fixed, but I don't include the `images` part, so it's just a reference to what the file should be within the `storageDir/manager/images` directory.

richturner (changes requested):

manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
```diff
@@ -42,14 +47,60 @@ public ConfigurationResourceImpl(TimerService timerService, ManagerIdentityServi
...
 
     @Override
-    public String fileUpload(RequestParams requestParams, String path, FileInfo fileInfo) {
-        this.configurationService.saveImageFile(path, fileInfo);
-        return path;
+    public Object fileUpload(RequestParams requestParams, String path, FileInfo fileInfo) {
+        try {
+            this.configurationService.saveConfigImageFile(path, fileInfo);
+        } catch (Exception e) {
+            LOG.warning("Couldn't store custom manager_config.json image: " +e.getMessage());
```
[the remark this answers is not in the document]
pankalog: What exactly would you expect as an error message here?

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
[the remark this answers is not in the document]
pankalog: But this would mean that currently-deployed projects will not be migrated to the `storageDir`. Would you like to add the feature as an env var or something? Because I thought that was part of the reason of this PR, so that users don't lose their configuration upon upgrading to a newer version.

ui/component/or-app/src/index.ts
```diff
@@ -371,12 +371,16 @@ export class OrApp<S extends AppStateKeyed> extends LitElement {
 
             if (!this._config.logo) {
                 this._config.logo = DefaultLogo;
+            }else{
```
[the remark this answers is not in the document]
pankalog: Not sure what you mean by this? So should it be the URL for the GET request to the backend, or the location of the file in the images directory? The hardcoded part I can't do much, since `manager.rest` is not initialized by this point in the startup sequence and this is the most central part I can edit these. Any ideas?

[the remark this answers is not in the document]
richturner: I don't understand why the URL doesn't contain the correct path in the first place? All paths in the config should be absolute and not require manipulation.

manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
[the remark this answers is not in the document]
richturner: Sorry didn't even see the word image in the message, change to this:

`"Save image failed '" + path + "': " + e.getMessage()`

ui/component/or-app/src/index.ts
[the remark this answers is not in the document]
pankalog: I'm not sure I understand what you mean. How are the paths in the config manipulated? The config contains the paths to the images from the images directory in storageDir. The string you see there is simply there to ensure that the correct API endpoint is called. If you want to I can reduce the amount of times we hardcode the string, but I am guessing that's not what you mean.

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -318,7 +318,8 @@ export class PageConfiguration extends Page<AppStateKeyed> {
         // POST manager config if changed...
         if (this.managerConfigurationChanged) {
             managerPromise = manager.rest.api.ConfigurationResource.update(config).then(() => {
-                fetch(this.urlPrefix + "/manager_config.json", {cache: "reload"});
+                // Not sure what this does and how to adapt it to fetch from axios, so I just replaced the urlPrefix with the managerUrl
+                fetch(this.managerConfiguration.manager.managerUrl + "/api/master/configuration/manager", {cache: "reload"});
```
richturner, line 322: Result is not used and anyway why is this needed? The ConfigurationResource.update should return the updated configuration and this should be used.

ui/component/or-app/src/index.ts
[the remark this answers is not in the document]
richturner: I think this makes the most sense:

When user saves manager config
- All image paths (if they don't already start with ```/api/master/configuration/manager/image/``` then change them to this
- If image doesn't exist in storage dir then check in deployment dir and copy it across otherwise fail save

Frontend should not have to add any prefix to image URLs.

richturner (changes requested):

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -69,48 +82,227 @@ public void init(Container container) throws Exception {
...
+                if (!isValid) return Optional.empty();
+                } catch (IOException e) {
+                    return Optional.empty();
+                }
         }
 
+        return file.isFile() ? Optional.of(file) : Optional.empty();
+    }
+
+    public ObjectNode checkAndFixImageReferences(ObjectNode managerConfig) throws Exception{
```
richturner, line 261: Should be protected, same for any other methods only used internally

ui/app/insights/src/index.ts
```diff
@@ -30,7 +29,7 @@ export const DefaultAppConfig: AppConfig<RootState> = {
 };
 
 // Try and load the app config from JSON and if anything is found amalgamate it with default
-const configURL = (CONFIG_URL_PREFIX || "") + "/manager_config.json";
+const configURL = (MANAGER_URL || "") + "/api/master/configuration/manager";
```
richturner, line 32: The `getManagerConfig` endpoint can return `null` as a valid response but there is no null check here (previously the fileservlet would have thrown a 404 but I like the server returning a null as you don't see an error in the browser console but you need to make sure you handle that situation.

ui/app/insights/src/index.ts
```diff
@@ -54,13 +53,6 @@ fetch(configURL).then(async (result) => {
         }
     }
 
-    // Add config prefix if defined (used in dev)
```
richturner, line 57: Any reason you have removed this?

ui/component/or-app/src/index.ts
```diff
@@ -28,8 +28,8 @@ export * from "./or-header";
 export * from "./types";
 
 // Declare MANAGER_URL and KEYCLOAK_URL - Global var injected by webpack
-declare var MANAGER_URL: string | undefined;
-declare var KEYCLOAK_URL: string | undefined;
+declare const MANAGER_URL: string;
+declare const KEYCLOAK_URL: string;
```
richturner, line 32: These can be `undefined`

ui/app/manager/src/index.ts
```diff
@@ -48,6 +48,7 @@ import { ManagerAppConfig } from "@openremote/model";
 import {pageGatewayTunnelProvider} from "./pages/page-gateway-tunnel";
 
 declare var CONFIG_URL_PREFIX: string;
+declare var MANAGER_URL: string;
```
richturner, line 51: Can be `undefined` good practice to declare all possible scenarios so type checking can make sure you handle the different situations

richturner (changes requested):

deployment/manager/app/manager_config.json
richturner: Backtracking here but I've put all images in this demo `manager_config.json` to the API image endpoint for consistency as they should correctly resolve.

ui/app/insights/src/index.ts
pankalog: I don't exactly know how translations work but I assumed that there wouldn't be a need to use CONFIG_URL_PREFIX now that the managerConfig is served by the backend. Do you think that's not the case?

pankalog: I've reverted that just in case.

richturner (approved):

richturner:
Will deploy this to staging demo for @DonWillems and @pierrekil to test

richturner (approved):

pankalog:
@wborn this fixes the issue encountered with the map settings, can we get this deployed somewhere so that @pierrekil can take a look at it? Thanks!

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -69,48 +85,247 @@ public void init(Container container) throws Exception {
...
+        File file = getManagerConfigImagePath().resolve(filename).toFile();
+
+        if(!file.isFile()){
+            // fallback to OR_CUSTOM_APP_DOCROOT
+            file = pathPublicRoot.resolve("images").resolve(filename).toFile();
+        }else{
+            try {
+                //Check if the file retrieved is somewhere within the storageDir/manager directory.
+                //If it is, return an Optional.empty file to denote Not Found
+                Boolean isValid = file.getCanonicalPath().contains(getManagerConfigImagePath().toFile().getCanonicalPath() + File.separator);
```
wborn, line 271:
```suggestion
                boolean isValid = file.getCanonicalPath().contains(getManagerConfigImagePath().toFile().getCanonicalPath() + File.separator);
```

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -69,48 +85,247 @@ public void init(Container container) throws Exception {
...
 
     @Override
     public void stop(Container container) throws Exception {
         /* code not overridden yet */
     }
 
+    @SuppressWarnings("unchecked")
+    protected ConcurrentMap<String, ObjectNode> getMapSettings(){
+        return (ConcurrentMap<String, ObjectNode>) getMapConfig().get("settings");
+    }
```
wborn, line 136: This seems to be unused?

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -69,48 +85,247 @@ public void init(Container container) throws Exception {
...
+                mapSettingsPath = persistencePath;
+            }
+        }
+        if(Files.isRegularFile(persistencePath)) {
+            mapSettingsPath = persistencePath;
+        }
+        LOG.info("Configuration Service Used files:");
+        LOG.info("\t- manager_config.json: " + getManagerConfigPath().toAbsolutePath().toString());
+        LOG.info("\t- mapsettings.json: " + mapSettingsPath.toAbsolutePath().toString());
+        LOG.info("\t- mapdata.mbtiles: " + mapTilesPath.toAbsolutePath().toString());
```
wborn, line 121: Redundant `toString` calls:

```suggestion
        LOG.info("\t- manager_config.json: " + getManagerConfigPath().toAbsolutePath());
        LOG.info("\t- mapsettings.json: " + mapSettingsPath.toAbsolutePath());
        LOG.info("\t- mapdata.mbtiles: " + mapTilesPath.toAbsolutePath());
```

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -69,48 +85,247 @@ public void init(Container container) throws Exception {
...
+        mapTilesPath = Paths.get(getString(container.getConfig(), OR_MAP_TILES_PATH, defaultMapTilesPath.toAbsolutePath().toString()));
+        mapTilesPath = Files.isRegularFile(mapTilesPath) ? mapTilesPath : defaultMapTilesPath;
+
+        mapSettingsPath = Paths.get(getString(container.getConfig(), OR_MAP_SETTINGS_PATH, defaultMapSettingsPath.toAbsolutePath().toString()));
+        Path persistencePath = persistenceService.getStorageDir().resolve("manager").resolve("mapsettings.json").toAbsolutePath();
+
+        if(!Files.isRegularFile(mapSettingsPath)) {
+            if(!Files.isRegularFile(persistencePath)){
+                persistencePath.getParent().toFile().mkdirs();
+                LOG.warning("Copying defaults to and using " + persistencePath.toString() + " for mapsettings.json");
```
wborn, line 110:
```suggestion
                LOG.warning("Copying defaults to and using " + persistencePath + " for mapsettings.json");
```

richturner (changes requested):

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -173,19 +204,22 @@ public Path getManagerConfigImagePath(){
         return this.persistenceService.resolvePath("manager").resolve("images");
     }
 
-    public void saveMapConfigFile(Map<String, MapRealmConfig> mapConfiguration) {
+    public void saveMapConfigFile(ObjectNode mapConfiguration) {
```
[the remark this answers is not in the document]
pankalog: So what happens here is that we have "mapconfig" and "mapsettings" if we can put it that way; mapconfig is the whole JSON file that is being saved. That is what contains the realm configs. This method is being used only in `saveMapConfig`, which adds the new realm configs to `this.mapConfig` and then uses `saveMapConfigFile` to perform the actual file ops which overwrite the file with the contents of `this.mapConfig`. So we don't only save the map realm configs, but we save the whole file, hence `ObjectNode`.

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -127,6 +140,20 @@ protected static Metadata getMetadata(Connection connection) {
         return metadata;
     }
 
+    @SuppressWarnings("unchecked")
+    protected static ConcurrentMap<String, MapRealmConfig> getMapConfigRealms(ObjectNode mapRealms){
```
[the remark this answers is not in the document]
pankalog: It's a casting helper for converting the "options" JSON node from `ObjectNode` to `(ConcurrentMap<String, ObjectNode>)`

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -127,6 +140,20 @@ protected static Metadata getMetadata(Connection connection) {
...
+        return (ConcurrentMap<String, MapRealmConfig>) ValueUtil.JSON.convertValue(mapRealms, ConcurrentMap.class);
+    }
+
+    protected static ObjectNode loadMapSettingsJson() {
+        return configurationService.getMapConfig();
+    }
+
+    @SuppressWarnings("unchecked")
+    protected ConcurrentMap<String, ObjectNode> getMapSettings(){
+        return (ConcurrentMap<String, ObjectNode>) ValueUtil.JSON.convertValue(this.mapConfig, ConcurrentMap.class);
```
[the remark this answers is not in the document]
pankalog: It's used when calling `HTTP GET /map`, and I don't really understand what happens there, since it dynamically builds the Mapbox GL settings from mapsettings.json

ui/app/insights/src/index.ts
```diff
@@ -30,10 +30,10 @@ export const DefaultAppConfig: AppConfig<RootState> = {
 };
 
 // Try and load the app config from JSON and if anything is found amalgamate it with default
-const configURL = (MANAGER_URL || "") + "/api/master/configuration/manager";
+const configURL =  (typeof MANAGER_URL !== 'undefined' ? MANAGER_URL : "") + "/api/master/configuration/manager";
 
 fetch(configURL).then(async (result) => {
-    if (!result.ok) {
+    if (!result.ok || result.status === 204) {
```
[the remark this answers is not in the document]
pankalog: I do that right below, if `appConfig === null` will then return DefaultAppConfig to the .then function

richturner:
Having checked this out locally I feel the scope of the changes is larger than what is needed to achieve the goals of the issue particularly around map settings (also given map server will be exernalised soon) so I am making some adjustments to get this issue off the backlog.

richturner (dismissed):

wborn:
I retriggered the build as it seems to fail on the `MQTTClientProtocolTest` which I've also seen failing in other PRs (#1528)

richturner (approved):

PR #1419, opened 2024-09-16
Clean up rules header

DonWillems:


ui/component/or-rules/src/or-rule-list.ts
```diff
@@ -615,6 +616,21 @@ export class OrRuleList extends translate(i18next)(LitElement) {
 
     }
 
+    protected _onSettingsClicked() {
+        if (this._globalRulesets) {
+            showOkCancelDialog("Global rules settings", html`
```
wborn, line 621: Can you make the string translatable using `i18next.t(...)` which is used elsewhere?

ui/component/or-rules/src/or-rule-viewer.ts
```diff
@@ -20,7 +20,7 @@ import {translate} from "@openremote/or-translate";
 import {InputType, OrInputChangedEvent, OrMwcInput} from "@openremote/or-mwc-components/or-mwc-input";
 import i18next from "i18next";
 import {GenericAxiosResponse} from "@openremote/rest";
-import {showErrorDialog} from "@openremote/or-mwc-components/or-mwc-dialog";
+import {showErrorDialog, showOkDialog} from "@openremote/or-mwc-components/or-mwc-dialog";
```
wborn, line 23: This import can also be removed now that the dialog is no longer used.

```suggestion
import {showErrorDialog} from "@openremote/or-mwc-components/or-mwc-dialog";
```

wborn (approved):

Issue #1416, opened 2024-09-13
GatewayTest is unstable


wborn:
This test is unstable, see: https://github.com/openremote/openremote/actions/runs/10845895994/job/30097701804

```
GatewayTest > Verify gateway client service FAILED
    Condition not satisfied after 15.00 seconds and 88 attempts
        at app//spock.util.concurrent.PollingConditions.within(PollingConditions.java:185)
        at app//spock.util.concurrent.PollingConditions.eventually(PollingConditions.java:140)
        at org.***.test.gateway.GatewayTest.Verify gateway client service(GatewayTest.groovy:922)

        Caused by:
        Condition not satisfied:

        attributeEvents.size() == 4
        |               |      |
        |               6      false
[...]

Issue #1409, opened 2024-09-11
TcpClientTest is unstable


wborn:
The test often fails, see: https://github.com/openremote/openremote/actions/runs/10815978529/job/30006111865

```
org.***.test.protocol.TcpClientTest

  Test Check client FAILED (26s)

  Condition not satisfied after 20.00 seconds and 109 attempts
      at app//spock.util.concurrent.PollingConditions.within(PollingConditions.java:185)
      at app//spock.util.concurrent.PollingConditions.eventually(PollingConditions.java:140)
      at org.***.test.protocol.TcpClientTest.Check client(TcpClientTest.groovy:159)
  Caused by: Condition not satisfied:
  
  connectionStatus == ConnectionStatus.CONNECTING
  |                |  |
  CONNECTED        |  class org.***.model.asset.agent.ConnectionStatus
                   false
      at org.***.test.protocol.TcpClientTest.Check client_closure14(TcpClientTest.groovy:161)
      at org.***.test.protocol.TcpClientTest.Check client_closure14(TcpClientTest.groovy)
      at app//groovy.lang.Closure.call(Closure.java:412)
      at app//spock.util.concurrent.PollingConditions.within(PollingConditions.java:168)
      ... 2 more

org.***.test.protocol.TcpClientTest > Check client took: 26026ms

TcpClientTest > Check client FAILED
    Condition not satisfied after 20.00 seconds and 109 attempts
        at app//spock.util.concurrent.PollingConditions.within(PollingConditions.java:185)
        at app//spock.util.concurrent.PollingConditions.eventually(PollingConditions.java:140)
        at org.***.test.protocol.TcpClientTest.Check client(TcpClientTest.groovy:159)

        Caused by:
        Condition not satisfied:

        connectionStatus == ConnectionStatus.CONNECTING
        |                |  |
        CONNECTED        |  class org.***.model.asset.agent.ConnectionStatus
                         false
            at org.***.test.protocol.TcpClientTest.Check client_closure14(TcpClientTest.groovy:161)
            at org.***.test.protocol.TcpClientTest.Check client_closure14(TcpClientTest.groovy)
            at app//groovy.lang.Closure.call(Closure.java:412)
            at app//spock.util.concurrent.PollingConditions.within(PollingConditions.java:168)
            ... 2 more
```

Issue #1388, opened 2024-08-19
Improve metrics collection to provide better empirical data of the system state


DonWillems:
1. In order to judge the progress when working on scaling related improvements
2. To use when evaluating and troubleshooting live deployments

PR #1432, opened 2024-09-30
Major threading improvements, gateway logic updates

richturner:


richturner:
Fixes #1409

test/src/test/groovy/org/openremote/test/energy/EnergyOptimisationAssetTest.groovy
```diff
@@ -128,8 +127,8 @@ class EnergyOptimisationAssetTest extends Specification implements ManagerContai
                     managerTestSetup.electricityBatteryAssetId,
                     ElectricityAsset.POWER_SETPOINT.name,
                     new AssetDatapointIntervalQuery(
-                            LocalDateTime.ofInstant(optimisationTime, ZoneId.systemDefault()).plus((long)(optimiser.intervalSize * 60), ChronoUnit.MINUTES),
-                            LocalDateTime.ofInstant(optimisationTime, ZoneId.systemDefault()).plus(1, ChronoUnit.DAYS).minus((long)(optimiser.intervalSize * 60), ChronoUnit.MINUTES),
+                            optimisationTime.atZone(ZoneId.systemDefault()).plus((long)(optimiser.intervalSize * 60), ChronoUnit.MINUTES).toLocalDateTime(),
+                            optimisationTime.atZone(ZoneId.systemDefault()).plus(1, ChronoUnit.DAYS).minus((long)(optimiser.intervalSize * 60), ChronoUnit.MINUTES).toLocalDateTime(),
```
wborn, line 131: It might be an idea to add an `Instant` constructor to `AssetDatapointIntervalQuery` for ease of use so there will be less code to write/read.

test/src/test/groovy/org/openremote/test/mqtt/MqttBrokerTest.groovy
```diff
@@ -94,8 +94,8 @@ class MqttBrokerTest extends Specification implements ManagerContainerTrait {
 
         then: "the client resources should be freed"
         conditions.eventually {
-            assert clientEventService.sessionKeyInfoMap.isEmpty()
-            assert !clientEventService.sessionKeyInfoMap.containsKey(getConnectionIDString(currentConnection))
+            assert client.getConnectionStatus() == ConnectionStatus.DISCONNECTED
+            assert mqttBrokerService.getUserConnections(keycloakTestSetup.serviceUser.id).size() == 0
```
wborn, line 98:
```suggestion
            assert mqttBrokerService.getUserConnections(keycloakTestSetup.serviceUser.id).isEmpty()
```

test/src/test/groovy/org/openremote/test/protocol/http/HttpClientProtocolTest.groovy
```diff
@@ -235,7 +235,7 @@ class HttpClientProtocolTest extends Specification implements ManagerContainerTr
                     def yearMonthDay = doDynamicTimeReplace('%TIME:yyyy-MM-dd%', getInstantTimeOf(container))
                     def epochMillis = getInstantTimeOf(container).plus(1, ChronoUnit.HOURS).toEpochMilli()
                     def epochSeconds = getInstantTimeOf(container).minus(50, ChronoUnit.DAYS).getEpochSecond()
-                    def localInstant = LocalDateTime.ofInstant(getInstantTimeOf(container), ZoneId.systemDefault())
+                    def localInstant = getInstantTimeOf(container).atZone(ZoneId.systemDefault()).toLocalDateTime()
```
wborn, line 238: The `localInstant` is no longer an `Instant` with this change so it could use a better name.

DonWillems:
I tested the gateway functionality:
1. On test4 I created a gateway from one realm to the other with 10000 test devices. When deleting the gateway, the error dialog came up that it couldn't be deleted, but after a while it was deleted. 1 minute later the whole instance was down.
2. The copy button is still available when selecting a child asset of a gateway, while that action is not allowed.
3. I couldn't double check because I broke the instance. But the connection status on the 'Manager Interconnect' page would not update to 'disconnected' if the gateway was deleted in the realm it tries to connect with.

wborn:
Do you know if this PR also fixes #1416? I retriggered this PR build several times and haven't seen that test fail. It fails continuously on the master branch.

profile/deploy.yml
```diff
@@ -279,12 +279,14 @@ services:
...
-      # the rules engine.
-      OR_SCHEDULED_TASKS_THREADS_MAX: ${OR_SCHEDULED_TASKS_THREADS_MAX:-4}
+      # The number of threads to be used for the scheduled executor service which can be used for scheduling
+      # non-blocking, short-lived tasks in a standard way; defaults to number of CPU cores.
+      OR_SCHEDULED_EXECUTOR_THREADS:
+
+      # The number of threads (min and max) to be used for the executor service which can be used for executing
+      # tasks in a standard way; defaults to number of CPU cores and number of CPU cores x 10 respectively.
+      OR_EXECUTOR_THREADS_MIN:
+      OR_EXECUTOR_THREADS_MAX:
```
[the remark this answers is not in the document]
richturner: The typical virtual servers we run on don't have that many cores available but we can add a max cap for `OR_SCHEDULED_EXECUTOR_THREADS` and `OR_EXECUTOR_THREADS_MIN` also but don't the MAX is only what the system is allowed to do under high loads for short periods of time.

test/load1/run.sh
```diff
@@ -64,6 +66,10 @@ else
 
 cd deployment1
 
+if [ -n "$DEPLOYMENT1_MQTT_RATE_LIMIT" ]; then
+  \$DEPLOYMENT1_ENV = "MQTT_RATE_LIMIT=$DEPLOYMENT1_MQTT_RATE_LIMIT $DEPLOYMENT1_ENV"
```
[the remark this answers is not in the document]
richturner: Becuase the `$` needs to be passed to the shell on the host as it is executed over scp

test/src/test/groovy/org/openremote/test/benchmark/AttributeEventBenchmarkTest.groovy
```diff
@@ -0,0 +1,167 @@
...
+
+        then: "A subscription should exist"
+        assert subscribed
+        assert client.topicConsumerMap.get(topic) != null
+        assert client.topicConsumerMap.get(topic).size() == 1
+        assert mqttBrokerService.getUserConnections(keycloakTestSetup.serviceUser.id).size() == 1
+        def connection = mqttBrokerService.getUserConnections(keycloakTestSetup.serviceUser.id)[0]
+        assert connection != null
+        assert defaultMQTTHandler.sessionSubscriptionConsumers.containsKey(getConnectionIDString(connection))
+        assert defaultMQTTHandler.sessionSubscriptionConsumers.get(getConnectionIDString(connection)).size() == 1
```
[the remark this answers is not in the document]
richturner: You can but when inside polling conditions you **must** include them so for good practice I generally include them in a simple then as well

richturner:
> Do you know if this PR also fixes #1416? I retriggered this PR build several times and haven't seen that test fail. It fails continuously on the master branch.

Yes it likely does as there were concurrency issues in the tests which I have tried to resolve, will reference it

richturner:
> I tested the gateway functionality:
> 
> 1. On test4 I created a gateway from one realm to the other with 10000 test devices. When deleting the gateway, the error dialog came up that it couldn't be deleted, but after a while it was deleted. 1 minute later the whole instance was down.
> 2. The copy button is still available when selecting a child asset of a gateway, while that action is not allowed.
> 3. I couldn't double check because I broke the instance. But the connection status on the 'Manager Interconnect' page would not update to 'disconnected' if the gateway was deleted in the realm it tries to connect with.

@DonWillems As discussed I have created a separate issue for the memory usage on deletion of 10K assets but I have also increased the disk size on this instance so the manager shouldn't fall over again. Maybe you can create two new realms and gateway link those rather than use the master realm with 10K test assets?

wborn:
> will reference it

Adding a new comment does not make GitHub link a PR to close an issue. 

What does work is:

* referencing it by updating your first PR comment
* link it via "Development" in the top right section of your PR (that's what I did)
* add it as comment in a new commit

profile/deploy.yml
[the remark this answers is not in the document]
wborn: Even though it is still wasting threads and memory, it's still an improvement compared to the previous threading situation which was even worse on my 28 available CPU setup :smile: :

## Before

image


## After

image

container/src/main/java/org/openremote/container/Container.java
```diff
@@ -283,7 +293,12 @@ public <T extends ContainerService> T getService(Class<T> type) throws IllegalSt
...
     @Override
-    public ScheduledExecutorService getExecutorService() {
-        return EXECUTOR_SERVICE;
+    public ScheduledExecutorService getScheduledExecutor() {
+        return SCHEDULED_EXECUTOR;
+    }
+
+    @Override
+    public ExecutorService getExecutor() {
+        return EXECUTOR;
```
wborn, line 302: Can you document what these executors are typically used for as you can submit runnables to either of them?
When you do not know the difference, you might end up just using the `SCHEDULED_EXECUTOR` for everything because the `ScheduledExecutorService` class extends `ExecutorService` so it can be used for everything.

richturner: You would think they are the same for the reason you mentioned but they are different beasts, if you try using a `ScheduledExecutorService` for submitting runnables you will quickly discover they don't execute.

I haven't read too much into the logic for having both within java or why the scheduled implementation didn't behave the same for submitted runnables but I have seen both used in other frameworks as well.

wborn: But does it matter as a developer which one you use or can you use whatever executor you want?

E.g. you can also submit a task with a delay of 0 to a `ScheduledExecutorService` to get it executed as fast as possible instead of using an `ExecutorService`.

The issue is that if you use start using the wrong executor, the allocated thread pool might not be big enough and you need to have an idea what tasks run on it to know what would be a reasonable size.

Having to deal with two thread pools may also create unneccesary threads which are sitting idle while the other thread pool is exhausted and could use more threads.

What usually makes sense is to have a different thread pools to make sure certain threads are guaranteed to run without being impacted by other threads.

Without any documentation I would think that these are just generic reusable thread pools and you can use them for whatever you want?

Does some (external?) code use an `ExecutorService` while other code uses an `ScheduledExecutorService` and did these pools come into existence to be able to share the threads in order to bring down the total number of threads?

wborn (changes requested):
With these changes I get errors whenever I try to delete an asset. Looks like the assetId is missing from the request:

Screenshot from 2024-10-23 10-48-13

richturner:
> With these changes I get errors whenever I try to delete an asset. Looks like the assetId is missing from the request:
> 
> Screenshot from 2024-10-23 10-48-13

Good spot, we need to get the UI testing sorted!

Will look into this shortly

wborn:
Does this PR also fix #1388?

container/src/main/java/org/openremote/container/Container.java
richturner: I've now added some javadoc for the executors, key point is the `SCHEDULED_EXECUTOR` is a `ScheduledThreadPoolExecutor` which is a fixed pool implementation in java and this is intended for scheduled task execution only whilst the `EXECUTOR` is a dynamic pool which is for general task execution with a standardised approach to thread management.

richturner:
> > I tested the gateway functionality:
> > 
> > 1. On test4 I created a gateway from one realm to the other with 10000 test devices. When deleting the gateway, the error dialog came up that it couldn't be deleted, but after a while it was deleted. 1 minute later the whole instance was down.
> > 2. The copy button is still available when selecting a child asset of a gateway, while that action is not allowed.
> > 3. I couldn't double check because I broke the instance. But the connection status on the 'Manager Interconnect' page would not update to 'disconnected' if the gateway was deleted in the realm it tries to connect with.
> 
> @DonWillems As discussed I have created a separate issue for the memory usage on deletion of 10K assets but I have also increased the disk size on this instance so the manager shouldn't fall over again. Maybe you can create two new realms and gateway link those rather than use the master realm with 10K test assets?

I have now tested the gateway deletion locally and the different scenarios work as expected.

manager/Dockerfile
```diff
@@ -61,7 +60,7 @@ ENV OR_GATEWAY_TUNNEL_LOCALHOST_REWRITE ${OR_GATEWAY_TUNNEL_LOCALHOST_REWRITE:-1
 ENV OR_JAVA_OPTS ${OR_JAVA_OPTS:--Xms500m -Xmx2g \
     -XX:NativeMemoryTracking=summary \
     -Xlog:all=warning:stdout:uptime,level,tags \
-    -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/dump.hprof}
+    -XX:+HeapDumpOcontainernOutOfMemoryError -XX:HeapDumpPath=/dump.hprof}
```
wborn, line 63: I think this change was not intended.

```suggestion
    -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/dump.hprof}
```

richturner:
> > > > I tested the gateway functionality:
> > > > 
> > > > 1. On test4 I created a gateway from one realm to the other with 10000 test devices. When deleting the gateway, the error dialog came up that it couldn't be deleted, but after a while it was deleted. 1 minute later the whole instance was down.
> > > > 2. The copy button is still available when selecting a child asset of a gateway, while that action is not allowed.
> > > > 3. I couldn't double check because I broke the instance. But the connection status on the 'Manager Interconnect' page would not update to 'disconnected' if the gateway was deleted in the realm it tries to connect with.
> > > 
> > > 
> > > @DonWillems As discussed I have created a separate issue for the memory usage on deletion of 10K assets but I have also increased the disk size on this instance so the manager shouldn't fall over again. Maybe you can create two new realms and gateway link those rather than use the master realm with 10K test assets?
> > 
> > 
> > I have now tested the gateway deletion locally and the different scenarios work as expected.
> 
> Me too, 2 and 3 are still items that can be fixed though. Should we move them to separate issues?

I've fixed 2 and 3 should also be fixed but will confirm by redeploying

wborn (approved):
LGTM now! :+1:

manager/src/main/java/org/openremote/manager/system/HealthService.java
```diff
@@ -119,63 +108,63 @@ public void init(Container container) throws Exception {
...
+//                }
+//
+//                @Override
+//                public Tags getExchangeStatusTags(Route route) {
+//                    return Tags.of(
+//                        ROUTE_ID_TAG, route.getId());
+//                }
+//            });
+//            micrometerRoutePolicyFactory.setMeterRegistry(container.getMeterRegistry());
+//            brokerService.getContext().addRoutePolicyFactory(micrometerRoutePolicyFactory);
```
wborn, line 167: Perhaps we should all the commented code should also have been from the `HealthService`?

PR #1433, opened 2024-10-02
Fix rules triggering on startup

### LINKED ISSUE #1242: Rules: when an environment is started, rules are triggered if the condition is fulfilled
When an environment is started, the rules of which the condition is matched at that moment, will be triggered.

Rules should trigger based on change in attribute value.

wborn:
Fixes #1242

### DIFF
--- a/manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
+++ b/manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
@@ -419,7 +419,6 @@ public class JsonRulesBuilder extends RulesBuilder {
         protected Set<String> otherwiseMatchedAssetIds;
         protected long nextRecur;
         protected boolean matched;
-        protected boolean executeActionsWhenMatched;
         protected Map<String, Long> nextRecurAssetIdMap = new HashMap<>();
 
         public RuleState(JsonRule rule) {
@@ -663,8 +662,6 @@ public class JsonRulesBuilder extends RulesBuilder {
 
         return facts -> {
             ruleState.update(timerService::getCurrentTimeMillis);
-            // Execute the actions initially only when the rules engine has already fired and the rule state is matched
-            ruleState.executeActionsWhenMatched = rulesEngine.hasPreviouslyFired() && ruleState.matched;
             return ruleState.matched;
         };
     }
@@ -678,7 +675,8 @@ public class JsonRulesBuilder extends RulesBuilder {
         return facts -> {
 
             try {
-                if (ruleState.executeActionsWhenMatched) {
+                // Execute actions only when the rules engine has already fired to prevent execution during startup
+                if (rulesEngine.hasPreviouslyFired()) {
                     if (ruleState.thenMatched()) {
                         log(Level.FINEST, "Triggered rule so executing 'then' actions for rule: " + rule.name);
                         executeRuleActions(rule, rule.then, "then", false, facts, ruleState, assetsFacade, usersFacade, notificationsFacade, webhooksFacade, alarmsFacade, predictedDatapointsFacade, scheduledActionConsumer);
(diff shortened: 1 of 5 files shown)


wborn:
WDYT about this approach @richturner?

richturner:
Hey,

Not a big fan of time based mechanism, I would go with a previouslyRun flag in `RuleState` this flag can be set to true in two different places:

1. At the end of `buildLhsCondition` method if `ruleState.matched == false`
2. In finally section of `buildRhsAction` action lambda

This means the flag gets updated after the first execution of the rule whether or not the rule matches.

Then in `buildRhsAction` lambda check this flag before executing the actions but make sure the finally block still runs.

Also I would base your branch off the `rich/optimisation` branch as there are some changes to JsonRulesBuilder in there.

richturner (commented):
See my above comment

wborn:
What do you think about these new changes @richturner?

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -658,6 +663,8 @@ protected Condition buildLhsCondition(JsonRule rule, RuleState ruleState) {
 
         return facts -> {
             ruleState.update(timerService::getCurrentTimeMillis);
+            // Execute the actions initially only when the rules engine has already fired and the rule state is matched
+            ruleState.executeActionsWhenMatched = rulesEngine.hasPreviouslyFired() && ruleState.matched;
```
richturner, line 667: Don't think we need this in the ruleState, just use `rulesEngine.hasPreviouslyFired()` directly in the L681

richturner (changes requested):

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
wborn: Yes that makes sense now that the rules engine keeps track of it. :+1:

manager/src/main/java/org/openremote/manager/rules/RulesEngine.java
```diff
@@ -148,6 +148,7 @@ public RulesEngine(TimerService timerService,
                        MeterRegistry meterRegistry) {
         this.timerService = timerService;
         this.rulesService = rulesService;
+        this.previouslyFired = rulesService.startDone;
```
richturner, line 151: When a rule engine is instantiated then `previouslyFired` should be false. A rule engine can be created after the system has started (e.g. a new realm is added, or the first rule is added to a realm).

Only once rule engine has fired all deployments once should this flag be updated...I don't see anywhere that this flag gets updated anymore.

richturner (changes requested):
This functionality should also be covered in the rules tests somewhere.

manager/src/main/java/org/openremote/manager/rules/RulesEngine.java
wborn: The issue is that it doesn't fire if there are no rules in a realm because then there is no `RulesEngine`. It only creates a `RulesEngine` and starts it after you create a new rule when there are no rules. That was the reason why the tests failed because they added rules to a running instance in a realm that did not have any rules. It can have a better name though. :wink:

wborn: How about also naming it `startDone`?

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -582,6 +584,8 @@ public JsonRulesBuilder(Logger logger, Ruleset ruleset, TimerService timerServic
         }
     }
 
+
+
```
richturner, line 588: ?

richturner (changes requested):

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
wborn, line 588:
```suggestion
```

richturner (approved):

PR #1436, opened 2024-10-04
Support for multiple languages in push- and email notifications

### LINKED ISSUE #1401: Notifications: Add option to set multiple languages for the notifications
As a user I want to set multiple languages for notifications I sent using When-Then rules, so that end users get the notification in their preferred language.

User attribute:
- Store the preferred language of a user against a user attribute. This language comes from the phone browser language, and later from the user preferred language as selected in the app.
- Should we also store it in an attribute of the console? If a user account is used by multiple people, they could still have their own language set in this case. When sending a notification the system would first check if there is a language set in the console, if not, use the language set on the user account.

Manager_config.json and appearance page:
- Set the languages that should be available for notifications localization.
- Set the default language if the user's language is not present.
- Multi-select dropdown that shows selected languages
- Select dropdown to set a language from the previously selected languages as default. (if the language that is currently default is removed from the list above, set the first language in the list as new default)
Image

Manager UI, When-Then rule:
- Add a select field in the dialog to switch between languages if more than one is set in the manager config OR if the particular JSON rule already has multiple languages (so even if its not in the manager config, it should show the dropdown for that rule). All fields can be different per language. The default language is shown when the dialog is opened.
- When switching to a different language, all fields except title and content are already filled with the contents of the default page. All can be overwritten.
- When a languages isn't fully set, indicate with a red number icon next to the language field. On hover (in the title of the icon) you can see which languages are not set.
- Or if they are not set at all, does it just fall back to the default? Should it be a warning, not an error? Everything that is not filled in just uses default?
Image

JSON rule and notification handler:
- Should add language to their structure for email and push notifications
- Should be backwards compatible, if there is not language in the JSON, it should just use default
- This would be implemented in the notification message and the notification handlers will need to understand it and deal with language preference lookup.

Open questions:
- Should the manager also keep track of preferred language of user?
- Should the user page and account page show preferred language?

MartinaeyNL:
Fixes #1401 

#### Models:
- Added new `LocalizedNotificationMessage` model, that can be used with `type: localized`.
It uses a `languages` field, of type `Map<String, AbstractNotificationMessage>`,
to configure a different message for each language string:
  - The key should be a two lettered string like `nl`, `en` or `de`. 
  - The value can be **any** type of notification; so both `email` and `push` are supported.
- Added `ManagerAppRealmNotificationConfig` to configure the languages for notifications in the `manager_config.json` file.
- Added `locale` field to the `Notificaton.Target` model, to override the locale for sending a notification.
This bypasses the check with the identity provider to check the preffered language of this target.
- Added `allowedLocales` field to the `Notification.Target` model, to cache the list of 'allowed messages' for this user.

#### Notification handling:
- Added new `LocalizedNotificationHandler` class. Not much new logic here, as it utilizes existing logic from `PushNotificationHandler` and `EmailNotificationHandler`. For example, `sendMessage()` checks the users' language **before** sending a notification, and dispatches the correct message for their language (using the correct handler).

#### HTTP API:
- Added /user/locale endpoint for updating the Keycloak user locale attribute.

#### UI - Core / Manager app:
- Improved `or-core` and manager UI code to gather user locale from the Keycloak access token,
and prioritized it along with the other configured languages. (from consoles, realm configuration, etc.)
- Added mechanic to the Manager app, so the list of languages in `manager_config.json` are inserted into the `RulesConfig`.
- Added `email_localized` and `push_localized` entries in the `RulesConfig`.

#### UI - Appearance page
- Each realm now has an option to configure the list of allowed notification languages, and select a default language.
- Added a language search field using the `ISO6391` library / specification.

#### UI - Rules page:
- Added a custom `or-rule-form-localized` component, with a dropdown to navigate through the list of languages.
Based on the message type (`email` or `push`) a different form is shown. Changes are saved for each language individually.
- Minor changes to the `or-rule-form-push-notification` component.

#### Other:
- Minor fix for alarm status button being fetched twice.


MartinaeyNL:
Updated the changelog.

Currently busy with writing the tests for the localized notifications.
The frontend code (Rules UI, manager_config, and OR-Core), and the Manager code (API, models, handler) should be complete.

Probably will do some adjustments based on the code review of @richturner, and the UI check with @DonWillems.
Deploying on test1 as we speak.

richturner:
Hi @MartinaeyNL,

I don't think you want a whole new `NotificationHandler`, they are intended for handling different message types for localised notifications I would expect support to be added to `AbstractNotificationMessage` and then code re-used/shared between the existing NotificationHandler implementations to lookup localised messages.

manager/src/main/java/org/openremote/manager/security/UserResourceImpl.java
```diff
@@ -336,6 +336,22 @@ public void updateClientRoles(RequestParams requestParams, String realm, Role[]
         }
     }
 
+    @Override
+    public void updateCurrentUserLocale(RequestParams requestParams, JsonNode locale) {
```
richturner, line 340: What if admins want to alter user locales?

model/src/main/java/org/openremote/model/security/UserResource.java
```diff
@@ -182,6 +183,12 @@ public interface UserResource {
     @Operation(operationId = "updateUserRealmRoles", summary = "Update realm roles for a user in a realm")
     void updateUserRealmRoles(@BeanParam RequestParams requestParams, @PathParam("realm") String realm, @PathParam("userId") String userId, Role[] roles);
 
+    @PUT
+    @Path("locale")
+    @Consumes(APPLICATION_JSON)
+    @Operation(operationId = "updateCurrentUserLocale", summary = "Update locale for the current user in a realm")
+    void updateCurrentUserLocale(@BeanParam RequestParams requestParams, JsonNode locale);
```
richturner, line 190: Why is `locale` a `JsonNode` rather than a `String`?

richturner (changes requested):

manager/src/main/java/org/openremote/manager/notification/LocalizedNotificationHandler.java
richturner, line 1: Please include copyright headers and javadoc comments about what the file is for

MartinaeyNL:
Hi @richturner, thanks for the review!

Yeah that was my first approach when I started on this issue haha
Adding the logic to the existing handlers (email and push) turned out way more complex than I was imagining.
I had to update many places in the codebase (rules, alarms..), making breaking changes to many classes,
and it overall had a large impact.

So I was scouting for a different solution, with less breaking changes, and backwards compatibility with existing deployments.
Which turned out to make a separate `LocalizedNotificationHandler`, that calls functions from the existing handlers.
This would minimalize the amount of extra code, and simplify its use. A nice benefit is the flexibility of using multiple message types for one notification 😉 

To my mistake; I should've contacted you first before changing this approach 😅 

I'll look into the comments you made on moving logic to the `NotificationService`,
and to minimize the effort needed for supporting new notification types 👍

MartinaeyNL:
Will work on the requested changes Rich described.

As an update, here's the list of covered scenarios in tests:
- All push notification- and email tests from previously, with no changes, to ensure backwards compatibility.
- Multi language push notification, different msg in NL&EN -> send to all MASTER users -> only admin receives in English.
  - Add new dutch user to MASTER realm -> Send the same notification -> admin receives English, new user receives Dutch.
  - The user changes from dutch to italian -> Send the same notification (defaultLanguage = "en") -> both get English msg
- Multi language notification: configured so EN gets a push notification, NL gets an email notification -> Both the Dutch, and the English user get their personalized message.

manager/src/main/java/org/openremote/manager/notification/LocalizedNotificationHandler.java
```diff
@@ -0,0 +1,208 @@
...
+
+import static org.openremote.manager.notification.NotificationProcessingException.Reason.*;
+import static org.openremote.model.security.User.LOCALE_ATTRIBUTE;
+
+import java.util.*;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+/**
+ * Class description
```
wborn, line 41: Please add the description.

model/src/main/java/org/openremote/model/manager/ManagerAppRealmConfig.java
```diff
@@ -27,4 +27,5 @@ public class ManagerAppRealmConfig {
     protected String favicon = null;
     protected String language = null;
     protected String[] headers = null;
+    protected ManagerAppRealmNotificationConfig notifications = null;
```
wborn, line 30: Java variables are `null` by default and assigning `null`  to them again is redundant and results in unnecessary CPU instructions.

See: http://www.javapractices.com/topic/TopicAction.do?Id=14

```suggestion
    protected ManagerAppRealmNotificationConfig notifications;
```

model/src/main/java/org/openremote/model/manager/ManagerAppRealmNotificationConfig.java
```diff
@@ -0,0 +1,25 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.openremote.model.manager;
+
+public class ManagerAppRealmNotificationConfig {
+    protected String[] languages = null;
+    protected String defaultLanguage = null;
```
wborn, line 24:
```suggestion
    protected String[] languages;
    protected String defaultLanguage;
```

model/src/main/java/org/openremote/model/notification/Notification.java
```diff
@@ -46,6 +47,8 @@ public static class Target {
 
         protected TargetType type;
         protected String id;
+        protected String locale;
+        protected List<String> allowedLocales = new ArrayList<>();
```
wborn, line 51: If the order is not important it might also make sense to prevent multiple occurences of the same locale:

```suggestion
        protected Set<String> allowedLocales = new HashSet<>();
```

test/src/test/groovy/org/openremote/test/notification/NotificationTest.groovy
```diff
@@ -674,4 +677,323 @@ class NotificationTest extends Specification implements ManagerContainerTrait {
...
+                if (throwPushHandlerException) {
+                    throw new Exception("Failed to send notification")
+                }
+                pushNotificationIds << id
+                pushNotificationTargetTypes << target.type
+                pushNotificationTargetIds << target.id
+                pushNotificationMessages << message
+                callRealMethod()
+        }
+        // Auto accept the messages sent to FIrebase (without calling its logic)
```
wborn, line 720:
```suggestion
        // Auto accept the messages sent to Firebase (without calling its logic)
```

model/src/main/java/org/openremote/model/notification/Notification.java
```diff
@@ -54,6 +57,12 @@ public Target(@JsonProperty("type") TargetType type, @JsonProperty("id") String
             this.id = id;
         }
 
+        public Target(TargetType type, String id, String locale) {
```
wborn, line 60: Can you add some JavaDocs to this class so it is clear that the `locale` contains ISO 639-1 values?

ui/app/manager/src/components/configuration/or-conf-realm/or-conf-realm-card.ts
```diff
@@ -274,6 +277,16 @@ export class OrConfRealmCard extends LitElement {
     render() {
         const colors = this._getColors();
         const app = this;
+        const languageNames = ISO6391.getAllNames();
+        const languages: string[][] = ISO6391.getAllCodes()
+            .map((code, index) => ([code, languageNames[index]]))
+            .sort((a, b) => a[1].localeCompare(b[1]));
```
wborn, line 283: This mapping an sorting looks like expensive string operations to do everytime the component is rendered. Wouldn't it make sense to cache the result and use that?

ui/component/or-rules/src/json-viewer/forms/or-rule-form-localized.ts
```diff
@@ -0,0 +1,175 @@
...
+            <div style="margin-top: 20px; display: flex; justify-content: end; color: ${DefaultColor6};">
+                <or-translate value="languagesInvalidError"></or-translate>
+                <span style="padding-left: 4px;">${invalidLocalesStr}</span>
+            </div>
+        `;
+    }
+
+    /**
+     * A public function that checks if the localized {@link message} is valid.
+     * By looping through the languages, we very for each email- and push notification if some required variables are empty.
```
wborn, line 146:
```suggestion
     * By looping through the languages, we verify for each email- and push notification if some required variables are empty.
```

wborn (changes requested):

MartinaeyNL:
I've addressed the requested changes from both Rich and @wborn. Feel free to verify.
I'll deploy it to a test deployment in a minute.

wborn (dismissed):

wborn (changes requested):
I gave it a test but it now seems no longer possible to localize messages in a realm based on the "default" realm styling when no styling is configured for a realm.

wborn (approved):

wborn (dismissed):

wborn (approved):

PR #1439, opened 2024-10-08
Bugfix solar forecast service

igorrutka:
- Fixed bug where solar forecast service was not started automatically when an asset was created programmatically with the 'isIncludeForecastSolarService' attribute set to 'true'
- Reverted accidental change of attribute  name  'setActualSolarPowerValueWithForecast' back to its original name 'setActualSolarValueWithForecast'


igorrutka:
Good point, I am not familiar with the test. I'll look into it in a few weeks. I think we will need to update this test before we can use it

wborn (approved):

PR #1441, opened 2024-10-10
Add Spotless Gradle plugin configuration and apply it

### LINKED ISSUE #1403: Fix license header issues and improve tooling
There are some issues with the license headers that need to be fixed:

- [ ] There are many files from which the header is missing. It might also be an idea to add it to the TypeScript files.
- [ ] The reference to `CONTRIBUTORS.txt` in the header can be be removed because this information is already part of the Git repository history.
- [ ] The Gradle build should check that the license headers are correct and not missing. This could for instance be done using the [gradle-license-plugin](https://github.com/jaredsburrows/gradle-license-plugin).

This also applies to other repostories like console-android, console-ios.

wborn:
As a test for improving the code quality I added a [Spotless](https://github.com/diffplug/spotless/) Gradle plug-in configuration.

You can apply the changes using:

`gradle spotlessApply`

It can also check code if it complies using:

`gradle spotlessCheck`

which can be used as a [pre-push hook](https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#git-hook) and during CI builds.

This plugin can help with updating and checking if license headers are in place (https://github.com/openremote/openremote/issues/1403).

We can also add a configuration to the plugin so it can check/format using Google/Eclipse/Prettier code formatters.

After reformatting the code we can add a `.git-blame-ignore-revs` file to ignore this from git blame, see:

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

TODO:

- [x] Agree upon coding standards and formatters being used
- [x] Check if formatting rules work well with IDEs
  * Google Java Format: [IntelliJ IDEA plug-in](https://plugins.jetbrains.com/plugin/8527-google-java-format), [VS Code plug-in](https://marketplace.visualstudio.com/items?itemName=JoseVSeb.google-java-format-for-vs-code)
  * Spotless: [IntelliJ IDEA plug-in](https://plugins.jetbrains.com/plugin/18321-spotless-gradle), [VS Code plug-in](https://marketplace.visualstudio.com/items?itemName=richardwillis.vscode-spotless-gradle)
- [x] Adapt GHA workflow so it runs the `spotlessCheck` task in PR builds
- [x] Fix or remove commented code that does not look nice after formatting
- [x] Fix or exclude JavaDocs from formatting that does not look nice after formatting
- [x] Add/update documentation for this
  - [x] https://github.com/openremote/documentation/pull/183
- [x] Apply the same to other projects
  - [x] https://github.com/openremote/custom-project/pull/12
  - [x] https://github.com/openremote/extensions/pull/55


tools/spotless/openremote-license-header.txt
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright $YEAR, OpenRemote Inc.
```
wborn, line 2: Many files are missing the header and it is quite some work to figure out the exact year files got added. We could also change it into a range so it is always correct. The Git history keeps track of all the changes anyhow.

```suggestion
 * Copyright 2015-$YEAR, OpenRemote Inc.
```

tools/spotless/openremote-license-header.txt
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * See the CONTRIBUTORS.txt file in the distribution for a
+ * full listing of individual contributors.
+ *
```
wborn, line 6: The Git history keeps track of all contributors so we could remove the file and reference to it:

```suggestion
```

tools/spotless/openremote-license-header.txt
```diff
@@ -0,0 +1,19 @@
...
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
```
wborn, line 19: Adding a SPDX license identfier to the header also helps other software with identifying software licenses (ISO/IEC 5962:2021).

See: https://spdx.org/licenses/

```suggestion
  *
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
```

tools/spotless/openremote-license-header.txt
richturner: I think a fixed year of last edit is fine, a range becomes more problematic unless you are trying to show the company origin year which is not what we need to convey here.

wborn:
I will just drop the last commit, rebase the PR and apply Spotless once in a while until this gets merged. :wink:

wborn:
WDYT about the TODO list in the first comment @richturner?

richturner:
> WDYT about the TODO list in the [first comment](https://github.com/openremote/openremote/pull/1441#issue-2578124837) @richturner?

List looks good, I see google gave up on XML coding standard DSL for IntelliJ and Eclipse and now have a standalone formatting tool:

https://github.com/google/styleguide/issues/687

I think real time formatting in the IDE is helpful but not against a cleanup script applying additional styling that might be missed with IDE.

Google also have an IntelliJ plugin for handling styling (which I believe offloads to their standalone tool but not certain).

richturner:
Don't know if there are advantages to tsfmt vs eslint with typescript plugin?

richturner:
`google-java-format` has plugins for IntelliJ, Eclipse, VSCode and Spotless so seems like a good option.

We can have realtime formatting applied in the IDE and precommit and/or CI/CD actions to enforce.

richturner (approved):

PR #1451, opened 2024-10-21
Skip the mTLS test when test.mosquitto.org fails the signing request (with HTTP 502).

pankalog:
To skip a test programmatically, make sure to throw `TestAbortedException` on the top level (https://github.com/spockframework/spock/issues/1185). The throw needs to happen in the top level of the test, couldn't get it to work using a Util function I made.


richturner:
I cannot see how this change stops the test from failing if mosquitto server is not available/malfunctioning

richturner (approved):

PR #1453, opened 2024-10-21
Fixes JSON rules notification with linked user realm query in the action target (Fixes #1447)

### LINKED ISSUE #1447: Rules: notifications are sent to a user even if that user does not have a realm role matching the targeted role
**Reproduce:**
1. Create user with realm role, e.g. "installer"
2. Create When-Then rule that sends a push notification to a linked user with a realm role other than "installer"
3. Trigger the rule
4. The user gets a push notification

**Expected:**
4. The user does not get a notification, because he is not of the correct role.

richturner:


manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -785,7 +785,7 @@ protected RuleActionExecution buildRuleActionExecution(JsonRule rule, RuleAction
 
             Notification notification = ValueUtil.clone(notificationAction.notification);
             String body;
-            boolean linkedUsersTarget = ruleAction.target != null && ruleAction.target.linkedUsers != null && ruleAction.target.linkedUsers;
+            boolean linkedUsersTarget = ruleAction.target != null && Objects.equals(Boolean.TRUE, ruleAction.target.linkedUsers);
```
wborn, line 788:
```suggestion
            boolean linkedUsersTarget = ruleAction.target != null && Boolean.TRUE.equals(ruleAction.target.linkedUsers);
```

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -807,7 +807,7 @@ protected RuleActionExecution buildRuleActionExecution(JsonRule rule, RuleAction
             // Transfer the rule action target into notification targets
             Notification.TargetType targetType = Notification.TargetType.ASSET;
             if (ruleAction.target != null) {
-                if (ruleAction.target.users != null
+                if ((ruleAction.target.users != null || Objects.equals(Boolean.TRUE, ruleAction.target.linkedUsers))
```
wborn, line 810:
```suggestion
                if ((ruleAction.target.users != null || Boolean.TRUE.equals(ruleAction.target.linkedUsers))
```

wborn (commented):
As the constant is never `null` you can use a simple equals here:

richturner:
Thanks for the review @wborn suggestions incorporated and branch has been updated to latest master

wborn (approved):

PR #1456, opened 2024-10-22
Fix rules triggering on startup

wborn:
The fix in #1433 did not work when the engine has multiple rules because `previouslyFired` was set after firing the first rule.


DonWillems (approved):

richturner:
Apologies I should have picked up that obvious error in my previous review.

PR #1492, opened 2024-11-14
Gateways: Add option to close tunnels automatically after X minutes

### LINKED ISSUE #1443: Gateways: Add option to close tunnels automatically after X minutes
As a user I want gateway tunnels to close after a set period, so that the connection does not stay open if someone forgets to close it after use.

- Add environment variable in which a system admin can set after how many minutes a created tunnel should be closed
- Automatically close a tunnel after the amount of minutes set in the variables
- If no value is set, tunnels stay open indefinitely
- Add a timer in the gateway tunnel and Insight tunnel widgets in the manager that shows how much time is left on a tunnel

wborn:
Closes #1443

### DIFF
--- a/manager/src/main/java/org/openremote/manager/gateway/GatewayService.java
+++ b/manager/src/main/java/org/openremote/manager/gateway/GatewayService.java
@@ -56,6 +56,7 @@ import org.openremote.model.syslog.SyslogCategory;
 import org.openremote.model.util.TextUtil;
 import org.openremote.model.value.MetaItemType;
 
+import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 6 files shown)


manager/src/main/java/org/openremote/manager/gateway/GatewayService.java
```diff
@@ -600,11 +613,11 @@ public void stopTunnel(GatewayTunnelInfo tunnelInfo) throws IllegalArgumentExcep
         try {
             stopFuture.get(20, TimeUnit.SECONDS);
         } catch (ExecutionException e) {
-            String msg = "Failed to start tunnel: An error occurred whilst waiting for the tunnel to be started: id=" + gatewayId;
+            String msg = "Failed to stop tunnel: An error occurred whilst waiting for the tunnel to be stopped: id=" + gatewayId;
```
richturner, line 616: Good spot 😃

manager/src/main/java/org/openremote/manager/gateway/GatewayService.java
```diff
@@ -600,11 +613,11 @@ public void stopTunnel(GatewayTunnelInfo tunnelInfo) throws IllegalArgumentExcep
...
             LOG.log(Level.WARNING, msg, e.getCause());
             throw new RuntimeException(msg, e.getCause());
         } catch (InterruptedException | TimeoutException e) {
-            String msg = "Failed to start tunnel: An error occurred whilst waiting for the tunnel to be started: id=" + gatewayId;
+            String msg = "Failed to stop tunnel: An error occurred whilst waiting for the tunnel to be stopped: id=" + gatewayId;
```
richturner, line 620: Good spot 😃

richturner (commented):
Frontend should probably auto refresh whenever a tunnel expires, a `setTimeout` that is cleared onDisconnect of the web component (`gateway-widget` and `page-gateway-tunnel`)

manager/src/main/java/org/openremote/manager/gateway/GatewayService.java
wborn: For sure it will prevent some confusion while debugging! :laughing:

richturner (approved):

PR #1494, opened 2024-11-14
feat: custom tile server and tile uploads

Ekhorn:


manager/src/main/java/org/openremote/manager/map/MapResourceImpl.java
```diff
@@ -71,4 +86,46 @@ public byte[] getTile(int zoom, int column, int row) {
...
+
+            int statusCode = connection.getResponseCode();
+            response.setStatus(statusCode);
+
+            String contentType = connection.getContentType();
+            response.setContentType(contentType);
+
+            String encoding = connection.getContentEncoding();
+            InputStream inputStream;
+            // TODO: GZIP not applied correctly
```
Ekhorn, line 112: Compression of tiles is removed in the proxy endpoint

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -443,6 +463,46 @@ public byte[] getMapTile(int zoom, int column, int row) {
...
+        // String path = exchange.getRequestPath().substring("/api/ma".length());
+        // exchange.setRequestURI(path, true);
+        // exchange.setRequestPath(path);
+        // exchange.setRelativePath(path);
+
+        // try {
+        //     proxyHandler.handleRequest(exchange);
+        // } catch (Exception exception) {
+        //     LOG.log(Level.WARNING, "Error trying to proxy request to tile server", exception);
+        // }
```
Ekhorn, line 503: Remove

model/build.gradle
```diff
@@ -27,6 +29,8 @@ dependencies {
 
     api "io.swagger.core.v3:swagger-annotations:$swaggerVersion"
 
+    // api "io.undertow:undertow-core:$undertowVersion"
+
```
Ekhorn, line 33: Remove

model/src/main/java/org/openremote/model/map/MapResource.java
```diff
@@ -19,13 +19,18 @@
  */
 package org.openremote.model.map;
 
+// import io.undertow.server.HttpServerExchange;
```
Ekhorn, line 22: Remove

model/build.gradle
```diff
@@ -15,7 +15,9 @@ dependencies {
 
     api "jakarta.annotation:jakarta.annotation-api:$javaxAnnotationsVersion"
     api "jakarta.ws.rs:jakarta.ws.rs-api:$jaxrsSpecVersion"
-
+    api ("org.jboss.resteasy:resteasy-undertow:$resteasyVersion") {
+        exclude group: "io.undertow",  module: "undertow-servlet-jakarta"
+    }
```
Ekhorn, line 20: can be removed?

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -79,16 +79,20 @@ public class MapService implements ContainerService {
...
         this.mapSettings.clear();
         ObjectNode mapSettingsJson = loadMapSettingsJson(mapSettingsPath);
         if(mapSettingsJson == null) {
             mapSettingsJson = ValueUtil.JSON.createObjectNode();
         }
         try(OutputStream out = new FileOutputStream(new File(mapSettingsPath.toUri()))){
-            mapSettingsJson.putPOJO("options", mapConfiguration);
+            mapSettingsJson.putPOJO("options", mapConfiguration.options);
+            mapSettingsJson.putPOJO("sources", mapConfiguration.sources);
+            // TODO: validate url is not the same as ext
```
Ekhorn, line 94: Need to prevent the proxy endpoint calling itself

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -79,16 +79,20 @@ public class MapService implements ContainerService {
...
+            // TODO: validate url contains /{z}/{x}/{y}
```
Ekhorn, line 95: Need to ensure this always included if the string is not empty.

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -442,6 +461,36 @@ public byte[] getMapTile(int zoom, int column, int row) {
...
+
+        if (!source.contains("{secret}")) {
+            return null;
+        }
+
+        final String secret = settings
+            .get("sources")
+            .get("external")
+            .get("secret")
+            .textValue();
```
Ekhorn, line 483: might want to retrieve from an encrypted file instead.

Ekhorn: The tile server tokens are usually public, so this can be ignored.

image

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -299,33 +390,52 @@ export class PageConfiguration extends Page<AppStateKeyed> {
 
     // FETCH METHODS
 
+    protected async uploadCustomMap(e: CustomEvent) {
+        const file = e.detail.value[0] as File;
+        this.tilesForUpload = file;
+        this.requestUpdate()
```
MartinaeyNL, line 396: You can add the `@state` decorator to the `tilesForUpload: File` field, so you don't have to manually request a state update anymore.

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -274,6 +327,44 @@ export class PageConfiguration extends Page<AppStateKeyed> {
                         </or-panel>
                         <or-panel .heading="${i18next.t("configuration.mapSettings").toUpperCase()}">
                             ${when(this.mapConfig, () => html`
+                                <div class="global-settings-container">
+                                    <div class="custom-server-group">
+                                        <div class="subheader">${i18next.t("configuration.global.tileServer")}</div>
+                                        <span>${i18next.t("configuration.global.tileServerDescription")}</span>
```
MartinaeyNL, line 333: We have a custom `<or-translate>` component that we use for translations like these.
It does a lot of things, but it also makes sure the text corrects itself when the user changes their language.
So you can replace all the `i18next.t("key")` appearances to `<or-translate value="key"></or-translate>`

ui/app/shared/locales/en/or.json
MartinaeyNL, line 1: Could you also add the Dutch translations for these? We try to support at least English and Dutch, other ones are optional.

MartinaeyNL (changes requested):
I've added some comments, requesting minor adjustments for the UI. I'll leave the architecture/backend logic to @wborn and @richturner

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -72,13 +81,28 @@ public class MapService implements ContainerService {
...
     protected ConcurrentMap<String, ObjectNode> mapSettingsJs = new ConcurrentHashMap<>();
 
-    public ObjectNode saveMapConfig(Map<String, MapRealmConfig> mapConfiguration) throws RuntimeException {
+    public ObjectNode saveMapConfig(MapConfig mapConfiguration) throws RuntimeException {
         if (mapConfig == null) {
             mapConfig = ValueUtil.JSON.createObjectNode();
         }
-        mapConfig.putPOJO("options", mapConfiguration);
+
+        ObjectNode vector_tiles = ValueUtil.JSON.valueToTree(mapConfiguration.sources.get("vector_tiles"));
```
wborn, line 89:
```suggestion
        ObjectNode vectorTiles = ValueUtil.JSON.valueToTree(mapConfiguration.sources.get("vector_tiles"));
```

model/src/main/java/org/openremote/model/manager/MapSourceConfig.java
```diff
@@ -0,0 +1,82 @@
...
+	 */
+	protected Integer maxzoom;
+	/**
+	 * Contains an attribution to be displayed when the map is shown to a user.
+	 * 
+	 * @implSpec Optional string.
+	 */
+	protected String attribution;
+	// protected String promoteId;
+	// protected boolean volatile;
```
wborn, line 81: Commented code:

```suggestion
```

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -372,6 +413,49 @@ public byte[] getMapTile(int zoom, int column, int row) {
         }
     }
 
+    public boolean saveUploadedFile(InputStream fileInputStream) {
+        // TODO: Specify target directory for uploaded file
```
wborn, line 417: Is there still something to be done here?

model/src/main/java/org/openremote/model/manager/MapSourceConfig.java
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright 2022, OpenRemote Inc.
```
wborn, line 2:
```suggestion
 * Copyright 2025, OpenRemote Inc.
```

wborn (changes requested):

wborn (changes requested):
Can you also update this branch to the latest code base?

manager/src/main/java/org/openremote/manager/map/MapService.java
Ekhorn: The line below states the following path for the custom map tiles to be saved at: 

```java
Path destinationPath = Paths.get("manager/src/map/", "mapdata-custom.mbtiles");
```

I think @Spider-Frog doesn't know where to save the file. Do you have any preferences?

wborn: It should store the file on a Docker volume when running OpenRemote in production.
The `manager/src/map` path is only used when running OpenRemote during development in your IDE.

So it would probably be better to add a `public Path getCustomMapTilesPath()` method in the `ConfigurationService` and use that to determine the location. E.g. similar to how the `mapdata.mbtiles` path is determined, see:

https://github.com/openremote/openremote/blob/2a077391a0cd4a47aa1ba1d25e2a3f06477838c5/manager/src/main/java/org/openremote/manager/app/ConfigurationService.java#L93-L97

MartinaeyNL:
Pulled the branch locally to test the frontend of the map tile upload.
The code itself looks OK, but it needs more testing of the "bad flow" to make this more robuust.
Scenario's such as;
- Uploading a file that is not of `.mbtiles` format still succeeds.
- Deletion of a maptiles file seems to throw errors without handling them in the UI.
etc.

MartinaeyNL (commented):
Pulled the branch locally to test the frontend of the map tile upload.
The code itself looks OK, but it needs more testing of the "bad flow" to make this more robuust.
Scenario's such as;
- Uploading a file that is not of `.mbtiles` format still succeeds.
- Deletion of a maptiles file seems to throw errors without handling them in the UI.
etc.

@wborn should take a look at the architecture-related changes once again.

Ekhorn:
Uploading tiles should now be:

1. validated on the server
2. default to .mbtiles in file picker
3. limited by default to 1GB

Issue #1464, opened 2024-10-25
Rules: Notifications: Add template for AssetID in URL field


DonWillems:
As a user I want to send notifications that navigate to an asset page when opened, so that my end-users immediately go to the mentioned asset.

- Check how %TRIGGER_ASSETS% and %RULESET_NAME% works
- Add %ASSET_ID% as template that places the asset id of the asset that triggered this rule. If there are multiple, use the asset id of the first asset listed on the LHS (possible?)
- Add %ASSET_IDS% as a template that places all asset ids that caused this rule to trigger comma separated
- Add these templates to the documentation

![Image](https://github.com/user-attachments/assets/9537a6d7-1819-4c0d-acdb-b71f2b5adb13)

Consider:
Adding more templates such as value, asset name, attribute name

mihaelaaleks:
Changed solution to handle: 

- `Asset ID` template in URL: currently replaces the first asset that triggers the rule (logically the use case for the URL is linking to a single asset). 
- `Asset ID` in body: can chain Asset IDs together. 

I've tested the following scenarios (successfully): 

1. Multiple `when` conditions
2. Child/parent asset trigger for correct Asset ID

However, since I can't locally send notifications, some more extensive testing is needed.

MartinaeyNL:
I've added my code review to the linked PR. Feel free to take a look.

PR #1497, opened 2024-11-15
Add assetid templates

mihaelaaleks:
Closes  [#1464](https://github.com/openremote/openremote/issues/1464)


manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -20,6 +20,8 @@
 package org.openremote.manager.rules;
 
 import jakarta.ws.rs.core.MediaType;
+import net.sf.saxon.event.Event.Text;
```
MartinaeyNL, line 23: Unused import

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -800,8 +804,21 @@ protected RuleActionExecution buildRuleActionExecution(JsonRule rule, RuleAction
                 if (isPush) {
                     PushNotificationMessage pushNotificationMessage = (PushNotificationMessage) notification.getMessage();
                     body = pushNotificationMessage.getBody();
+                    action = pushNotificationMessage.getAction();
+
+                    if (action != null && action.getUrl() != null) {
+                        String newUrl = replaceAssetIdPlaceholder(action.getUrl(), ruleState, useUnmatched, "notification URL", true);
+                        action.setUrl(newUrl);
+                    }
```
MartinaeyNL, line 812: Add `pushNotificationMessage.setAction()` to apply these changes. 😉 
Similar as you did with the body below.

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -785,6 +788,7 @@ protected RuleActionExecution buildRuleActionExecution(JsonRule rule, RuleAction
 
             Notification notification = ValueUtil.clone(notificationAction.notification);
             String body;
+            PushNotificationAction action;
```
MartinaeyNL, line 791: I think you can move this variable declaration down, to a spot after `if (isPush) { }`,
as this variable isn't use elsewhere.

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
MartinaeyNL, line 1: You'll need to specifically add support for localized messages. Sorry for that.

I'd add this using a `if (isLocalized) { }` check above line 798.
From there you loop through all the messages for each language using
```java
LocalizedNotificationMessage localizedMsg = (LocalizedNotificationMessage) notification.getMessage();
localizedMsg.getMessages().forEach((lang, msg) -> {
    if (msg instanceof PushNotificationMessage pushMsg) {
        pushMsg.setBody(replaceAssetIdPlaceholder( ... ));
        // ... and the same for the URL
    }
}
```

MartinaeyNL:
Additionally I'd like to add;
We should write Groovy tests for new functionality like this, to verify it's all working accordingly.
Think of adding tests for cases like these;
- Does the `%ASSET_ID%` placeholder gets replaced by the triggered asset
- If the rule triggers for 2 linked users, do they individually get their personal asset ID in the message?
- If a user should get a notification, but does not have access to the asset, is the asset ID still included in the message?

MartinaeyNL:
Let me know @mihaelaaleks if you need help figuring this out.

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
mihaelaaleks: I've added this change and reworked one of the localized notifications tests to also tests for the placeholder.

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -791,15 +793,47 @@ protected RuleActionExecution buildRuleActionExecution(JsonRule rule, RuleAction
...
+                            String newUrl = replaceAssetIdPlaceholder(action.getUrl(), ruleState, useUnmatched, "notification URL", true);
+                            action.setUrl(newUrl);
+                            pushMsg.setAction(action);
+                        }
+                        
+                        if (pushMsg.getBody() != null) {
+                            String newBody = replaceAssetIdPlaceholder(pushMsg.getBody(), ruleState, useUnmatched, "notification body", false);
+                            pushMsg.setBody(newBody);
+                        }
+                    }
```
wborn, line 812: AFAIK it's also possible that the `LocalizedNotificationMessage` can contain an `EmailNotificationMessage` instead of a `PushNotificationMessage`? Would be nice if support for this is also added in the future.

wborn (approved):

Issue #1462, opened 2024-10-25
Agents: 'agent disabled' checkbox does not seem to disable the agent


DonWillems:
**Reproduce:**
1. Create MQTTAgent
2. Configure it to connect to a broker
3. The agent connects
4. Click 'agent disabled'
5. It stays connected, nothing changes

**Expected:**
5. State changes to Disconnected and in the broker you see its not connected anymore

PR #1507, opened 2024-11-25
Use ThingsAsset in Agent to check for configurationAttributes.

Miggets7:
#1462


richturner (dismissed):

model/src/main/java/org/openremote/model/asset/agent/Agent.java
```diff
@@ -272,15 +273,14 @@ public T setUpdateOnWrite(boolean updateOnWrite) {
...
      */
     public boolean isConfigurationAttribute(String attributeName) {
-
         // This is an event for an agent so is it for an attribute that has a descriptor which is defined in an agent class
         // and it's not the status attribute (or we'll end up in a loop)
         return !attributeName.equals(Agent.STATUS.getName())
             && ValueUtil.getAssetInfo(getType())
             .map(info -> info.getAttributeDescriptors().containsKey(attributeName))
             .orElse(false)
             // Exclude attributes that have a descriptor from the base Asset class
```
wborn, line 282:
```suggestion
            // Exclude attributes that have a descriptor from the base ThingAsset class
```

richturner (approved):

wborn (approved):

Issue #1459, opened 2024-10-23
Assets: when creating the first asset in a new realm, it does not appear in the list nor is selected and shown


DonWillems:
Reproduce:
1. Create new realm
2. Add asset
3. See an empty list and an empty page

Expected:
3. See the asset in the list, selected, and shown on the right side.

Issue #1458, opened 2024-10-23
Asset cannot not be saved because UI thinks it was modified


wborn:
**Describe the bug**
Often I cannot save an Asset because the UI thinks it was modified while it was not.

It then show the message **"The asset was modified by another source and the page needs to refresh. You will lose your changes."** and I have to apply my changes again and save it.

This often occurs after creating the very first Asset in a new OR instance and then modifying it.

**To Reproduce**
1. Create a new OR instance
1. Go to 'Assets'
2. Create an Asset
3. Edit the Asset
4. Save the Asset

**Expected behavior**
It should always be possible to save an Asset when it has not been modified in the meantime.

**Screencapture**
It can alo be reproduced by reloading the browser as you can see in this screencapture:

![Recording 2024-10-23 at 13 40 55](https://github.com/user-attachments/assets/5a323837-92e7-4d55-abbc-c18404b64a8a)

Perhaps sometimes the UI receives a WebSocket message after the page has been fully loaded which then result in marking it as modified?

wborn:
I just gave #1509 a test and the issue no longer occurs. :+1:

Issue #556, opened 2021-12-09
Creating/duplicating asset not showing


pierrekil:
When creating or duplicating asset, they are not shown in the asset tree. It requires refresh to see them.

Expected: when added and saved, the created asset should be selected and opened in the asset view.

richturner:
I have also seen this but I am struggling to reproduce it consistently will keep trying.

DonWillems:
I can produce one instance of it quite consistently. I'm using the manufacturer demo setup. There are two parents with children.

1. Go in modify mode
2. Select the first parent (keep both collapsed)
3. Refresh the page
4. Select the second parent (keep both collapsed)
5. Create Thing asset
6. Save the Thing asset
7. It now shows 'Please select an asset on the left' and the following message in the console:

![image](https://github.com/user-attachments/assets/c37ed140-b20a-427b-8fab-4e456407f9f3)

DonWillems:
When following the reproduction path in my comment above, I found that step 7 does not occur anymore. What does happen is the following:
![addAssetToNewParent](https://github.com/user-attachments/assets/5ceb1425-ee5e-4a27-b18b-1984f53f5ceb)
6. Save the Thing asset
7 new. The save button does not become disabled, the saved Thing asset does not become selected.
8 new. Save again. Save button is still enabled and can be pressed again. The assets are being saved.
9 new. Try to navigate to a different asset, you get the discard changes pop up

PR #1509, opened 2024-11-26
UI update: Bugfixes for asset viewer and asset tree

richturner:


DonWillems:
While testing I found that when adding a new asset in view mode, the tree does not (always) expand anymore and the new asset is not selected. Because of this, it seems like nothing has happened.
AddToViewMode

This issue still occurs for me: https://github.com/openremote/openremote/issues/1458
Wouter reported is was solved, but on the test deployment (and with mobile internet) I could reproduce it consistently.

New issue I can consistently reproduce:
1. Go into modify mode
2. Change the name of an asset
3. Save (successfully)
4. Switch to a different asset by clicking it in the tree
5. A 'Discard changes' dialog pops up.
6. The changes were saved correctly, discard or cancel has no effect.

This one is fixed for me: https://github.com/openremote/openremote/issues/1459

richturner:
REST API has a default timeout of 10s I can only reproduce these issues if I throttle my connection enough to exceed this timeout so what we can do is increase the timeout but also ensure error handling is done correctly.

richturner:
Some of the problem(s) described above also suggest packet loss not just a slow connection. The app uses websockets which is an active connection, if there is packet loss related to this it can go undetected by the UI.

DonWillems:
> 1. New asset not selected
> 2. When saving changes to an asset for the first time in modify, dialog appears
> 3. Discard changes dialog

@richturner 2 and 3 are fixed for me now. 1 still occurs, although I don't see the endless loading I described in 1.(4) anymore.
Here is a gif of what happens about ~20% of the time while testing:
selectNewAsset

I can't find a clear reproducable steps except add a few assets and you will see it occur. It is not related to an asset already having children or not, it can appear in both.

DonWillems:
I tested #556 because I thought it might be solved by the changes you made here. The error described there does not occur anymore, but another issue does come up that might be interesting to look into regarding this PR (I've added info to #556 ). If you want to keep it as a separate issue, that is fine too.

richturner:
Should now be fixed and also fixed an issue where no discard warning was shown when add button was clicked

DonWillems:
#556 and the issues described are tested and fixed. I found two additional issues:
**Can't use browser back/forward navigation between assets.**
1. Select asset A
2. Select asset B
3. Use browser back button to go back to A
4. See URL change, but asset viewer does not show A.
5. The same goes for the forward button to go to B.

**Asset tree collapses _sometimes_ in _some_ places when adding a new asset**
This seems a likely path to reproduction:
1. Expand Asset X
2. Expand Asset Y that is on the same level as X
3. Add an asset to Y
4. X collapses

I would say this second one isn't that big of a deal, so we could save that as a separate item.

DonWillems (approved):

richturner:
I couldn't reproduce the last item you added, if you can reproduce it then please create a new issue for it.

MartinaeyNL:
Quickly went through the changes as requested by @DonWillems . The changes look fine to me.
In addition to that; I'd really like to clean up the asset page (and the asset viewer) in the future, as it's really messy.

Issue #1488, opened 2024-11-11
Flow rules: add processors


pierrekil:
Add processors for flow rules:

- [x] Sum (multiple inputs)
- [x] Historic value
- [x] Average
- [x] Max
- [x] Min
- [ ] Differentiate
- [ ] Integrate

see: https://docs.google.com/document/d/1A6ZSy1VM4bHI__IA1tEcjAOsv-LvFCARy4WclhGZUio?usp=drive_fs

Example use case: 
- Forum example: https://forum.openremote.io/t/energy-consumption-graph-in-openremote-for-different-time-intervals/3481?u=pierre
- For calculation of e.g. 'Self consumption or Autarky over a period': see https://docs.google.com/document/d/1YIk-LVrIlJ-b9EOQY14wm0Pbte166qC4/edit#bookmark=id.qm57yf7djorx

![Image](https://github.com/user-attachments/assets/42acc59d-5d31-45b0-b37b-0b0d781b190f)

pankalog:
Notes

Derivative block has a value as an input, and behind the scenes, it goes in and retrieves the latest datapoint value for the attr. Only input is the TimeUnit (per second, per hour, per day).
Same thing for integral.
Sum - unchanged
Historical value: Stopwatch image - can pick a datapoint from x amount of TimeUnit
Constant time DP retrieval: Clock - Retrieves a single datapoint from a specific time (Midnight of 1/1/2024)
Only allow derivative/integral from read attribute inputs, not from calculations

Steps:
Start with Sum
Min-max
Average-mean
Derivative
integral

DonWillems:
Updated the doc and issue description. Reduced functionality to introduce to 'Sum', 'Historic data', and after that 'Min', 'Max', 'Average' (mean).

pierrekil:
For historic value, take the most recent value for the requested time, not an average.

pankalog:
> For historic value, take the most recent value for the requested time, not an average.

@pierrekil @richturner `AssetDatapointNearestQuery`, the datapoint query type I made for this, retrieves the database nearest to the date specified. So from the screenshot in #1525 , It'll take the current server time, remove 10 days, and try to retrieve the nearest datapoint to that time. For each datapoint, we take the absolute value of the difference between that timestamp and each datapoint, and returns a single datapoint that has the smallest absolute value. So no averages are being taken, we are talking about 1 datapoint.

Hackerberg43:
[...]

PR #1525, opened 2024-12-06
[Flow rules] Add historic value, sum, maximum, minimum, average, median, derivative, and integral blocks

pankalog:
This PR adds several blocks to the flow rules, namely:
	- `HISTORIC_VALUE`,
	- `SUM_PROCESSOR`,
	- `MAX_PROCESSOR`,
	- `MIN_PROCESSOR`,
	- `AVG_PROCESSOR`,
	- `MEDIAN_PROCESSOR`
	- `DERIVATIVE`
	- `INTEGRAL`

Main issue was allowing the input of multiple lines into a single socket. After that, some UI changes to allow multiple pickers to be created, along with their styling, using `NodeInternal.BreakType`. Partial support for array types of any primitive type were also added, by adding `_ARRAY` after the data type, and adding that to `NodeDataType`. 

image


pankalog:
fixes #1488

pankalog:
TODO: Still need some guidance about tests, and removing `DEBUG_TO_CONSOLE` output block.

wborn:
I retriggered the build as it seems to fail on the `MQTTClientProtocolTest` which I've also seen failing in other PRs (#1528)

richturner (changes requested):
Hey,

I see various changes to existing code (styling, files deleted etc.) but I don't know the reason for these, some comments explaining what is being fixed/improved would be helpful.

ui/component/or-rules/src/flow-viewer/components/internal-picker.ts
```diff
@@ -108,7 +109,7 @@ export class InternalPicker extends translate(i18next)(LitElement) {
             case PickerType.MULTILINE:
                 return this.multilineInput;
             case PickerType.NUMBER:
-                return this.numberInput;
+                return this.getNumberInput(0, 365);
```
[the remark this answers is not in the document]
pankalog: We worked on this with Martin, and since it only makes sense to have a max number of 365 (365 days in a year). This helps with spacing the input field.

manager/src/main/java/org/openremote/manager/rules/RulesFacts.java
```diff
@@ -489,6 +489,10 @@ public boolean logVars(Logger logger) {
         }
         return haveLog;
     }
+	public void logVars(Logger logger, Level level, String string) {
+		logger.log(level, string);
+	}
+
```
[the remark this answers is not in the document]
pankalog: I need this in `FlowResourceImpl` to read `OR_DEV_MODE` that you mentioned below.

manager/src/main/java/org/openremote/manager/rules/RulesService.java
```diff
@@ -226,7 +226,8 @@ public void init(Container container) throws Exception {
         container.getService(ManagerWebService.class).addApiSingleton(
             new FlowResourceImpl(
                 container.getService(TimerService.class),
-                container.getService(ManagerIdentityService.class)
+                container.getService(ManagerIdentityService.class),
+                container
```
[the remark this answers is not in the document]
pankalog: I need this in `FlowResourceImpl` to read `OR_DEV_MODE` that you mentioned below.

manager/src/main/java/org/openremote/manager/rules/flow/FlowResourceImpl.java
```diff
@@ -15,25 +15,39 @@ public class FlowResourceImpl extends ManagerWebResource implements FlowResource
...
+    public FlowResourceImpl(TimerService timerService, ManagerIdentityService identityService, Container container) {
         super(timerService, identityService);
-        for (Node node : Arrays.stream(NodeModel.values()).map(NodeModel::getDefinition).toArray(Node[]::new)) {
-            LOG.finest("Node found: " + node.getName());
+        Arrays.stream(NodeModel.values()).map(NodeModel::getDefinition).forEach(n -> {
+            LOG.finest("Node found: " + n.getName());
+        });
+
+        // If OR_DEV_MODE, then include the DEBUG_TO_CONSOLE block.
+        if(MapAccess.getBoolean(container.getConfig(), "OR_DEV_MODE",false)){
```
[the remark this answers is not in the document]
pankalog: I made a getter for `RulesFacts.LOG` for this, is that okay?

manager/src/main/java/org/openremote/manager/rules/flow/NodeModel.java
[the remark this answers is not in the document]
pankalog: I also agree that the return value of `NodeImplementation.execute()` needs a bit more work, but those changes were way too convoluted to perform for now unfortunately.

If you want me to take up that task, we can make a new issue and I'll take it up.

ui/component/or-rules/src/flow-viewer/components/flow-node.ts
```diff
@@ -79,21 +94,29 @@ export class FlowNode extends SelectableElement {
         this.style.boxShadow = this.selected ? "var(--highlight) 0 0 0 3px" : "";
 
         const title = this.minimal ?
-            html`<div class="title minimal" ?singlechar="${this.node.displayCharacter!.length === 1}">${i18next.t(this.node.displayCharacter!)}</div>` :
-            html`<div class="title ${this.node.type!.toLowerCase()}" @mousedown="${this.startDrag}">${i18next.t(this.node.name!) || "invalid"}</div>`;
+            html`<div class="title minimal" ?singlechar="${this.node.displayCharacter!.length === 1}">${String(i18next.t("flow."+this.node.displayCharacter!, this.node.displayCharacter!))}</div>` :
```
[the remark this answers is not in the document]
pankalog: Fixed

manager/src/main/java/org/openremote/manager/rules/flow/NodeModel.java
```diff
@@ -23,6 +30,7 @@ public enum NodeModel {
                 AttributeInternalValue assetAttributePair = ValueUtil.JSON.convertValue(info.getInternals()[0].getValue(), AttributeInternalValue.class);
                 String assetId = assetAttributePair.getAssetId();
                 String attributeName = assetAttributePair.getAttributeName();
+                System.out.println(new AttributeRef(assetId, attributeName));
```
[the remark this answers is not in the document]
pankalog: Removed all `System.out.println`s from `NodeModel`

manager/src/main/java/org/openremote/manager/rules/RulesFacts.java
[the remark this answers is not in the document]
pankalog: Removed

ui/component/or-rules/src/flow-viewer/components/internal-picker.ts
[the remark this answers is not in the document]
MartinaeyNL: We indeed had a conversation about this.
However, I mentioned this should only be applied to your "date" field, instead of all numbers in FLOW rules.

MartinaeyNL, line 112: This minimum of 0, and maximum of 365, should only be applied to your "date" field.
Now it applies to ALL numbers in the FLOW rules.

MartinaeyNL (changes requested):
Wanted to address an specific issue on the `internal-picker` file.
Don't see this as a full code review.

manager/src/main/java/org/openremote/manager/rules/RulesService.java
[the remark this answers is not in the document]
richturner: I think a `LOG NODE` should exist always so the need to read `OR_DEV_MODE` goes away doesn't it?

manager/src/main/java/org/openremote/manager/rules/flow/FlowResourceImpl.java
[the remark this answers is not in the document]
richturner: It needs to use the Logger from the RulesetDeployment as each rule has its' own logger, you should be able to get hold of this during building

ui/component/or-rules/src/flow-viewer/components/internal-picker.ts
pankalog: @MartinaeyNL it now uses `this.getNumberInput(undefined, undefined);` which uses `ifDefined` to set those limits, practically giving the same as `this.numberInput` from `master`.

manager/src/main/java/org/openremote/manager/rules/flow/NodeExecutionRequestInfo.java
```diff
@@ -94,15 +97,35 @@ public NodeExecutionRequestInfo(NodeCollection collection, Node node, NodeSocket
...
+        return dict.entrySet().stream()
+                .map(entry -> {
+                    Node node = entry.getValue();
+                    final Object execute = NodeModel.getImplementationFor(node.getName()).execute(
+                            new NodeExecutionRequestInfo(getCollection(), entry.getValue(), entry.getKey(), getFacts(), getAssets(), getUsers(), getNotifications(), getHistoricDatapoints(), getPredictedDatapoints(), LOG)
+                    );
+                    return execute;
+                })
+                .toArray();
+    }
```
richturner, line 128: I don't see much value in this apart from avoiding code reuse, see suggestion for SUM_PROCESSOR

manager/src/main/java/org/openremote/manager/rules/flow/NodeExecutionRequestInfo.java
```diff
@@ -205,6 +228,8 @@ public RulesFacts getFacts() {
         return facts;
     }
 
+    public Logger getLog(){return LOG;}
```
richturner, line 231: We like spaces and new lines in our code base, but anyway this can be removed just make the `LOG` field `protected`.
```
    public Logger getLog() {
    return LOG;}
```

model/src/main/java/org/openremote/model/rules/flow/FlowResource.java
```diff
@@ -26,4 +26,10 @@ public interface FlowResource {
     @Produces(APPLICATION_JSON)
     @Operation(operationId = "getNodeDefinition", summary = "Retrieve a node definition by name")
     Node getNodeDefinition(@BeanParam RequestParams requestParams, @PathParam("name") String name);
+
+    @POST
+    @Path("socket/matcher/{from}/{to}")
+    @Produces(APPLICATION_JSON)
+    @Operation(operationId = "getDoesMatch", summary = "Check if socket types match")
+    boolean getDoesMatch(@BeanParam RequestParams requestParams, @PathParam("from") NodeDataType fromType, @PathParam("to") NodeDataType toType);
```
richturner, line 34: Do we need to move this to the backend? What is the benefit for the added load on the backend?

ui/app/shared/locales/en/or.json
richturner, line 1: Will need to move keys in all language files not just this one

ui/component/or-rules/src/flow-viewer/components/flow-node-socket.ts
```diff
@@ -52,7 +51,10 @@ export class FlowNodeSocket extends translate(i18next)(LitElement) {
     }
 
     public get socketTypeString() {
-        return this.socket.type!.toString().toLowerCase();
+        const socketType: string = this.socket.type!.toString().toLowerCase();
+        // If the socket type is NUMBER_ARRAY, remove the array part, since we only allow the input of NUMBER into that
+        // socket, just multiple ones. Also helps with displaying the correct color.
+        return socketType.replace("_array", "")
```
richturner, line 57: This looks hacky

ui/component/or-rules/src/flow-viewer/components/flow-node.ts
```diff
@@ -79,21 +94,29 @@ export class FlowNode extends SelectableElement {
...
-        ${(this.minimal) ? null : html`<div class="internal-container">${this.node.internals!.map((i) =>
-            html`<internal-picker style="pointer-events: ${(this.frozen ? "none" : "normal")}" @picked="${async () => {
-                this.forceUpdate();
-                await this.updateComplete;
-                project.removeInvalidConnections();
-            }}" .node="${this.node}" .internalIndex="${this.node.internals!.indexOf(i)}"></internal-picker>`)}</div>`}
+        ${(this.minimal) ? null : html`<div class="internal-container" style="display: flex; flex-wrap: wrap; justify-content: flex-start; max-width: 190px; padding: 10px; gap: 8px; height: fit-content;">
+            ${this.node.internals!.map((i) => {
+                const isNewLine = i.breakType === NodeInternalBreakType.NEW_LINE;
+                const style = isNewLine ? "flex-basis: 100%;" : "";
```
richturner, line 109: What is this styling and newline logic?

model/src/main/java/org/openremote/model/rules/flow/NodeDataType.java
```diff
@@ -2,6 +2,8 @@
 
 public enum NodeDataType {
     NUMBER,
+    NUMBER_ATTRIBUTE,
```
richturner, line 5: No usages of this

model/src/main/java/org/openremote/model/rules/flow/NodeDataType.java
```diff
@@ -2,6 +2,8 @@
...
+    NUMBER_ARRAY,
```
richturner, line 6: I don't see what value this is adding as it only seems to be used for matching logic

model/src/main/java/org/openremote/model/rules/flow/Node.java
```diff
@@ -4,6 +4,7 @@ public class Node {
     // Generation of ID is the responsibility of the front end and is unique for every node in a structure, not for every node in the system
     private String id;
     private NodeType type;
+    private int priority = 999;
```
richturner, line 7: Where is this used?

model/src/main/java/org/openremote/model/rules/flow/Node.java
```diff
@@ -12,6 +13,18 @@ public class Node {
...
+        this.type = type;
+        this.priority = priority;
+        this.position = new NodePosition(0, 0);
+        this.size = new NodePosition(0, 0);
+        this.internals = internals;
+        this.inputs = inputs;
+        this.outputs = outputs;
+        this.displayCharacter = null;
+    }
+
```
richturner, line 27: Where is this used?

richturner (changes requested):
Seem to be quite a few changes that are unused and also some changes (styling) don't seem to relate to the original issue and no explanation of why they are needed.

manager/src/main/java/org/openremote/manager/rules/flow/NodeExecutionRequestInfo.java
pankalog: Since `getValueFromInput` does not really support retrieving the values from multiple different previous nodes, I created this method to simplify that process. This method just figures out which nodes are connected to the node we are currently on, and then requests the execution of all nodes. It then returns the result as an array of Objects. This does most of the heavy lifting for allowing a socket to request the results from multiple different inputs to that socket.

model/src/main/java/org/openremote/model/datapoint/query/AssetDatapointTimestampQuery.java
```diff
@@ -0,0 +1,44 @@
...
+        return "WITH nearest_row AS ( " +
+                "    SELECT * " +
+                "    FROM " + tableName + " " +
+                "    WHERE entity_id = ? AND attribute_name = ? " +
+                "      AND timestamp <= to_timestamp(?) " +
+                "    ORDER BY timestamp DESC " +
+                "    LIMIT 1 " +
+                ") " +
+                "select timestamp as X, cast(value as numeric) as Y FROM nearest_row";
+    }
```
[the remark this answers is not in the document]
pankalog: Okay, this is the first time I was working with TimescaleDB and from what I read, if I used that `nearest_row` query first, it should for sure use the indices in the table.

model/src/main/java/org/openremote/model/rules/flow/FlowResource.java
pankalog: This used to be in `socket.type.matcher.ts`, it had a TODO of moving this to the backend, and I thought it was a good change, since we're enforcing the rules of matching sockets to other sockets by moving it to the backend. I also feel that this gives a bit more certainty, since the other version just used a hard-coded JSON object for the matching.

model/src/main/java/org/openremote/model/rules/flow/Node.java
pankalog: @pierrekil asked me to make it so that the processor blocks are displayed on the UI in a specific order, so that priority variable is what we use on all processors in the frontend to sort the processors in the desired order.

model/src/main/java/org/openremote/model/rules/flow/NodeDataType.java
pankalog: I couldn't really think of another way of allowing arrays of objects without seriously affecting the rest of the codebase. I think it makes sense, since it is a new data type that is supported, and has different properties than the rest of the data types. Do you have any other ideas?

ui/app/shared/locales/en/or.json
pankalog: Is there a fast way to do that?

ui/component/or-rules/src/flow-viewer/components/flow-node-socket.ts
pankalog: This is only used for properly coloring, naming the socket, and properly defining the input type of the socket that is created.  So for example the `SUM_PROCESSOR`, even if it has `NUMBER_ARRAY` as an input socket, it should not accept `NUMBER_ARRAY` inputs, but it should accept `NUMBER` inputs. The code of the processor itself collects the results of the connected input nodes and creates the number array, and then serves a number as output.

ui/component/or-rules/src/flow-viewer/components/flow-node.ts
pankalog: This is part of the formatting changes I have made for node internals, the `HISTORIC_VALUE` block contains 3 inputs that should be formatted correctly. `NodeInternalBreakType` deines `NEW_LINE` and `SPACER`, which is used to either break the next element to a new line in the UI, or add a spacer (keep everything in the same line).

ui/component/or-rules/src/flow-viewer/components/internal-picker.ts
```diff
@@ -107,8 +108,10 @@ export class InternalPicker extends translate(i18next)(LitElement) {
                 return this.dropdownInput;
             case PickerType.MULTILINE:
                 return this.multilineInput;
+            case PickerType.DATE:
+                return this.getNumberInput(0, 365);
```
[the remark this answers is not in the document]
pankalog: If I simply use the previous implementation of `getNumberInput` the number field is way too wide (Look at the NUMBER input block). Since people won't be using that big of a number when defining a period of time from/since now, I modified `NumberInput` to allow a variable amount of width depending on the situation. I believe that the biggest amount of time that they may reasonably require is 365, and it doesn't limit the actual number input, just the size of the number field.

model/src/main/java/org/openremote/model/datapoint/query/AssetDatapointTimestampQuery.java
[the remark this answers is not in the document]
richturner: Good to be thinking about indexes, you can use explain to see if indexes are being utilised for a given query and you shouldn't need a CTE here.

richturner (changes requested):

ui/app/shared/locales/en/or.json
richturner: Not that I can think of, just replicate what you have done for en

test/src/test/groovy/org/openremote/test/rules/residence/FlowRulesTest.groovy
```diff
@@ -74,4 +84,73 @@ class FlowRulesTest extends Specification implements ManagerContainerTrait {
         }
 
     }
+
+    def "Test HISTORIC_VALUE and SUM blocks"() {
```
richturner, line 88: Would be good to have other added functionality covered in test

richturner (changes requested):
Looking good, just need to update al or.json language files as you've moved flow strings and add additional testing.

test/src/test/groovy/org/openremote/test/rules/residence/FlowRulesTest.groovy
```diff
@@ -85,7 +85,7 @@ class FlowRulesTest extends Specification implements ManagerContainerTrait {
 
     }
 
-    def "Test HISTORIC_VALUE and SUM blocks"() {
+    def "Test newly-added blocks"() {
```
richturner, line 88: newly added is a time sensitive concept so please expand

test/src/test/resources/org/openremote/test/rules/HistoricAndNewProcessorTest.flow
richturner, line 1: Don't use New in file names, is it still new in 6 months?

Just use the same name you have given the ruleset in the test `HistoricValueAndProcessors`

test/src/test/groovy/org/openremote/test/rules/residence/FlowRulesTest.groovy
```diff
@@ -124,15 +128,18 @@ class FlowRulesTest extends Specification implements ManagerContainerTrait {
             assert dps.size() == 4
         }
 
-        when: "I make a HISTORIC_VALUE and SUM blocks"
-        String json = getClass().getResource("/org/openremote/test/rules/HistoricAndSumTest.flow").text
+        when: "I add the test"
```
richturner, line 131:
```suggestion
        when: "A flow rules is added"
```

test/src/test/groovy/org/openremote/test/rules/residence/FlowRulesTest.groovy
```diff
@@ -142,15 +149,19 @@ class FlowRulesTest extends Specification implements ManagerContainerTrait {
         conditions.eventually {
             assert rulesService.globalEngine.get() != null
             assert rulesService.globalEngine.get().isRunning()
-            assert rulesService.globalEngine.get().deployments.values().any({ it.name == "HistoricValueAndSum" && it.status == DEPLOYED})
+            assert rulesService.globalEngine.get().deployments.values().any({ it.name == "HistoricValueAndProcessors" && it.status == DEPLOYED})
         }
 
         and: "After the rule runs, the value the attribute had at that timestamp is written to the output attribute"
```
richturner, line 155:
```suggestion
        and: "The rule runs and the various attributes are updated accordingly"
```

richturner (changes requested):
Just some minor points

richturner (approved):

PR #1532, opened 2024-12-10
Fix error when deleting a gateway asset (Fixes #1531)

### LINKED ISSUE #1531: Cannot delete gateway asset under certain conditions
* Add a gateway asset to the asset tree root
* Connect the edge gateway
* Move the gateway asset to a child of another asset
* Try and delete the gateway asset
* If the gateway asset ID is alphabetically before any of the gateway asset's direct children then the delete will fail

This is caused by the existing `update_asset_parent_info` DB trigger function not updating the paths of all descendants of a modified asset.

richturner:


wborn (dismissed):
LGTM!

wborn:
I've retriggered the build to see if the failing tests are related to your changes.

wborn:
Looks like the CI/CD does not succeed because there is always one of those [unstable test ](https://github.com/openremote/openremote/issues?q=is%3Aissue+is%3Aopen+label%3ATest) that fails.
WIthout CI/CD succeeding the merge button remains disabled.

richturner:
I'm looking into the test issues at the moment

richturner:
There is still some additional work needed as the trigger function is not working correctly when an asset is re-parented to the root of the tree

richturner:
@wborn I dicovered a problem where an additional SQL function was created rather than replacing the existing one due to the signature change, I have now fixed this.

wborn (approved):

PR #1533, opened 2024-12-10
Added support for OR_DATA_POINTS_MAX_QUERY_POINTS

### LINKED ISSUE #1490: API: limit amount of data retrievable with datapoint query
If no query parameters are given for asset/datapoint/[assetID]/[attributeName] all datapoints are returned. This could be a huge amount, resulting in an overburdened system.

MartinaeyNL:
Fixes #1490 .

( Will add changelog later )


MartinaeyNL:
Sorry for the additional commit history; I accidentally branched this from a feature branch, instead of `master`.

MartinaeyNL:
**Important:** Before merging, the changes on file `profile/demo.yml` should be reverted, as it is only meant for demo purposes.

profile/demo.yml
```diff
@@ -103,6 +103,7 @@ services:
       OR_ADDITIONAL_HOSTNAMES:
       OR_WEBSERVER_ALLOWED_ORIGINS: ${OR_WEBSERVER_ALLOWED_ORIGINS:-}
       OR_DEV_MODE: ${OR_DEV_MODE:-false}
+      OR_DATA_POINTS_QUERY_LIMIT: 3
```
wborn, line 106: Don't forget to remove this. :wink: 

```suggestion
```

MartinaeyNL: No problem! 👍

model/src/main/java/org/openremote/model/datapoint/DatapointQueryTooLargeException.java
```diff
@@ -0,0 +1,8 @@
+package org.openremote.model.datapoint;
```
wborn, line 1: The header is missing from this file.

MartinaeyNL: Thanks! Good spot 👍

MartinaeyNL:
Sorry about that, should all be OK now 👍

Issue #1448, opened 2024-10-18
Rules: When-Then: Add 'duration' to a condition on the LHS, to trigger the RHS if it has been true for the set time


DonWillems:
As a user I want to create a condition that has to be true for x minutes, so that the action is not triggered immediately.

e.g.
As a user I want to specify that the connection status of a device has to be false for more than 10 minutes before sending a notifications, so that it does not send notifications for each brief internet disruption of a device.

The duration can be added to every condition.

- Add duration mechanic in the JSON rules to track how long a condition is true
- Add controls to When-Then UI so that the user can set it.
   - The duration button appears together with the value field. If it is pressed, the duration field appears
   - The duration can be set in minutes (positive values only. Do we need an upper limit?)
   - if left empty, no duration is saved. If you would open the rule later, the duration field will not be shown anymore. The user can press the button again to open the field.
   - The 'not updated for' operator will force this field, see issue #1449 

![Image](https://github.com/user-attachments/assets/2988d2c6-c62f-47ed-9942-f4a6d3a5a660)

richturner:
There is `RuleCondition.duration` in the model and implemented in the backend but this is an alternative LHS trigger to `RuleCondirion.assets`, this `RuleCondition.duration` is not used in the UI at all and is not utilised in the code base anywhere so this can be repurposed.

To implement this functionality look to enhance the `JsonRulesBuilder.assetPredicate` here:

https://github.com/openremote/openremote/blob/ca7aa7e517d53493726d5b153cc5bbdac012fda1/manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java#L197

Create a wrapper around the existing predicate that also stores and applies any time related state.

dominiquekleeven:
Had a talk with Rich regarding the existing duration property.

The duration property needs to be changed in RuleCondition to a map, allowing us to associate an attribute with a specific duration, since we can have multiple attributes with a duration as part of a condition and these located on a different level

PR #1534, opened 2024-12-10
Added support for duration in When-Then LHS conditions

dominiquekleeven:
#1448 

Added the ability to specify a map of durations as part of the ruleCondition model, key is the index of the attributePredicate and value is the ISO8601 string.

This is only the **back-end implementation** and should be done, needs a review to ensure nothing was missed/implementation is correct.

Front-end implementation will be a seperate PR


manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
```diff
@@ -184,65 +185,114 @@ public static Predicate<NameValueHolder<?>> asPredicate(Supplier<Long> currentMi
...
         List<Function<Collection<AttributeInfo>, Set<AttributeInfo>>> assetStateMatchers = new ArrayList<>();
-        List<Predicate<AttributeInfo>> attributePredicates = new ArrayList<>();
+        Map<Integer, List<Predicate<AttributeInfo>>> attributePredicates = new HashMap<>();
 
-        if (condition.getItems().size() > 0) {
+        if (!condition.getItems().isEmpty()) {
+            condition.getItems().stream().forEach(p -> {
+                int index = condition.getItems().indexOf(p);
+                attributePredicates.put(index, new ArrayList<>());
+                attributePredicates.get(index).add((Predicate<AttributeInfo>) (Predicate) asPredicate(currentMillisProducer, p));
```
richturner, line 216: This new map and attributeConditionTimeState object seems un-neccessary:

Couldn't you do this:

```
attributePredicates.add((Predicate<AttributeInfo>)(Predicate)asPredicate(currentMillisProducer, p, ruleCondition.duration != null ? ruleCondition.duration.get(i) : null));
```

All state can just be fields within the predicate rather than having the `AttributeConditionTimeState` `Map`

richturner (changes requested):

manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
dominiquekleeven: Removed both the map and TimeState

Duration is directly passed to the asPredicate, and the predicate now deals with the duration match state.

manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
```diff
@@ -175,8 +178,28 @@ public static Predicate<NameValueHolder<?>> asPredicate(Supplier<Long> currentMi
             });
         }
 
-        return nameValueHolder -> namePredicate.test(nameValueHolder.getName())
-            && valuePredicate.test(valueExtractor.get().apply(nameValueHolder));
+        AtomicReference<Map<String, Long>> matchTimestamps = new AtomicReference<>(new HashMap<>());
```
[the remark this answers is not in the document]
dominiquekleeven: Added a custom asAttributeMatcher  as discussed, which is added within the JsonRulesBuilder. Also I reset the AssetQueryPredicate to its upstream state since no changes were needed there anymore. 

The duration handling is then done after the custom assetPredicate is applied and the matches are returned.

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -286,15 +286,85 @@ void update(Map<String, Long> nextRecurAssetIdMap) {
...
+
+                        // If the assetDurationPredicate is set then we use that to to match the asset states
+                        if (assetDurationPredicate != null) {
+                            // Pair of the attribute info that match and the index of the duration map
+                            Set<Pair<AttributeInfo, Integer>> matches = assetDurationPredicate.apply(states);
+                            
+                            if (matches != null) {
+                                long currentTime = timerService.getCurrentTimeMillis();
+                                
+                                matches.forEach(pair -> {
```
richturner, line 301: Rename `pair` to something more meaningful e.g. `attributeInfoAndMatchedPredicateIndex`

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -286,15 +286,85 @@ void update(Map<String, Long> nextRecurAssetIdMap) {
...
+                                    AttributeInfo matchedState = pair.getKey();
+                                    Pair<AttributeRef, Integer> matchedStateRef = new Pair<>(matchedState.getRef(), pair.getValue());
```
richturner, line 303: Add a comment explaining what this is and why needed.
Rename `refAndMatchedPredicateIndex`

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
richturner, line 301: Don't get the nested loop over the matches and the multiple loops.
```
matches.stream().map(
  // Apply duration logic and return AttributeInfo if matched else remove from durationsMap and return null
).filter(Objects::nonNull)
.collect(Collectors.toSet());
```

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -1407,4 +1476,84 @@ protected static SunTimes.Parameters getSunCalculator(Ruleset ruleset, SunPositi
 
         return sunCalculator;
     }
+
+
+     /**
+     * Custom function that mimics the logic of {@link AssetQueryPredicate.asAttributeMatcher} but returns a {@link Pair} of {@link AttributeInfo} and {@link Integer}
```
richturner, line 1482: Explain what the Integer indicates

richturner (changes requested):
Almost there, well written test case!

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
dominiquekleeven: Was unneeded only used for some logging I've added during implementation. Will remove

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
dominiquekleeven: Added explanation for the integer

DonWillems:
When combining a duration condition with a non-duration condition of the same asset, it will also require the non-duration attribute to by true for the duration of the other condition.

Reproduce:
1. Create When-Then rule with When "Harvest Robot 1" "Vegetable type" equals "CUCUMBER" for 5 minutes, and "Speed" equals 5 (or shorter duration :) ), Then "Harvest Robot 1" "Notes" is "test"
2. Set Vegetable type of Harvest Robot 1 to CUCUMBER
3. Wait 5 minutes
4. Set Speed of Harvest Robot 1 to 5
5. Wait 5 minutes
6. Notes of Harvest Robot 1 becomes "test"

Expected:
5. Notes of Harvest Robot 1 becomes "test" immediately when speed is 5 and the condition is true, since the other condition was already met for 5 minutes.

dominiquekleeven:
> When combining a duration condition with a non-duration condition of the same asset, it will also require the non-duration attribute to by true for the duration of the other condition.
> 
> Reproduce:
> 
>     1. Create When-Then rule with When "Harvest Robot 1" "Vegetable type" equals "CUCUMBER" for 5 minutes, and "Speed" equals 5 (or shorter duration :) ), Then "Harvest Robot 1" "Notes" is "test"
> 
>     2. Set Vegetable type of Harvest Robot 1 to CUCUMBER
> 
>     3. Wait 5 minutes
> 
>     4. Set Speed of Harvest Robot 1 to 5
> 
>     5. Wait 5 minutes
> 
>     6. Notes of Harvest Robot 1 becomes "test"
> 
> 
> Expected: 5. Notes of Harvest Robot 1 becomes "test" immediately when speed is 5 and the condition is true, since the other condition was already met for 5 minutes.

Was able to reproduce this and also found the cause of the issue, will try and fix it during the weekend if time permits otherwise on monday.

dominiquekleeven:
> When combining a duration condition with a non-duration condition of the same asset, it will also require the non-duration attribute to by true for the duration of the other condition.
> 
> Reproduce:
> 
>     1. Create When-Then rule with When "Harvest Robot 1" "Vegetable type" equals "CUCUMBER" for 5 minutes, and "Speed" equals 5 (or shorter duration :) ), Then "Harvest Robot 1" "Notes" is "test"
> 
>     2. Set Vegetable type of Harvest Robot 1 to CUCUMBER
> 
>     3. Wait 5 minutes
> 
>     4. Set Speed of Harvest Robot 1 to 5
> 
>     5. Wait 5 minutes
> 
>     6. Notes of Harvest Robot 1 becomes "test"
> 
> 
> Expected: 5. Notes of Harvest Robot 1 becomes "test" immediately when speed is 5 and the condition is true, since the other condition was already met for 5 minutes.

This issue should be fixed with the latest commits, I've tried it out and could no longer reproduce it. 

Keep in mind these changes aren't (yet) on the UI branch, so this branch still needs to be merged into that one so it can be tested again).

dominiquekleeven:
Still looking into an issue where it sometimes does track duration properly, causing the rule to never trigger

richturner:
@dominiquekleeven Hey having spent some time looking at the logic here I have made some adjustments to simplify and improve efficiency, please review my commit and happy to have a call to go through it.

richturner (approved):

MartinaeyNL:
Deployed this together with the Rules UI on `test1`. Waiting for @DonWillems to review.
After that I'll request @wborn for a code review, so we can try to merge this today. 👍

richturner (approved):

PR #1535, opened 2024-12-10
Enable idle timeout for websocket connections

richturner:
This PR will ensure a websocket connection will timeout if the underlying socket is broken due to connection issues. The ping/pong used by clients should prevent the idle timeout triggering under normal conditions.


wborn (approved):
Thanks!

PR #1539, opened 2024-12-12
enhancement/tableWidgetImprovement

Hackerberg43:
I have added the functionality of selecting all assets of type to the table-widget
With the current functionality, you need to add every asset manually. This is still the default behavior.
Now with this toggle, when a new asset of the type is created, it will be automatically added to the table-widget, just like the map-widget behaves.

It felt like it made the most sense to add the functionality to the assettypes-panel.ts function, however as enabling/disabling specific asset selection was designed as an input of this function, I've added it to the table-widget itself.

I've tested it and it works fine, but I'm not one to trust (yet) 0:)

image
image

Fixes https://github.com/openremote/vdl-es/issues/92


DonWillems:
Nice addition :)
I would rather have this as an option to the assettypes-panel.ts though. Then it can be used in other places too, and the toggle can go after the asset types field, and before the asset selection button. The order in your screenshots separates the toggle from the asset select. I'll ask @MartinaeyNL if he has some time for these adjustments 👍

Hackerberg43:
Ah okay, I will try to fix it that way, it is good exercise

MartinaeyNL:
The code itself is OK to me. I agree with @DonWillems that it'd be better to move the logic into `assettypes-panel.ts`,
that includes an optional HTML attribute to show/hide the toggle.

Looking forward to the changes 👍

Hackerberg43:
Ok, so I rewrote this to include the base feature in assettypes-panel.ts
image
I also had to edit the map-widget with this change, with the nice addition that it is now also possible to only display selected assets with the map-widget :)
The asset fetching mechanism is a little different in both widgets, I tried to harmonize this but couldn't get that to work. So kept the unique mechanisms for each widget and expanded those.

Point of discussion:
-I haven't tested how these alterations effect an existing deployment. 
-I kept the default behavior identical to how it is now, so a table-widget has allOfType off by default, while the map-widget has it on by default. This will result in that existing deployments will not change their setup. However, I think it would be more convenient if the default setting of both widgets is identical?

Hackerberg43:
Update: Tested on existing deployment.
Both existing widgets get the allOfType=false.  For the map-widget, the existing included assets are converted to selected assets, meaning you will need to set All of this type for every existing widget to restore the previously deployed behavior.
Please let me know which bevavior is desired.

Hackerberg43:
Adding the following to a new sql script in https://github.com/openremote/openremote/tree/master/manager/src/main/resources/org/openremote/manager/setup/database will fix the issue with existing deployments, all pre-existing widgets will then keep their existing behavior.
```
WITH json_data AS (
    SELECT
        id,
        template,
        jsonb_set(
            template,
            '{widgets}',
            (
                SELECT jsonb_agg(
                    jsonb_set(
                        widget,
                        '{widgetConfig,allOfType}',
                        CASE
                            WHEN (widget ->> 'widgetTypeId') = 'map' THEN 'true'::jsonb
                            WHEN (widget ->> 'widgetTypeId') = 'table' THEN 'false'::jsonb
                            ELSE widget -> 'widgetConfig' -> 'allOfType'
                        END
                    )
                )
                FROM jsonb_array_elements(template -> 'widgets') AS widget
            )
        ) AS updated_json
    FROM dashboard
)
UPDATE dashboard
SET template = json_data.updated_json
FROM json_data
WHERE dashboard.id = json_data.id;
```
Now I could change the default of allOfType to true for the table widget, such that when creating a new widget, identical default settings for both map and table.

DonWillems:
Bug: selected assets of map widget are reset to all asset of a type when clicking the widget in modify mode.

Reproduce:
1. Add map widget, select an asset type with multiple assets
2. Toggle off "All assets of this type"
3. Select e.g. 2 assets
4. Save
5. Go into 'View' mode
6. Go into 'Modify' mode
7. Click the map widget
8. The selected assets are reset to all assets of that type (but the toggle is still off)

Expected:
8. The same 2 assets are selected.

Hackerberg43:
Good catch.
Found two more pre-existing bugs in the fixing progress, so working on that now.
- In the map settings, you can select optional attributes part of the asset, which currently aren't even active. Will result in a snackbar error therefore.
- Integer attributes are not showing up in the list, while numbers and positive integers are. I will review the different value types.

Please ignore the last commit, that was a late night quick fix ;)

MartinaeyNL (dismissed):

Hackerberg43:
Hey Martin, the sql script mentioned above is not yet integrated in the setup/database dir.  I'm not sure if I know the correct procedure.
Apart from that it is ready to go.

richturner (dismissed):

DonWillems:
> @DonWillems @MartinaeyNL A decision is needed about whether the existing behaviour should be maintained for map widgets given the requested changes have altered the schema.
> 
> If existing behaviour is desired (which seems sensible) then the SQL script @Hackerberg43 has provided should be included in the PR which is why I have put it to draft for now.

It makes sense to keep the existing behaviour indeed. Is there a downside to running the SQL script? Otherwise I would say add it

Hackerberg43:
@richturner migration script added with date set to today.

richturner (approved):

PR #1543, opened 2024-12-13
Bug/momentary button

Hackerberg43:
Fixes:
-the attribute format asMomentary flag, which resulted in an invisible button
-The attribute configuration items menu selectable "Momentary" MetaItem did not result in a momentary button.

Points of discussion:
-Icon choice
-The fact that it is integrated into or-mcw-input function, so assigns this default icon to any momentary button in the UI if no action/label/icon is supplied. However, is there a use-case for an invisible button anyway ?

image

Fixes https://github.com/openremote/vdl-es/issues/86


richturner:
@DonWillems can this be reviewed?

MartinaeyNL (approved):

PR #1545, opened 2024-12-16
added AWS Marketplace CloudFormation Template

denniskuijs:


.ci_cd/aws/cloudformation-aws-marketplace.yaml
```diff
@@ -0,0 +1,501 @@
...
+    ap-south-2: 
+        ImageId: ami-xxxxxxxxxxxxxxxxx
+    ap-southeast-4: 
+        ImageId: ami-xxxxxxxxxxxxxxxxx
+    il-central-1: 
+        ImageId: ami-xxxxxxxxxxxxxxxxx
+    ca-west-1: 
+        ImageId: ami-xxxxxxxxxxxxxxxxx
+    ap-southeast-5: 
+        ImageId: ami-xxxxxxxxxxxxxxxxx
```
wborn, line 164: Shouldn't there be images for other regions? I ran into issues when deploying the us-east-1 image in eu-west-1.

.ci_cd/aws/cloudformation-aws-marketplace.yaml
```diff
@@ -0,0 +1,501 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
wborn, line 1: Should this always the initial version or can it also be set to a date in 2025?

.ci_cd/aws/cloudformation-aws-marketplace.yaml
```diff
@@ -0,0 +1,501 @@
...
+                    exit 1
+                  fi
+                - 'echo "System packages and Docker updated successfully!"'
+
+Outputs:
+  PublicIP:
+    Description: The Public IP Address of the newly created EC2 instance
+    Value: !GetAtt InstanceOR.PublicIp
+    Export:
+      Name: !Sub "${AWS::StackName}-PublicIP"
```
wborn, line 501: Can you add a new line?

.ci_cd/aws/cloudformation-aws-marketplace.yaml
wborn, line 1: Can you make sure that the parameters match the documentation, see https://github.com/openremote/documentation/issues/42.

.ci_cd/aws/cloudformation-aws-marketplace.yaml
```diff
@@ -0,0 +1,501 @@
...
+      Description: Fill in your Home IP for SSH Access. Set CIDR to x.x.x.x/32.
+      Type: String
+      AllowedPattern: (\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})/(\d{1,2})
+      ConstraintDescription: Must be a valid IP CIDR range of the form x.x.x.x/x.
+    Password:
+      Description: Create an (admin) password for your OpenRemote instance
+      Type: String
+      MinLength: 8
+      AllowedPattern: ^(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z])(?=.*[^0-9a-zA-Z])(?=\S+$).{8,}$
+      ConstraintDescription: Password must contain 1 or more uppercase characters. Password must contain 1 or more digit characters. Password must contain 1 or more special characters. Password must be 8 or more characters in length.
```
wborn, line 64: It only checked the constraint after completing the wizard. Maybe the description can already instruct users to make it a complex password?

.ci_cd/aws/cloudformation-aws-marketplace.yaml
```diff
@@ -0,0 +1,501 @@
...
+    SMTPFrom:
+      Description: Fill in the From e-mail address you want to use for sending e-mails
+      Type: String
+    SMTPProtocol:
+      Description: Select the SMTP protocol you want to use for sending e-mails (smtps = SSL)
+      Type: String
+      AllowedValues:
+        - smtp
+        - smtps
+      Default: smtps
```
wborn, line 96: Would be nice if it is clear that most of these parameters are optional. Users may otherwise quickly get overwhelmed and abort the wizard and not try running OpenRemote at all.

.ci_cd/aws/img/ArchitectureDiagram_v4.png
wborn, line 1: It's probably more useful to have an image like this with some details in the [AWS Marketplace documentation](https://docs.openremote.io/docs/user-guide/deploying/aws-marketplace).

.ci_cd/aws/cloudformation-aws-marketplace.yaml
lmath56: 2010-09-09 is the latest version as per: [https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/format-version-structure.html](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/format-version-structure.html)

.ci_cd/aws/cloudformation-aws-marketplace.yaml
lmath56: As per here [https://docs.aws.amazon.com/marketplace/latest/userguide/cloudformation.html](https://docs.aws.amazon.com/marketplace/latest/userguide/cloudformation.html) we must use this mapping table. However we should use public parameters in AWS Systems Manager Parameter Store. We can use `{{resolve:ssm:/aws/service/ami-amazon-linux-latest/al2023-ami-minimal-kernel-default-arm64}}` instead of the initial AMI ID

.ci_cd/aws/cloudformation-aws-marketplace.yaml
lmath56: Done - will include with new PR.

.ci_cd/aws/cloudformation-aws-marketplace.yaml
lmath56: Added the password complexity note to the description of the parameter.

.ci_cd/aws/cloudformation-aws-marketplace.yaml
lmath56: After discussing this with the group, as part of the deployment checks with AWS Support we were required to complete AMI scanning, which cannot be done on an default AWS image. So we have created a copy ami-00b777c899f47763b and scanned this as per AWS support's guidance. Once the marketplace listing is approved this AMI will be duplicated to other regions, and the placeholder ami-xxx will automatically map to the duplicated image for each region.

Issue #964, opened 2023-02-17
Rules: two conditions with 'Any of this type' in When AND block do not trigger rule


DonWillems:
Reproduce:
- Create a rule with two conditions using 'Any of this type' in the When block (so AND). They can be of the same or a different asset type.
- The rule is not trigger when both become true.
- If one or both of these rules are set to a specific asset instead of Any of this type, the rule will trigger.

Expected:
- Rule will trigger when e.g. any LightAsset matches condition one and any other LightAsset matches the other condition. The same for two different Asset types used in the conditions.

dominiquekleeven:
I am not able to reproduce this issue on master 

Here is the rule setup I tested:
![Image](https://github.com/user-attachments/assets/d3748cc6-e26e-45be-ba5b-4cf48508a6fb)

## Tested the following scenarios:
Single asset matches conditions:
- **A single light asset** matches both of the LHS conditions > Notes gets updated to **Matched** as expected

Two assets of the same type combined matches the conditions:
- **Two light assets** The first asset matches the first condition, the second the second condition > Notes gets updated to **Matched** for both assets as expected

So the current behaviour functions as expected as described in the issue description
> Rule will trigger when e.g. any LightAsset matches condition one and any other LightAsset matches the other condition. The same for two different Asset types used in the conditions.

dominiquekleeven:
I am not able to reproduce this issue on master 

## First rule setup:
![Image](https://github.com/user-attachments/assets/d3748cc6-e26e-45be-ba5b-4cf48508a6fb)

**Tested the following scenarios**
Single asset matches conditions:
- **A single light asset** matches both of the LHS conditions > Notes gets updated to **Matched** as expected

Two assets of the same type combined matches the conditions:
- **Two light assets** The first asset matches the first condition, the second the second condition > Notes gets updated to **Matched** for both assets as expected

## Second rule setup:
![Image](https://github.com/user-attachments/assets/f9379bab-9616-45d8-96a7-5c6356c90bbb)

**Tested the following scenario**
- Both assets of different types match the rules specified condition > Notes gets updated to **Matched** for the matched LightAsset

So the current behaviour functions as expected as described in the issue description
> Rule will trigger when e.g. any LightAsset matches condition one and any other LightAsset matches the other condition. The same for two different Asset types used in the conditions.

Issue #798, opened 2022-08-25
Rules: trigger a rule with 2 LHS conditions


pierrekil:
Reproduce:
- Create 2 LHS conditions with and AND
- Change both LHS conditions to become TRUE
- Rule is executed
- Change one of the LHS conditions to become FALSE
- Change the same LHS condition to become TRUE again
- Rule is not executed
- Change the second of the LHS conditions to become FALSE
- Change the same LHS condition to become TRUE again
- Rule is executed

Expected:
- Create 2 LHS conditions with and AND
- Change both LHS conditions to become TRUE
- Rule is executed
- Change one of the LHS conditions to become FALSE
- Change the same LHS condition to become TRUE again
- Rule is executed

dominiquekleeven:
> Reproduce (check whether this still happens first)
The issue can be reproduced quite consistently, will investigate the exact cause of why the LHS conditions aren't being properly evaluated.

Note here, I had to change both conditions to false and then true again before the rule would trigger properly. Changing 1 condition to false and then to true again would not trigger the rule.

dominiquekleeven:
> > Reproduce (check whether this still happens first)
> 
> The issue can be reproduced quite consistently, will investigate the exact cause of why the LHS conditions aren't being properly evaluated.
> 
> Note here, I had to change both conditions to false and then true again before the rule would trigger properly. Changing 1 condition to false and then to true again would not trigger the rule.

The above leads me to believe that the root cause is related to the previouslyMatchedAssetState in RuleConditionState, which is used to avoid retriggering the rule THEN action.

PR #1546, opened 2024-12-16
Fix for rules with 2 LHS conditions not properly being able to trigger an action when only one condition becomes false and true again

dominiquekleeven:
#798 

**Root cause:**
When a rule contains multiple LHS conditions that have an AND relationship, the rule action would fail to trigger under certain circumstances. The issue is being caused by how `RuleConditionState` uses `previouslyMatchedAssetStates` to prevent re-triggering an action, and this not being taken into account when handling multiple `RuleConditionState`s

**Scenario that failed:**
1. Rule has two LHS conditions (A AND B)
2. Both conditions evaluate to true → Action triggers (working as expected)
3. Condition A becomes false
4. Condition A becomes true again
5. Action fails to trigger because Condition B is still considered a non-match because it has `previouslyMatchedAssetStates`


dominiquekleeven:
Added a test case for multi LHS conditions, manually tested in demo setup as well, issue seems to be resolved.

@richturner Unsure if this the ideal approach to resolving the issue. 

I've added logic to the updateMatches method in the RuleState class so that when a condition group (with multiple conditions) fails to match, the previouslyMatchedAssetStates is cleared for each of those conditions, provided that at least one of the conditions has no previouslyMatchedAssetStates.

dominiquekleeven:
Potentially also fixes #964 will confirm this later this week as time permits.

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
```diff
@@ -480,6 +480,25 @@ protected boolean updateMatches(LogicGroup<RuleCondition> ruleConditionGroup, Se
...
                         .map(ruleCondition -> conditionStateMap.get(ruleCondition.tag))
                         .allMatch(ruleConditionState -> ruleConditionState.lastEvaluationResult != null && ruleConditionState.lastEvaluationResult.matches);
+                    
+             
+                    // Prevent stale matches from blocking rule re-triggering for multi LHS condition groups
+                    if (!groupMatches && ruleConditionGroup.getItems().size() > 1) {
+
+                        boolean anyConditionHasNoPreviousMatches = ruleConditionGroup.getItems().stream()
+                            .map(ruleCondition -> conditionStateMap.get(ruleCondition.tag))
+                            .anyMatch(ruleConditionState -> ruleConditionState.previouslyMatchedAssetStates.isEmpty());
```
richturner, line 490: The `anyConditionHasNoPreviousMatches` check could be moved into the iteration at line 482 and then this if becomes:

`if (!groupMatches && anyConditionHasNoPreviousMatches && ruleConditionGroup.getItems().size() > 1)`

richturner (changes requested):
Just a minor adjustment to improve performance a little but this solution seems sensible and I gave up trying to think of an alternative without major changes.

manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
dominiquekleeven: Thanks for the review.

I am a bit unsure how to properly move the `anyConditionHasNoPreviousMatches` assignment into the existing stream at line 482, `.allMatches` is a short circuit operation, so it won't properly check all conditionStates for empty previouslyMatchedAssetStates, if I do the assignment inside of the `.map`.

richturner: Yes you're right I was a bit hasty in thinking the `.allMatches()` had to be true but it's the opposite!

richturner (approved):

PR #1550, opened 2024-12-18
Updated username regex pattern based on Keycloak validation

MartinaeyNL:
Currently, the username field only accepts regular A-Z, 0-9 and a few special characters.
This was causing issues in Keycloak configurations where email is used as a unique identifier.

I've updated the regex pattern to allow special characters, and be in line with Keycloak their validation.

According to the [Keycloak changelog](https://www.keycloak.org/docs/latest/upgrading/#default-validations), they apply two validators by default;
`username-prohibited-characters`, and `up-username-not-idn-homograph`. See the documentation on these validators [here](https://www.keycloak.org/docs/26.0.7/server_admin/#built-in-validators).

I've copied these regex patterns from the Keycloak open source codebase, and combined them myself.
The sources can be found in [UsernameProhibitedCharactersValidator.java](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/userprofile/validator/UsernameProhibitedCharactersValidator.java#L44) and [Validation.java](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/services/validation/Validation.java#L37) respectively.

The exact list of supported and unsupported characters can be found [here](https://github.com/keycloak/keycloak/blob/main/services/src/test/java/org/keycloak/userprofile/validator/UsernameProhibitedCharactersValidatorTest.java).


richturner (changes requested):
Hey @MartinaeyNL have you tested this regex after unescaping and testing at:

https://www.regexplanet.com/advanced/java/index.html

I couldn't get a password with any symbols to match

richturner:
Any update on this?

MartinaeyNL:
Thanks for the heads-up! I haven't worked on this since the holidays, so I'll pick this up.

richturner:
Should we not have regex check on frontend also?

MartinaeyNL:
Thanks! I unfortunately didn't manage to get the complex regexp working in JavaScript,
so I manually added the special characters instead.

ui/app/manager/src/pages/page-users.ts
```diff
@@ -855,8 +857,8 @@ export class PageUsers extends Page<AppStateKeyed> {
                                   .label="${i18next.t("username")}"
                                   .type="${InputType.TEXT}" minLength="3" maxLength="255" 
                                   ?required="${isServiceUser || !this._registrationEmailAsUsername}"
-                                  pattern="[A-Za-z0-9\\-_]+"
-                                  .value="${user.username}" autocomplete="false"
+                                  pattern="[A-Za-z0-9\-_+@.ßçʊ]+"
+                                  .value="${user.username}"
```
richturner, line 861: Was the autocomplete attribute removed on purpose?

MartinaeyNL: Probably a result of merging with the latest changes in #1566.
So yes, the `autocomplete` attribute should be set to `false`, I'll commit that right away.

MartinaeyNL:
See my comments. Feel free to review again! 👍

richturner (approved):

PR #1551, opened 2024-12-18
UI fix: Gateway widget navigating to invalid URL

MartinaeyNL:
When a gateway tunnel is already active, and you press "Open" button on the `gateway-widget`, it fails to navigate.
This was because the unique identifier `id` from `GatewayTunnelInfo` was not used correctly.

The solution was quite simple; it now correctly utilizes the cached HTTP API response for constructing the URL.


MartinaeyNL:
Opened this in a draft for now, since we need to test this on a special gateway tunneling deployment.

richturner (approved):

PR #1554, opened 2024-12-19
Fix custom project CI/CD

wborn:
The Git submodules are no longer used so use the repository name to determine if the workflow runs in a custom project.


wborn:
I see the `is_main_repo` check is very similar. So I'll adjust that one and use it instead.

richturner:
FYI: There is a dependency in `deploy.sh` on the openremote submodule

richturner (approved):

PR #1556, opened 2024-12-19
Fix for manager app config to allow empty properties include field

### LINKED ISSUE #575: Manager Config in UI
First add Manager Config settings page with JSON text field
Next adjust to enable it to show in Form editor

### LINKED ISSUE #1544: Manager config is not not including properties in asset viewer layout
When applying panel layout specification to an asset type using the manager config, the properties used to be hidden from every field by including non:
```                 
"properties": {
   "include": []
},
```
However, the "include": [] is not stored/gets removed from the manager_config in the running system. Also when you try to add it using the UI.

This can be seen on staging.demo in e.g. the HarvestRobotAsset. In the repo the properties of the panels include: [], but in the running system this is not shown and each panel has the properties of the asset listed.

If you add e.g.:
```
"properties": {
   "include": [
      "hi"
   ]
},
```
Then only the non-existing "hi" property is shown, and the rest is not. So in that sense the include works.

richturner:
Fixes #1544

### DIFF
--- a/manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
+++ b/manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
@@ -19,7 +19,6 @@
  */
 package org.openremote.manager.app;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import jakarta.ws.rs.InternalServerErrorException;
 import jakarta.ws.rs.core.Response;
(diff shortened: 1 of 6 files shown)


manager/src/main/java/org/openremote/manager/app/ConfigurationResourceImpl.java
```diff
@@ -19,6 +19,7 @@
  */
 package org.openremote.manager.app;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
```
MartinaeyNL, line 22:
```suggestion
```

MartinaeyNL (approved):

Issue #1473, opened 2024-11-04
Data export: set max datapoints to 1000000


DonWillems:
If a user tries to export a lot of data at once, a big csv needs to be created on the disk which could lead to a too high disk utilization %. This should be limited (for now) until we find a more elegant way of handling data export. 

- Check the amount of datapoints when the user presses the export button. If its higher than 1.000.000 show an okdialog that says "The amount of datapoints selected ([NUMBER OF DATAPOINTS]) exceeds the limit of 1.000.000. Please adjust your selection."
- Add a limit to the API that gives the error if a larger amount than 1000000 is requested
- The 'export' button should be disabled if no attributes are selected. The API endpoint should decline the request if no attributes are added to the request.

While working on this, look into this item: #1396

Hackerberg43:
Related to the Data export feature, I saw this as the most recent issue topic so will put some additional suggestions here.
In addition to the performance issues you are working on, the actual output of the .csv file is a little hard to process.
Most people I know that analyse .csv files use excel with the text to columns command.
The current format makes that a little difficult and requires some conversion, because datapoints of different attributes are mixed up.
![image](https://github.com/user-attachments/assets/497145a6-913b-4e4d-9010-743fcfd6b51a)

And in flow csv viewer the raw output makes no sense at all:
![image](https://github.com/user-attachments/assets/f83c956e-8552-467b-ad6c-4d3b33758427)

I would propose to alter the output, such that the text-to-columns command would result in the following:
![image](https://github.com/user-attachments/assets/8535abd2-ebfe-446f-b6ee-c1e6cb024b93)

I am working on this now: https://github.com/Hackerberg43/openremote/tree/enhancement/dataExport
One downside is that my current solution requires an additional SQL package, so would result in a breaking change.

PR #1557, opened 2024-12-19
Enhancement/data export

Hackerberg43:
As mentioned in #1473 , an alternative formatting of the .csv data export output.

- Makes data easier to process further
- Reduces file size 

Note: uses the tablefunc sql extension, so would result in a breaking change.

From:
image

To:
image

Fixes https://github.com/openremote/vdl-es/issues/96


richturner:
Seems this would require an alteration to our postgres image.

I would like to understand is this really necessary. Excel has always been able to open CSV files using import data and specifying the delimiter.

richturner:
Can you explain a little about mixed? You mean you want it in a pivot table with a column per attribute?

Hackerberg43:
Yes, to my knowledge that is the most common format for csv timeseries

Hackerberg43:
https://github.com/openremote/openremote/issues/1473#issuecomment-2553056710

Hackerberg43:
@richturner I can make an issue which links to this branch, you can put it in the backlog and revisit when a postgres image update is required?

Then I will close this pr.

richturner:
> @richturner I can make an issue which links to this branch, you can put it in the backlog and revisit when a postgres image update is required?
> 
> Then I will close this pr.

That sounds great thanks. Not against having an alternative export format but I think it should be configurable in the export UI

PR #1558, opened 2024-12-20
Added limit options on datapoint CSV export

### LINKED ISSUE #1473: Data export: set max datapoints to 1000000

MartinaeyNL:
Closes #1473 

Added support for the `OR_DATA_POINTS_EXPORT_LIMIT` environment variable,
to refuse CSV exports above the threshold. By default, this limit is set to 1 million data points.

### DIFF
--- a/manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointService.java
+++ b/manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointService.java
@@ -267,7 +267,7 @@ public class AssetDatapointService extends AbstractDatapointService<AssetDatapoi
 
         return scheduledExecutorService.schedule(() -> {
             String fileName = UniqueIdentifierGenerator.generateId() + ".csv";
-            StringBuilder sb = new StringBuilder(String.format("copy ("))
+            StringBuilder sb = new StringBuilder("copy (")
                     .append(getSelectExportQuery(attributeRefs, fromTimestamp, toTimestamp))
                     .append(") to '/storage/")
                     .append(EXPORT_STORAGE_DIR_NAME)
--- a/manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointService.java
+++ b/manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointService.java
@@ -2,6 +2,7 @@ package org.openremote.manager.datapoint;
 
 import org.openremote.agent.protocol.ProtocolDatapointService;
 import org.openremote.container.timer.TimerService;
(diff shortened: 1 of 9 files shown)


MartinaeyNL:
Converted to draft as I'm working on test coverage for the PR.

MartinaeyNL:
Ready for review! 👍 

_(I won't assign a reviewer for now, because of the holidays)_

manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointService.java
```diff
@@ -233,20 +245,51 @@ protected String buildWhereClause(List<Pair<String, Attribute<?>>> attributes, b
...
         return scheduledExecutorService.schedule(() -> {
             String fileName = UniqueIdentifierGenerator.generateId() + ".csv";
-            StringBuilder sb = new StringBuilder(String.format("copy (select ad.timestamp, a.name, ad.attribute_name, value from asset_datapoint ad, asset a where ad.entity_id = a.id and ad.timestamp >= to_timestamp(%d) and ad.timestamp <= to_timestamp(%d) and (", fromTimestamp / 1000, toTimestamp / 1000))
-                .append(Arrays.stream(attributeRefs).map(attributeRef -> String.format("(ad.entity_id = '%s' and ad.attribute_name = '%s')", attributeRef.getId(), attributeRef.getName())).collect(Collectors.joining(" or ")))
-                .append(")) to '/storage/")
-                .append(EXPORT_STORAGE_DIR_NAME)
-                .append("/")
-                .append(fileName)
-                .append("' delimiter ',' CSV HEADER;");
+            StringBuilder sb = new StringBuilder(String.format("copy ("))
```
wborn, line 270: Redundant String formatting:

```suggestion
            StringBuilder sb = new StringBuilder("copy (")
```

wborn (changes requested):

wborn (approved):

PR #1565, opened 2024-12-23
Add password policy to account page

Hackerberg43:
- In continuation of #1524 , copied @MartinaeyNL his code and applied it to the account page aswell.
- Fixed that snackbar on account page didn't correctly display password update result.

Points of discussion:
- I've simply copied the _getPasswordPolicyTemplate function, so there is duplicate code now.
Might want to create a component wrapping the password change fields + policy ?

- While working on this, realized that there is no "current password" field. Is this on purpose ? Seems like a security risk regarding devices with an open session left unguarded. Wouldnt allowing only a forced password reset on the user page and requiring "current password" on the account page make sense?

Fixes https://github.com/openremote/vdl-es/issues/101

### DIFF
--- a/ui/app/manager/src/pages/page-account.ts
+++ b/ui/app/manager/src/pages/page-account.ts
@@ -7,10 +7,11 @@ import {Store} from "@reduxjs/toolkit";
 import {Page, PageProvider, AppStateKeyed} from "@openremote/or-app";
 import {when} from "lit/directives/when.js";
 import {until} from "lit/directives/until.js";
+import {map} from 'lit/directives/map.js';
 import {guard} from "lit/directives/guard.js";
 import {i18next} from "@openremote/or-translate";
 import {InputType, OrInputChangedEvent, OrMwcInput} from "@openremote/or-mwc-components/or-mwc-input";
-import {ClientRole, Role, User, UserAssetLink, UserQuery} from "@openremote/model";
+import {ClientRole, Credential, Role, User, UserAssetLink, UserQuery} from "@openremote/model";
 import {showSnackbar} from "@openremote/or-mwc-components/or-mwc-snackbar";
 
 export function pageAccountProvider(store: Store<AppStateKeyed>): PageProvider<AppStateKeyed> {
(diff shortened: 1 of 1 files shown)


DonWillems:
> * While working on this, realized that there is no "current password" field. Is this on purpose ? Seems like a security risk regarding devices with an open session left unguarded. Wouldnt allowing only a forced password reset on the user page and requiring "current password" on the account page make sense?

Yes that makes sense, I will add it to our backlog: https://github.com/openremote/openremote/issues/1570

ui/app/manager/src/pages/page-account.ts
```diff
@@ -180,6 +184,7 @@ export class PageAccount extends Page<AppStateKeyed>  {
         }
 
         const readonly = !manager.hasRole(ClientRole.WRITE_ADMIN);
+        this._getPasswordPolicy();
```
MartinaeyNL, line 187: `render()` is a Lit function that gets called often during every state update,
Expensive functions like this `_getPasswordPolicy()` HTTP request should happen only once.
I would place this in a `connectedCallback()` function, which gets called before initial page render.
```typescript
connectedCallback() {
    super.connectedCallback();
    this._getPasswordPolicy().then(policy => {
        this._passwordPolicy = policy;
    })
}
```

ui/app/manager/src/pages/page-account.ts
```diff
@@ -430,15 +437,83 @@ export class PageAccount extends Page<AppStateKeyed>  {
...
-        }).catch(e => {
-            console.error(e);
-            showSnackbar(undefined, "saveUserFailed");
-        });
+        })
     }
+
+    protected async _getPasswordPolicy(): Promise<void> {
+        const realmResponse = await manager.rest.api.RealmResource.get(manager.getRealm());
+        this._passwordPolicy = realmResponse.data.passwordPolicy;
```
MartinaeyNL, line 455: I would just return the async function here;
```typescript
return await manager.rest.api.RealmResource.get(manager.getRealm());
```
Catching errors, and updating local state, should be handled outside of this function.

MartinaeyNL (changes requested):

ui/app/manager/src/pages/page-account.ts
Hackerberg43: I've tried this but it resulted in a type mismatch I couldn't get around. Also, I found no other places where this is actually done in the code. So I kept the function in line with similar functions .. ?

MartinaeyNL (approved):
LGTM

PR #1575, opened 2025-01-08
Bug fix for or-util when used from a custom project

richturner:
This PR fixes an issue with trying to run or-util in a custom project when it is located in `node_modules`, it couldn't find the gradle wrapper directory so I have changed the logic to use the `process.cwd()` as a starting point and work up from there.


wborn (approved):
LGTM!

Issue #1449, opened 2024-10-18
Rules: When-Then: add operator 'not updated for' that checks for the latest update of an attribute


DonWillems:
As a user I want to monitor when attribute x last updated, so that I can take action if it did not update within the time I expected.

e.g.
As a user I want to receive a notification when my weight sensor did not send new measurements for 60 minutes AND the device connection status is true, so that I know something is wrong with the sensor.

- Add new mechanic to JSON rules: check for how long ago the last updated value was.
- Add controls to When-Then UI LHS: operator 'not update for' should be added to the operator list (after asset attribute selection). 
- When 'not updated for' is selected, no attribute value field will appear. Instead the duration field is added and needs to be filled in. (see #1448 for duration)

![Image](https://github.com/user-attachments/assets/c4019e22-8e1b-4b52-b0d2-a03c96a91d35)

PR #1583, opened 2025-01-10
Added: Rules When-Then operator 'not updated for' 

dominiquekleeven:
Adds the 'Not updated for' operator for When-Then attribute conditions #1449 

Extended the AttributePredicate class with timestampOlderThan as specified here https://github.com/openremote/openremote/issues/1449#issuecomment-2504351582 and extended the asAttributeMatcher to take the timestampOlderThan field into account when its set for the given predicate

### DIFF
--- a/manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
+++ b/manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
@@ -198,16 +198,21 @@ public class AssetQueryPredicate implements Predicate<AttributeInfo> {
         if (condition.getItems().size() > 0) {
 
             condition.getItems().stream()
-                .forEach(p -> {                 
+                .forEach(p -> {         
+                    AtomicReference<Predicate<AttributeInfo>> predicate = new AtomicReference<>();
+                    
+                    // timestampOlderThan is set then we use this as a predicate
                     if (p.timestampOlderThan != null) {
                         long durationMillis = TimeUtil.parseTimeDuration(p.timestampOlderThan);
-                        long cutoffTime = currentMillisProducer.get() - durationMillis;
-                        Predicate<AttributeInfo> timestampPredicate = assetState -> assetState.getTimestamp() < cutoffTime;
-                        attributePredicates.add(timestampPredicate);
+                        Predicate<AttributeInfo> timestampPredicate = assetState -> assetState.getTimestamp() < currentMillisProducer.get() - durationMillis;
+                        predicate.set(timestampPredicate);
                     } else {
+                        // otherwise we use the value predicate (regular predicate)
                         Predicate<AttributeInfo> attributePredicate = (Predicate<AttributeInfo>)(Predicate)asPredicate(currentMillisProducer, p);
-                        attributePredicates.add(attributePredicate);
+                        predicate.set(attributePredicate);
                     }
+
+                    attributePredicates.add(predicate.get());
 
                     AtomicReference<Predicate<AttributeInfo>> metaPredicate = new AtomicReference<>(nameValueHolder -> true);
                     AtomicReference<Predicate<AttributeInfo>> oldValuePredicate = new AtomicReference<>(value -> true);
--- a/manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
+++ b/manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
@@ -200,17 +200,17 @@ public class AssetQueryPredicate implements Predicate<AttributeInfo> {
             condition.getItems().stream()
                 .forEach(p -> {         
                     Predicate<AttributeInfo> wrappedPredicate = attributeInfo -> {
-                        Predicate<NameValueHolder<?>> basePredicates = asPredicate(currentMillisProducer, p);
-                        
-                        // Include older than timestamp check if timestampOlderThan is provided
+
+                        // Check timestamp not updated for condition
                         if (p.timestampOlderThan != null) {
                             long currentTime = currentMillisProducer.get();
                             long durationMillis = TimeUtil.parseTimeDuration(p.timestampOlderThan);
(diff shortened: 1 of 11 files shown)


dominiquekleeven:
Feature is functional can be set as an operator in When-Then rules, still need to add a test case for it

dominiquekleeven:
This issue https://github.com/openremote/openremote/issues/1589

Is blocking, since the Not updated for operator also needs to properly reassign durations if was previously set for the same predicate

dominiquekleeven:
> This issue #1589
> 
> Is blocking, since the Not updated for operator also needs to properly reassign durations if it was previously set for the same predicate.
> 
> Made a PR #1594 that fixes the blocking issue

No longer an issue as the fix has been merged.

This PR can be reviewed

DonWillems:
Looks good to me; tested on test2

MartinaeyNL (commented):
The UI code is OK with me. I'll leave the manager / model changes up to @richturner for review.

manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
```diff
@@ -197,9 +198,23 @@ public static Function<Collection<AttributeInfo>, Set<AttributeInfo>> asAttribut
...
-
+                .forEach(p -> {         
+                    Predicate<AttributeInfo> wrappedPredicate = attributeInfo -> {
+                        Predicate<NameValueHolder<?>> basePredicates = asPredicate(currentMillisProducer, p);
+                        
+                        // Include older than timestamp check if timestampOlderThan is provided
+                        if (p.timestampOlderThan != null) {
+                            long currentTime = currentMillisProducer.get();
+                            long durationMillis = TimeUtil.parseTimeDuration(p.timestampOlderThan);
+                            if (!(attributeInfo.getTimestamp() < currentTime - durationMillis)) {
```
richturner, line 209: Can change the logic here to avoid the negation

manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
```diff
@@ -197,9 +198,23 @@ public static Function<Collection<AttributeInfo>, Set<AttributeInfo>> asAttribut
...
+                                return false;
```
richturner, line 210: Not sure the whole predicate should be false based on this `timestampOlderThan` condition.

Shouldn't it be OR logic with the other predicate conditions? So if this `timestampOlderThan` is true then return true otherwise evaluate the rest of the predicate.

test/src/test/groovy/org/openremote/test/rules/residence/JsonRulesTest.groovy
```diff
@@ -693,6 +693,173 @@ class JsonRulesTest extends Specification implements ManagerContainerTrait {
...
+            assert realmBuildingEngine.facts.assetStates.size() == DEMO_RULE_STATES_SMART_BUILDING + 2 // 2 additional rule states
+            assert realmBuildingEngine.lastFireTimestamp == timerService.getNow().toEpochMilli()
+            lastFireTimestamp = realmBuildingEngine.lastFireTimestamp
+        }
+
+        when: "the light asset is updated with onOff set to true and brightness set to 100"
+        lightAsset = assetStorageService.find(lightId)
+        lightAsset.getAttribute("onOff").get().setValue(true)
+        lightAsset.getAttribute("brightness").get().setValue(100)
+        assetStorageService.merge(lightAsset)
```
richturner, line 760: Use asset processing service to simply udpate an attribute value

test/src/test/groovy/org/openremote/test/rules/residence/JsonRulesTest.groovy
```diff
@@ -693,6 +693,173 @@ class JsonRulesTest extends Specification implements ManagerContainerTrait {
...
+        }
+
+        when: "brightness is updated to 100"
+        lightAsset = assetStorageService.find(lightId)
+        lightAsset.getAttribute("brightness").get().setValue(100)
+        assetStorageService.merge(lightAsset)
+
+        then: "the light asset should have its brightness set to 100"
+        conditions.eventually {
+            def light = assetStorageService.find(lightId)
```
richturner, line 843: Maybe check `notes` to make sure it hasn't updated

richturner (changes requested):

manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
dominiquekleeven: Fixed

test/src/test/groovy/org/openremote/test/rules/residence/JsonRulesTest.groovy
dominiquekleeven: Using the processing service now, will use this method going forward worth noting theres plenty of tests that do the merge method

test/src/test/groovy/org/openremote/test/rules/residence/JsonRulesTest.groovy
dominiquekleeven: Also now checks if tests hasn't been updated

test/src/test/groovy/org/openremote/test/rules/residence/JsonRulesTest.groovy
richturner: Generally `assetStorageService.merge()` is for making structural changes to the asset rather than just sending attribute events. The backend does some magic to generate `AttributeEvent`(s) during asset merge for attributes whose value has changed so the changes can be detected by rules etc.

manager/src/main/java/org/openremote/manager/rules/AssetQueryPredicate.java
dominiquekleeven: I reached out on skype for some additional clarification here

richturner (approved):

PR #1584, opened 2025-01-11
Bugfix for hidden semicolon on `or-mwc-snackbar`

pankalog:
I was browsing the manager UI with a dark mode reader enabled, and after trying to move an asset, I saw this snackbar:

image

Everything looks good, other than the questionmark that's on the right side. Testing it without the dark mode reader, it's barely visible, so it makes sense that it wasn't noticed:
image

So this PR fixes that. Very small PR, but I saw it and I needed to fix it. 😆😆😆


MartinaeyNL (approved):
Thanks @pankalog !

Issue #1591, opened 2025-01-16
CI/CD to test deployment broken with latest version


MartinaeyNL:
Referring to these GitHub Actions runs;
- (deploying #1550)
https://github.com/openremote/openremote/actions/runs/12812276657/job/35723536941
- (deploying #1586)
https://github.com/openremote/openremote/actions/runs/12766924464

Tested both with the latest version of master. (as of January 16th 2025)
It looks like the docker build tag is missing?

MartinaeyNL:
For context: I deployed these manually using the CI/CD action in GitHub Actions of this repo.
See the image:

![Image](https://github.com/user-attachments/assets/127eec2f-16cf-4009-b8f4-8a755a5d5a68)

wborn:
I just had a brief look and it seems to be caused by #1530 which removed the `commitShaShort` variable that is used for the tag.

MartinaeyNL:
Unfortunately the CI/CD is still broken with the latest changes.
See https://github.com/openremote/openremote/actions/runs/12829809447,

You can also take a look at @wborn his comment in the linked PR:
> Unfortunately branch deployments still seem to fail, see: https://github.com/openremote/openremote/actions/runs/12826106713/job/35765376021
> 
> ```
>   Executing: grype -o sarif --fail-on critical openremote/manager:35b957b
>   1 error occurred:
>   	* failed to catalog: unable to load image: unable to use DockerDaemon source: pull failed: Error response from daemon: manifest for openremote/manager:35b957b not found: manifest unknown: manifest unknown
> ```
> 
> CC: @ebariaux

PR #1594, opened 2025-01-17
Fix duration reassignment for When-Then rules conditions

### LINKED ISSUE #1589: UI When-Then: removing a rule condition's attribute duration causes incorrect duration reassignment when the first condition has no duration set
**Issue**
Removing an attribute predicate's duration causes an incorrect duration reassignment when the first attribute predicate has no duration set. This does **not** occur if the first attribute predicate has a duration condition set.

**Screenshots:**
Before deleting the last attribute predicate with a duration:
image

After deleting the predicate
image

Note that the duration is removed, but the order of the durations map is now incorrect after deleting the third attribute predicate.

dominiquekleeven:
Fixes #1589 

Also fixes another issue where if you (example) delete the first attribute predicate and it had no duration that it would also incorrectly re-assign the durations. This was caused by the duration.has(index) so it would not reassign if the predicate deleted had no duration set even though it would change the order.

### DIFF
--- a/ui/component/or-rules/src/json-viewer/or-rule-asset-query.ts
+++ b/ui/component/or-rules/src/json-viewer/or-rule-asset-query.ts
@@ -824,12 +824,16 @@ export class OrRuleAssetQuery extends translate(i18next)(LitElement) {
         if (index >= 0) {
             group.items!.splice(index, 1);
 
-            if(this.duration.has(index)) {
-                const newEntries = Array.from(this.duration.entries())
-                    .filter(([k, v]) => k !== index)
-                    .map(([k, v], i) => [i, v] as [number, string | undefined]);
+            if (this.duration) {
+                // re-assign durations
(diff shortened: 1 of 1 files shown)


MartinaeyNL:
Thanks for the fix!

Code looks fine to me. Did you deploy this somewhere for us to test this?
Maybe @DonWillems wants to take a look?

dominiquekleeven:
Tried to deploy it to test2 but deployment seems to be broken currently, see https://github.com/openremote/openremote/issues/1591

dominiquekleeven:
Branch is deployed to test2 and can be tested

MartinaeyNL:
Looks OK to me, also on the live deployment.
I'll ask @DonWillems to review the UI as well. If he has no further comments, I will approve the PR and merge it.

MartinaeyNL (approved):

Issue #1572, opened 2025-01-08
Notification management API integration


mihaelaaleks:
### Description
The notification table is not properly fetching data from the main API endpoint. Currently using temporary development endpoints as workaround.

### Current Behavior
- Using getAllNotifications temporary endpoint
- Time range filter may have incorrect end date (both using fromDate)

### Expected Behavior
- Successfully fetch notifications using main getNotifications endpoint
- Proper filtering by time range
- Respect realm-based access control

### Technical Notes
- Check error handling in NotificationService
- Verify API contract matches implementation
- Add proper logging for debugging API calls
- Consider adding API response type validation

### Acceptance Criteria
- [x] Main getNotifications endpoint successfully returns data
- [x] Correct time range filtering
- [x] Proper error handling and user feedback
- [x] Logging for debugging

PR #1615, opened 2025-01-21
Feature/notification management

### LINKED ISSUE #1601: Notifications: UI/UX improvements
- [ ] 1. Give table same padding/margins as Alarms table
- [ ] 2. Give header and interactive components same spacing as Alarms header
- [ ] 3. Remove the icon from the header and place the item in the secondary menu (in the top right menu)
- [ ] 4. Remove the white background
- [ ] 5. Do not make the time filtering a toggle, just always filter on the last day, the user can extend the period manually.
- [ ] 6. make the source filter wider, and have "All sources" selected by default. Translate the items so they are not all caps
- [ ] 7. Move the target column between status and sent on
- [ ] 8. Give status, sent on, delivered on columns less space like on the Alarms page
- [ ] 9. Change "+ ADD" to "+ SEND NEW"
- [ ] 11. (Skipped 10 for whatever reason) the targets in this column should be assets (asset name derived from the ID, ideally with a link to the asset), users (name derived from ID, ideally with link to the user)

Image

- [ ] 1. Make these titles the same as on e.g. a user page, h5
- [ ] 2. Extend the items to fill the space
- [ ] 3. Give it the same with as the other inputs, translate the contents so its not all caps
- [ ] 4. Translate the items and add info like on the alarms page: "Realm rule (ID:2342)"
- [ ] 5. If you open an existing notification, make it so that it closes if you click outside of the dialog (but do not close it if you are creating a new notification)
- [ ] 6. Notifications should be sent to the linked users of the selected asset. This mechanism can be seen in the rules. (its a bit extended their with realm roles, but ignore that bit for now)

Image

- [ ] 1. If the type is an asset, show the asset tree selector, maybe it has to be a button then with "4 assets" like it is in other places. The assets shown should be filtered by if they have a linked user.
- [ ] 2. Translate this list

Image

mihaelaaleks:
This PR addresses the following issues: 
1. #1601 
2. #1572


DonWillems:
Hi @mihaelaaleks , do you have a status update?

mihaelaaleks:
@DonWillems  Apologies for the overdue update 😅 I have some last tests to add and I'll push my changes for review by tomorrow.

DonWillems:
Hi Michaela, it took a while for me to find the time, but here are some points that stood out to me:

### Bugs / Incorrect Behavior
- [ ] As you mentioned the deliveredOn and acknowledgedOn timestamps aren't set correctly by the apps. I created a separate bug for this: #1819. If I use the endpoints manually, the table does update the status correctly.  
- [ ] When creating a notification but not filling in all required fields (like title and content) you can press the create button, but you will get an error. The Create button should only be enabled if you filled in all the required fields.  
- [ ] When creating a notification with an asset as target that does not have a linked user, the notification correctly can't be sent. However, the notification says: "Failed to create notification:" with no explanation.  
- [x] `realm` field stays invalid even if a realm is set:  
  Target field stays invalid  
- [ ] Every time I tab back to the table page, it's recollecting all the rows?  
- [ ] Something weird is going on with sorting on sent date time:  
  image

### UX/UI Improvements
- [x] Asset tree is awkward in the dialog, as it doesn't fit with many assets (and otherwise looks off too).  
  I understand why you chose this over opening it as a dialog. I think we should do it like it is in rules: first select asset type, then a (multi)select of the assets of that type. With at the top of that list the option to select 'All of this type'.  
  assettreeInDialog  
- [x] Target is clickable, underline the target so that it is clear that it is individually clickable as opposed to the click anywhere else on the row.  
- [ ] When a filter does not return any results it should state so in a message in the table.

### Functionality Enhancements
- [ ] Service users shouldn't be shown in the users list  
- [x] You can't multiselect users, which will make sending notifications tedious. Add functionality to add multiple users.
- [ ] If a notification is sent to user, it still shows as target: asset. I assume there is no way to read this from the notification in the db. I think we should always show the linked user(s) of the target asset to give the needed information.  
  The downside is that if the linked user is changed, the info won't be correct historically. But let's accept that for now.  
- [ ] The rule id that sent the notification is not available? Should we add that?  
- [ ] What does priority do when creating a notification?

### Performance
- [ ] Why are you doing a call per row? Is it to get the name and icon of the target asset? Find out if there is a better way, if there isn't, maybe we need to leave the name out?  
- [ ] Remove the console messages

### Nice to Haves
- [ ] Add a search that matches anything in the 'title', asset id, and 'linked user'. Filter out the rows that match  
- [ ] Is there a way to combine notifications that are sent from one rule trigger/from one manual action? If it is sent to 100 consoles at once, you will have 100 entries. Would be nice to have that as 1 entry; or an entry per status.
- [ ] Ideally you would also want to be able to select all users that have a certain realm role or role.  


Turned out to be quite a list. If there is anything unclear, feel free to give me a call. Also, if you need some help from the developers don't hesitate to contact them.

DonWillems:
- [ ] Replace DeliveredOn column with Linked Users column. In practice the table only shows the assets the messages are sent to (needs triple checking). This means the Target is always an asset, even if the user or realm is targeted in the message creation. The assets (consoles) a message can be sent to always have a linked user. It would be helpful to also see that linked user immediately.
- [ ] Make the target field in the dialog of an existing form nicer by using the same layout as in the table cell. However, similar as above, probably all existing notifications will only have assets as target. So consider to replace "Target Type" and "Target" input fields with the icon+text combination for the asset and the user. Maybe show only the user? Especially if all Console Assets are hidden (when this is applied).
- [ ] Should 'Source' have a column since you can filter by it? (remove asset column if we have users column)
- [ ] Clean up the dev-only endpoints. Keep them in commented out for future development?

MartinaeyNL:
I'm planning to pick this up, so we can get this feature merged 💪

wborn:
@codex review

Ekhorn:
There are quite some comments in this PR. Would like to collect all open issues in this PR and create a new clean PR.

MartinaeyNL:
As mentioned above, see pull request #2673 for the continuation of this feature. 👍

Issue #830, opened 2022-10-10
Assets: Add 'Required attribute values' to or-add-asset-dialog


DonWillems:
**Maybe we do not want this at all. It could be annoying to have to fill in these values on creation. It might be valuable to see which attributes are a must for certain agents, but we could also do that by using the manager config and creating a panel that holds those attributes. For the GroupAsset, we might also want to let this requirement completely go. If you have all children of the same asset type, some additional functionality becomes available, otherwise you just won't see the table. Instead that panel shows a message that it can't show info because the children aren't all the same**

As a user I want to set the required attribute values of an asset, so that the asset is created with the required attributes filled in. 

This makes sense for certain agents that will never work if certain attributes are not filled in. Currently only the childAssetType attribute of the GroupAsset asset type is a 'Required attribute value' (this concept does not exist yet).

Add the meta item (configuration item) 'Required attribute value'. If this is applied to an attribute. That attribute needs a value when creating an asset.

1. Add a select to the add asset dialog only for the GroupAsset asset type to set the attribute value of 'childAssetType'. You should be able to select the assettypes that are also shown in the column on the left. The user can only add the GroupAsset if this value is set. The ADD button should be disabled if GroupAsset is selected and no value is set.
2. Regardless of view or modify mode, a new asset is immediately created and selected in the tree when pressing ADD (as is now when you are in view mode)
3. Regardless of view or modify mode, a copied asset is immediately created and selected in the tree when pressing the copy button (as is now when in view mode)

DonWillems:
The "required attribute value" is currently only used in the GroupAsset. A workaround has recently been introduced to be able to create a GroupAsset in view mode (because it was confusing you could only create it in Modify mode). #1989

The "required attribute value" could be nice for certain agents, but is not a must. We'll close this for now until a better flow is designed.

Issue #385, opened 2021-04-01
Assets: Add Asset: required attribute values


pierrekil:
As a user I want to add an asset of any assettype so that I can create the project I need.

As a user I want to fill in attribute values required for an asset to be created, so that it is created correctly in my project.

Create section in add asset modal that shows the attribute inputs for attributes values that are required for creating the asset. 

Verify this functionality by adding an asset of type Group Asset

Design of add asset modal:
https://docs.google.com/document/d/1LZOKCIng-_mssSwULkTBFgeaR5Okjo6I1gM3VEAqHvE/edit#heading=h.ocgrfxilx1xd

PR #1629, opened 2025-01-27
bugfix/groupasset creation

Hackerberg43:
## Description
This PR fixes the problem of being unable to create a groupAsset via the regular way when not in edit asset mode.
Altered the validation process such that childAssetType is set to an empty string when it isn't set yet.

- Now you can create the groupAsset, but can't drop in assets until you change the childAssetType to a valid assettype.
- You can only change the childAssetType when it's still empty.

## Consideration: More specific snackbar messages
The childAssetType validation has quite some different error cases that occur when the user is handling it the wrong way.
Something to consider here, is that the snackbar very often shows general messages like "An error has occurred" or "Failed to create".   The backend does actually provide specific information about the problem and also seems to pass it through the api exception handler.
Providing the user feedback about what he/she is actually doing wrong, would help a lot. I understand that this would require a lot more effort in the translation files, but for the acceptance factor of a new user, this would be very beneficial.
As a compromise, maybe it would help to add the exceptions to the snackbar in a small/italics font, not too obtrusive but always giving the user the possibility to google translate the error ?
## Checklist

- [ ] Try to create and alter some groupassets, move assets into it etc. to verify


richturner:
I think it should be the responsibility of the frontend to ensure a valid GroupAsset is submitted to the backend rather than hacking it through the backend.

I was never that onboard with assets being auto saved when added and not in modify mode. See assets like HTTP Agent which get created but just generate an error as there is no valid host etc.

UI should require users to provide input for any required attributes.

Hackerberg43:
While I understand you from a technical point of view, keeping bugs like this in place for a long time, because it takes quite some effort to fix it the right way, is not the best either. Don't underestimate the impact these kinds of  bugs have upon users deciding whether to keep using the product. But that's just my 2cts.

richturner:
Definitely agree bugs/quirks are not great for our users and the product in general.

@DonWillems maybe we can create an issue for a bigger item of work to ensure new assets meet their schema requirements before they are sent to the backend, so the new asset dialog would also take care of configuring the attributes, what is there currently was a bit of a hack some time ago.

Maybe this solution can be integrated for now with some TODO comments in the code referencing the new issue?

DonWillems:
We thought about this before #385 #830 (doing this on mobile so not sure these links are working)
Also from ~4 years ago: 

Mark certain attributes as 'value required on creation' and show those in the add asset dialog.

richturner (approved):

PR #1638, opened 2025-01-31
Hotfix for users page not functioning

### LINKED ISSUE #1636: User page broken; cannot save user
Since the changes in #1566 and #1550, errors have appeared on the user page.
The user cannot be saved after changing the password.

I've already identified the cause of these;
- [x] After the ID change from `id="password"` to `id="new-password"` all query selectors have been broken. Refactor these.
- [x] The pattern attribute on the username field seems invalid for HTML. Please correct this.

MartinaeyNL:
Fixes #1636 

Changelog:
- Hotfix for repeat password verification not working (both user- and account page), causing the save button to break.
- Corrected the username input pattern, which was an invalid expression.

### DIFF
--- a/ui/app/manager/src/pages/page-account.ts
+++ b/ui/app/manager/src/pages/page-account.ts
@@ -152,10 +152,10 @@ export class PageAccount extends Page<AppStateKeyed>  {
     @state()
     protected _passwordPolicy: string[] = [];
 
-    @query("#password")
+    @query("#new-password")
     protected _passwordElem?: OrMwcInput;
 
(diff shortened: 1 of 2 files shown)


DonWillems:
@MartinaeyNL are there also changes needed to the account page related to the new-password field?

MartinaeyNL:
> @MartinaeyNL are there also changes needed to the account page related to the new-password field?

Yes you are correct. Just committed a fix for it. 👍

wborn (approved):

PR #1641, opened 2025-02-01
Bugfix for exception on the logs page

### LINKED ISSUE #1637: Exceptions on Logs page without logs
When there are no logs on the "Logs" page an exeception occurs:

```
bundle.0a4d762718a5f6fe460f.js:698 Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL
    at t.value (bundle.0a4d762718a5f6fe460f.js:698:14097)
    at t.<anonymous> (bundle.0a4d762718a5f6fe460f.js:698:13903)
    at f (bundle.0a4d762718a5f6fe460f.js:8:2089975)
    at Generator.<anonymous> (bundle.0a4d762718a5f6fe460f.js:8:2091321)
    at Generator.next (bundle.0a4d762718a5f6fe460f.js:8:2090404)
    at a (bundle.0a4d762718a5f6fe460f.js:698:945)
```

It then creates an endless number of requests for fetching more logs:

Image

dominiquekleeven:
## Description
  This fixes #1637 
  
I've added an undefined check for the lastLink.uri property which fixes the issue and the exception that was occurring due to an undefined value.

Edit: I also changed the function order to make sure this._data is set before calling _getPageCount() because if the latter fails it would cause shouldUpdate() to be called again which would end up in a recursive call because this._data was still undefined.

## Checklist

### DIFF
--- a/ui/component/or-log-viewer/src/index.ts
+++ b/ui/component/or-log-viewer/src/index.ts
@@ -486,7 +486,7 @@ export class OrLogViewer extends translate(i18next)(LitElement) {
         if (linkHeaders) {
             const links = linkParser.parse(linkHeaders);
             const lastLink = links.rel("last");
-            if (lastLink) {
+            if (lastLink && lastLink.uri) {
                 const url = new URL(lastLink.uri);
                 return Util.getQueryParameters(url.search)['page'] as number;
             }
--- a/ui/component/or-log-viewer/src/index.ts
+++ b/ui/component/or-log-viewer/src/index.ts
@@ -476,8 +476,8 @@ export class OrLogViewer extends translate(i18next)(LitElement) {
 
         if (response.status === 200) {
             // Get page count
-            this._pageCount = this._getPageCount(response);
             this._data = response.data;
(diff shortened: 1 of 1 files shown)


MartinaeyNL:
Can you explain how this solves the recursion? It's been confusing to me why the function gets called so often.

(I'm not able to look in the codebase right now, so that's why I'm asking)

I'll try to find the time to verify this fix locally and/or on an deployment.

dominiquekleeven:
> Can you explain how this solves the recursion? It's been confusing to me why the function gets called so often.
> 
> (I'm not able to look in the codebase right now, so that's why I'm asking)
> 
> I'll try to find the time to verify this fix locally and/or on an deployment.

For some reason when the call to `_getPageCount()` fails it causes the `shouldUpdate` cycle to be triggered again, because `this_data` is still unset it'll call the `_loadData()` function again.

If we change the function order so `this._data = response.data` is called before ` this._pageCount = this._getPageCount(response);` also fixes the recursion issue.

richturner (approved):

PR #1643, opened 2025-02-03
Store outdated attribute events in the asset_datapoint table

### LINKED ISSUE #1642: Outdated AttributeEvents should still be committed to the asset_datapoint table to record historical data
## User Story
As a device can send attribute updates out of order and also the multithreaded nature of the backend event processing, any attribute event that is older than the current attribute value should still be committed to the historical data (`asset_datapoint`) table so the values can still be queried.

richturner:
Resolves #1642


dominiquekleeven (approved):
LGTM

PR #1644, opened 2025-02-03
UI update: Fix link header parsing for log viewer

richturner:
## Description
The link header is incorrectly parsed by the log viewer so the page count is not retrieved.


dominiquekleeven (approved):
Looks good to me, checked locally and it does now properly get the additional pages + show the pagination 👍

PR #1661, opened 2025-02-10
Fix for Asset query when DB contains meta set to JSON 'null'

### LINKED ISSUE #1660: Asset queries can fail on backend when one or more attributes have meta set to JSON null value

richturner:
Fixes #1660


wborn (approved):
Thanks for the fix!

PR #1667, opened 2025-02-13
Fix manager docker version issue with CI/CD deployment in custom projects

### LINKED ISSUE #1666: CI/CD Github action doesn't use the correct manager docker image
Currently CI/CD is outputting `MANAGER_VERSION=""` in the `temp/env` file on the host which then means the stack uses `latest` image.

The `openremoteVersion` specified in the `gradle.properties` file should be used as the version of the manager docker image to use during deployment as that is what the code is compiled against.

If `openremoteVersion` ends with `-SNAPSHOT` then CI/CD should fail with output `SNAPSHOT openremote dependencies not currently supported for custom project deployment`

richturner:
Fixes #1666

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -1,6 +1,6 @@
 # ---------------------------------------------------------------------------------------------------------------------
 # This workflow handles all CI/CD related tasks and can be re-used in custom projects
-# (https://github.com/openremote/custom-project i.e. has openremote submodule) or forks of this repo, it handles:
+# (https://github.com/openremote/custom-project) or forks of this repo, it handles:
 #
 # - Running tests on push/release
 # - Distributing openremote/manager images to docker hub (DOCKERHUB_USER and DOCKERHUB_PASSWORD must be set) on push/release
@@ -182,13 +182,11 @@ jobs:
         with:
           files_ignore: |
             .ci_cd/**
-            console/**
             deployment/**
-            setup/**
             ui/**
             **/*.md
             **/*.yml
-            manager/src/web/**
+            yarn.lock
 
       # Check which files have changed to only run appropriate tests and checks
       - name: UI files changed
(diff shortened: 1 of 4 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -213,32 +210,28 @@ jobs:
...
-              echo "ref=master" >> $GITHUB_OUTPUT
-            else
-              echo "ref=$version" >> $GITHUB_OUTPUT
+
+            if [ -z "$version" ]; then
+              echo "openremoteVersion must be set in gradle.properties"
+              exit 1
             fi
+
+            echo "ref=$version" >> $GITHUB_OUTPUT
```
wborn, line 234: The master value was used so you can more easily use/test snapshot versions in a custom project. That's no longer possible with these changes.

richturner: We don't build snapshot versions of docker containers

.github/workflows/ci_cd.yml
```diff
@@ -415,13 +322,18 @@ jobs:
               if 'npm' in eventConfig['distribute']:
                  npmPublishTag = eventConfig['distribute']['npm']
 
-          if isMainRepo and ((eventName == "release" and isRelease) or (eventName != "release" and not isRelease)):
+          if isMainRepo and (isRelease or (eventName != "release" and not isRelease)):
```
wborn, line 325: This construct was there so it only tries to release on tags and not on the master branch. It could try to re-release for instance if you deploy a test instance just after creating a release. If there are no commits after just creating a release the release plugin resolves the version to a release and not a snapshot version.

richturner: Thanks for the heads up.

The interaction with the versioning plugin and github actions seems a little confusing, I'll reinstate this for now.

richturner:
@wborn some info:

* Removed support for `managerDockerPush` in `ci_cd.json` which was intended to auto update custom projects when ta new manager docker image was available, wasn't used and doesn't make much sense now submodule is removed
* Added `CLEAN_INSTALL=true` to env for demo and staging.demo deployment (this was missing)

wborn (approved):

wborn:
@richturner the CI/CD workflow now fails with this changes, see: https://github.com/openremote/openremote/actions/runs/13367671078/job/37328995588

PR #1672, opened 2025-02-16
Don't allow recursive AssetQuery to retrieve `ACCESS_PUBLIC_READ` assets

### LINKED ISSUE #1671: Unauthenticated users are able to query for any asset
**Describe the bug**
An unauthenticated use is able to query for any asset. This is inconsistent with the documented behavior:
 ```
     /**
     * Retrieve assets using an {@link AssetQuery}.
     * <p>
     * If the authenticated user is the superuser then assets referenced in the query or returned by the query can be in
     * any realm. Otherwise assets must be in the same realm as the authenticated user, and for a restricted user, the
     * assets must be linked to the user. An empty result is returned if the user does not have access to the assets.
     * What is populated on the returned assets is determined by the
     * {@link AssetQuery#select} value.
     */
```
**To Reproduce**
1. Have a realm with some assets
2. perform a recursive query without the Authorization header and all assets are returned
```
curl -X 'POST' \
  'https://<openremote-endpoint>/api/master/asset/query' \
  -H 'Content-Type: application/json' \
  -H 'accept: application/json' \
  -d '{"recursive":true}' 
```

**Expected behavior**
The request should be denied with Unauthorized if no Authorization header is present

**Desktop (please complete the following information):**
 - OS: OSX 
 - Browser: N/A

pankalog:
fixes #1671

### DIFF
--- a/manager/src/main/java/org/openremote/manager/asset/AssetStorageService.java
+++ b/manager/src/main/java/org/openremote/manager/asset/AssetStorageService.java
@@ -1674,7 +1674,7 @@ public class AssetStorageService extends RouteBuilder implements ContainerServic
                 binders.add((em, st) -> st.setParameter(pos, query.userIds));
             }
 
-            if (level == 1 && query.access == Access.PUBLIC) {
+            if (query.access == Access.PUBLIC) {
                 sb.append(" and A.ACCESS_PUBLIC_READ is true");
             }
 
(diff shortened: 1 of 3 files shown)


MartinaeyNL:
@pankalog I would add a test that verifies the issue is resolved, and won't appear again in the future. 😉

richturner (approved):

PR #1681, opened 2025-02-18
Added test environment file

MartinaeyNL:
Added test environment file for deployments that should not restart.
It's quite common we share "reproducible problems" on testing servers, but turn out to be gone the next day.

The goal is to support a `test` configuration for the CI/CD GitHub action, that we'd use for test deployments.


MartinaeyNL:
I assume that environment files override values in the `.env` file, so these changes are enough.
Feel free to review and modify where necessary @richturner

richturner (approved):

PR #1700, opened 2025-02-20
Feature/move or attribute history to echarts

Hackerberg43:
## Description
Ok, so started with the or-attribute-history for migration to echarts, because its functionality is more basic then or-chart.
Functionality replication is complete, with the addition of a crosshair, zoom, image download and bar view.
afbeelding

In theory this is merge ready, however I am still working on some additional functionality

## Design Considerations:

- [ ] I have not used the this.timeUnit and period for xAxis formatting. Echarts default setting handles this pretty well, and with modification it seems hard to get full control over the axis ticks. Result is that sometimes, the labels overlap a little bit. Please let me know what you think about this.

## To Do:

- [ ] Add datazoom eventlistener, which queries the database again. Zooming in on downsampled data is pretty useless. Have this working in an isolated environment so should not be too much work.

- [ ] Add eventlistener for restore button. Currently, pressing restore resets the graph to the default timeframe, but the timeframe selector is not updated. Restore button is handy because you can easily get lost with zooming in.


Hackerberg43:
@MartinaeyNL

Hackerberg43:
Hi Don, thanks for your feedback, I will look at it. Regarding applying it to insights, I've been working on that. First version is almost ready.

MartinaeyNL:
Hi Wouter! Thanks for working on this!

I have no problems with migrating `or-attribute-history` to ECharts, as it's probably a better library for it as well.
Please make sure that the API is exactly the same as before, especially with those CSS variables it consumes.
I don't know what impact this change has on performance, as querying datapoints could be quite heavy.

Moving `or-chart` from Chart.js to ECharts however is more tricky.
There are custom projects that consume this component, and inherit on the functions it provides.
Same applies to CSS modifications that have been made there. (might need a major version change at that point)
So I'd like to prevent a breaking change there haha

For some time I have a long-term strategy in mind that I'd love to share.
Maybe not applicable to this pull request, but might be useful to take into consideration;

My ideal scenario here is to move ECharts to a separate component that is decoupled from any OpenRemote model.
So the exposed HTML attributes would be X- and Y-coordinates, and no `Asset[]` or `AssetAttribute[]` inputs for example.
(this is currently one of my pain points when using `or-chart` without the use of assets)
Whereafter you can consume the new ECharts component in both `or-attribute-history` and `chart-widget`.
The "querying datapoints logic" will then be handled by these consumers, instead of the ECharts component itself.
Later on, we can migrate all custom projects away from `or-chart`, and use this instead.

Hackerberg43:
Hey Martin, thanks for your perspective. Too bad the or-chart migration is already done for 95% https://github.com/Hackerberg43/openremote/tree/feature/move-or-chart-to-echarts. We can ofcourse make it live next to the existing or-chart & create a separate widget, but changing the way it consumes data now would be quite some additional work. However, the data consumption approach isn't really different from how it is done now, so if you have a new approach in mind for the current or-chart, it will probably transfer to the new variant quite well.

MartinaeyNL:
Hi Wouter, sorry for the late response.
That is no problem, simply continue with the work you're doing for `or-attribute-history` and `or-chart`. 👍 
We have versioning on the published NPM packages, so upgrading charts in custom projects is a manual action anyway.
There also won't be too many breaking changes I suppose.

I'll try to find some time, hopefully later this week, to review and locally test this.
Thanks for the work!

Hackerberg43:
> 
>    * [ ]  If there is not much data, the bottom bar doesn't line up nicely with the chart:
>     
This is a known ECharts bug : https://github.com/apache/echarts/issues/20374
  
  
 >
 >  * [ ]  While sliding the bottom bar with only one datapoint (~now) this error is shown (while testing locally):
 >
Thanks, fixed.


I've re-evaluated this code based on the new insights gained from or-chart.

Ready for review 👍

DonWillems:
- Locally, when I view the history chart of an asset and resize the screen or when I switch to a different asset, I get this message:
```
index.ts:520 Uncaught TypeError: Cannot read properties of undefined (reading 'resize')
    at ResizeObserver.<anonymous> (index.ts:520:78)
```

- There seems to be quite some white space on the right side of the chart (also a bit on the left). Can you reduce this?
image
- The 'now' indicator does feel a bit redundant here indeed. It also doesn't update:
image
- When you are zoomed in, you see your range in the bottom bar. If you then switch to a different timeframe, it keeps that zoom range and location for the new timeframe. That doesn't make much sense, so I think the zoom should reset when you switch timeframe.
- I think the zoom reset function that is in the echarts toolbox would be nice to have at the top
- Nice to have: When you save the image, can you give it a sensible name (date_assetName_attributeName) or something.

Hackerberg43:
>     * Locally, when I view the history chart of an asset and resize the screen or when I switch to a different asset, I get this message: index.ts:520 Uncaught TypeError: Cannot read properties of undefined (reading 'resize')
>     at ResizeObserver.<anonymous> (index.ts:520:78)
> 
I cannot reproduce this, it is pretty interesting. It indicates that the chart element does not exist, but it is created literally 5 lines before this. Also, one line before 520 the element is also called but that doesn't throw the error.  I've changed it to truthy, but now there's a change that the observers wont get attached.
Are you using the default assetviewer template here?
> 
>     * There seems to be quite some white space on the right side of the chart (also a bit on the left). Can you reduce this?
>       image
> 
I have enhanced it a bit now:  with containLabel feature, but that only applies to y axis labels, for the x-axis, the min and max labels aren't always fully visible. Removed min and max labels, less informative but a little more appealing.
>     * The 'now' indicator does feel a bit redundant here indeed. It also doesn't update:
>       image
> 
Removed.
>     * When you are zoomed in, you see your range in the bottom bar. If you then switch to a different timeframe, it keeps that zoom range and location for the new timeframe. That doesn't make much sense, so I think the zoom should reset when you switch timeframe.
> 
Good point, done
>     * I think the zoom reset function that is in the echarts toolbox would be nice to have at the top
The echarts restore button not only resets the zoom, but also the data to the first state after creation, which means it takes the initial timeframe, not the timeframe it is currently set to.
> 
>     * Nice to have: When you save the image, can you give it a sensible name (date_assetName_attributeName) or something.
Good point, done

DonWillems:
Thanks for the changes!

- I can't reproduce the first point anymore either.
- I've noticed you left in a console log message 'onZoomChange' when you change the timeframe.
- You went a bit over the top with the naming I think :p `History 4w1CElB2Olsn17uoKA1ktn flowPerMeter Thu Apr 10 2025 10_00_00 GMT+0200 (Central European Summer Time) - Fri Apr 11 2025 10_00_00 GMT+0200 (Central European Summer Time).png`, change to something like: `History 4w1CElB2Olsn17uoKA1ktn flowPerMeter 10-04-25 10_00_00 - 10-04-25 10_00_00.png`

Hackerberg43:
fixed

ui/component/or-chart/package.json
```diff
@@ -30,6 +30,7 @@
     "chart.js": "^3.6.0",
     "chartjs-adapter-moment": "^1.0.0",
     "chartjs-plugin-annotation": "^1.1.0",
+    "echarts": "^5.6.0",
```
MartinaeyNL: This change should be in `or-attribute-history`, not in `or-chart`.

yarn.lock
```diff
@@ -2618,6 +2618,7 @@ __metadata:
     chart.js: "npm:^3.6.0"
     chartjs-adapter-moment: "npm:^1.0.0"
     chartjs-plugin-annotation: "npm:^1.1.0"
```
MartinaeyNL: When the ECharts dependency is added to `or-attribute-history`, we should remove the ChartJS related dependencies.

ui/component/or-attribute-history/src/index.ts
```diff
@@ -247,25 +252,32 @@ export class OrAttributeHistory extends translate(i18next)(LitElement) {
     @property()
     protected _loading: boolean = false;
 
+    @property()
+    protected _zoomChanged: boolean = false;
+
+    @property()
+    protected _zoomReset: boolean = false;
```
MartinaeyNL: Double check; do these `@property` tags need to be here?
Because if it doesn't need a UI render / lifecycle event, we can leave them as simple variables.

Instead of `@property` you can also use `@state()`, so it isn't exposed as an HTML attribute. 😉

DonWillems:
Need to do some more testing on an instance with more data. But this stood out to me when testing locally:

https://github.com/user-attachments/assets/f170a849-5bab-409b-937d-411cad6a9a28

When first selecting an attribute in the history panel, the datapoint only shows when I start hovering over the chart. If I then use the double arrow to go to 'now', it shows up without needing to hover.

Hackerberg43:
fixed

DonWillems:
Did some more testing, looks good to me. With more datapoints the issue I reported above does not occur anymore. I think its because there is no line to draw with only one datapoint. So lets ignore it for now. @MartinaeyNL could you do a code review?

Hackerberg43:
It determines whether to show the circles based on the amount of datapoints, but this was set only after an update, during first initialization it was set to false, thats why you saw that behavior.

ui/component/or-attribute-history/src/index.ts
```diff
@@ -381,90 +396,136 @@ export class OrAttributeHistory extends translate(i18next)(LitElement) {
...
                                 },
-                                gridLines: {
-                                    color: "#cccccc"
+                                handleLabel: {
+                                    show: true
                                 }
+                            },
+                            handleLabel: {
+                                show: false
                             }
```
MartinaeyNL: I tested this branch out locally to see what it looked like.
When I was dragging the slider left/right, I was quite annoyed by the tooltip showing time.
I don't think it's necessary, especially since it's cut off at the left- and right side of the chart.

There's an option in ECharts config for this;
```suggestion
                            },
                            showDetail: false
```

ui/component/or-attribute-history/src/index.ts
```diff
@@ -381,90 +396,136 @@ export class OrAttributeHistory extends translate(i18next)(LitElement) {
...
-                        maintainAspectRatio: false,
-                        onResize:() => this.dispatchEvent(new OrAttributeHistoryEvent('resize')),
-                        plugins: {
-                            legend: {
-                                display: false
+                    toolbox: {
+                        right: '2%',
+                        top: '5%',
+                        feature: {
+                            dataView: {readOnly: true},
```
MartinaeyNL, line 421: Can we update the data view, so that "timestamp" and "value" texts are shown in the first row?
I saw that you can copy paste this into MS Excel, which is nice 👍

ui/component/or-attribute-history/src/index.ts
MartinaeyNL, line 1: General question; I see that, after a 1500ms debounce on zoom, we fetch new data, and overlay the chart with a "loading indicator" while it's busy. Is it possible to run this fetch in the background, and apply the data silently/smoothly? I think the debounce can be shorter once applied.

ui/component/or-attribute-history/src/index.ts
```diff
@@ -381,90 +396,136 @@ export class OrAttributeHistory extends translate(i18next)(LitElement) {
...
-                                pointHitRadius: Number(this._style.getPropertyValue("--internal-or-attribute-history-graph-point-hit-radius")),
-                                fill: false
+                this._chartOptions = {
+                    animation: false,
+                    grid: {
+                        show: true,
+                        backgroundColor: this._style.getPropertyValue("--internal-or-asset-viewer-panel-color"),
+                        borderColor: this._style.getPropertyValue("--internal-or-attribute-history-text-color"),
+                        left: 15,
+                        right: 15,
```
MartinaeyNL: *(If we remove the tooltip as mentioned in the comment on line 498, we can remove the left-right margin)*
```suggestion
                        left: 0,
                        right: 4,
```

MartinaeyNL (commented):
Sorry for the delay on this PR review.
Was really confused with the status on this, so my bad I didn't review it.

I tested this branch locally, together with one of our production databases, and it was looking very nice 👍 
Let me know what you think of the review comments.

ui/component/or-attribute-history/src/index.ts
Hackerberg43: yep, adjusted to 4,4 now, would cut some of the y axislabels text for large numbers otherwise.

ui/component/or-chart/package.json
Hackerberg43: 👍

yarn.lock
Hackerberg43: 👍

ui/component/or-attribute-history/src/index.ts
Hackerberg43: good point, fixed

ui/component/or-attribute-history/src/index.ts
Hackerberg43: good point, just kept the properties that were allready in now

ui/component/or-attribute-history/src/index.ts
Hackerberg43: The first row displays the dataset titles. So for this case there is an easy hack by just naming the dataset to "timestamp    value", which would be ok because it isn't used for other purposes. In the chart widget with multiple datasets it will require a custom function to alter the content, Is possible though: https://echarts.apache.org/en/option.html#toolbox.feature.dataView.optionToContent

ui/component/or-attribute-history/src/index.ts
Hackerberg43: Good point. I've changed it such that the the graph gets slightly opaque only during a zoom reload (such that you still get the indication the the data you are looking at will be refreshed). Reduced debounce time by half. What about that?

MartinaeyNL:
Let me know when either me or @DonWillems should review this again, @Hackerberg43 👍

Hackerberg43:
Hey @MartinaeyNL yes this is ready to review again. How could I notify you better?

MartinaeyNL:
Normally we'd re-request a review, (shown in the reviewers section on the right),
or simply place a comment with a ping to the reviewer, notifying that the changes are addressed and it's ready for review again.

I'll try to find the time to take a look 👍
@DonWillems don't think much changed on a visual perspective, but feel free to check them out (if you want).

MartinaeyNL:
Thanks a lot again for working on this @Hackerberg43 ! 🥳 
I'm happy with the changes, let's merge 💪

Hackerberg43:
Great! 💯  You could make a draft pull for the line & bar chart widget that don made in your repo. We could store these last improvements there as todo's , as they might be relevant.

Issue #1716, opened 2025-02-28
Add QoS support to the MQTT Agent and MQTTAgentLink


richturner:
## User Story
As an MQTT agent user I want to be able to specify the default QoS level for all publish and subscribe requests but also be able to override this for each agent link so that I can make use of this MQTT functionality.

## Acceptance Criteria
- [ ] 1. Add `publishQos` and `subscribeQos` attributes to `MQTTAgent` with allowed values 0,1,2
- [ ] 2. Add `qos` option to `MQTTAgentLink` this should be an Integer with Min 0 and Max 2 value constraint
- [ ] 3. Update classes to allow these parameters to be passed down to the MQTT client during publish and subscribe operations

## Implementation Notes
HiveMQ client defaults to QoS level 2, we should default to level 0 (more widely supported and less resource intensive)

PR #1717, opened 2025-02-28
Add MQTT QoS level support to MQTTAgent and MQTTAgentLink

richturner:
Implements #1716


model/src/main/java/org/openremote/model/value/ValueType.java
```diff
@@ -147,7 +147,7 @@ public static class MultivaluedStringMap extends HashMap<String, List<String>> {
 
     public static final ValueDescriptor<Integer> DIRECTION = new ValueDescriptor<>("direction", Integer.class,
         new ValueConstraint.Min(0),
-        new ValueConstraint.Max(359)
+        new ValueConstraint.Max(360)
```
richturner, line 150: Saw this throwing an exception in demo setup with weather API using the value `360` for wind direction so I've made this accept that value.

dominiquekleeven:
I have no comments on the code, looks good to me.

I also manually tested the MQTTAgent with the QoS levels configured, which works as intended. Also tested setting the QoS level via the Agent Link which also works. I used a Mosquitto MQTT server I have running locally for testing the connection/publish/subscribe.

dominiquekleeven (approved):

dominiquekleeven (approved):
I have no comments on the code, looks good to me.

I also manually tested the MQTTAgent with the QoS levels configured, which works as intended. Also tested setting the QoS level via the Agent Link which also works. I used a Mosquitto MQTT server I have running locally for testing the connection/publish/subscribe.

PR #1721, opened 2025-03-02
feat: custom tile server and tile uploads

Ekhorn:
## Description

Adds the ability to configure a custom tile server and upload custom vector tiles.

Tile Server
- Only accepts tile servers that support `/{z}/{x}/{y}` scheme.
- Takes precedence over custom tile uploads.

Tile Upload
- By default only accepts 1GB tile uploads and expects `*.mbtiles`.
- Validates the tile database file is a valid SQLite database.
- Allows overwriting tile upload limit with `OR_CUSTOM_MAP_SIZE_LIMIT` environment variable.

original pr: https://github.com/openremote/openremote/pull/1494

## Checklist

### DIFF
--- a/manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
+++ b/manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
@@ -58,6 +58,9 @@ public class ConfigurationService implements ContainerService {
     public static final String OR_MAP_TILES_PATH = "OR_MAP_TILES_PATH";
     public static final String OR_MAP_TILES_PATH_DEFAULT = "manager/src/map/mapdata.mbtiles";
 
+    public static final String OR_CUSTOM_MAP_TILES_PATH = "OR_CUSTOM_MAP_TILES_PATH";
+    public static final String OR_CUSTOM_MAP_TILES_PATH_DEFAULT = "manager/src/map/mapdata-custom.mbtiles";
+
     protected ManagerIdentityService identityService;
     protected PersistenceService persistenceService;
     protected Path pathPublicRoot;
@@ -65,6 +68,7 @@ public class ConfigurationService implements ContainerService {
     private static final Logger LOG = Logger.getLogger(ConfigurationService.class.getName());
 
     protected Path mapTilesPath;
+    protected Path customMapTilesPath;
     protected Path mapSettingsPath;
     protected Path managerConfigPath;
     protected AtomicReference<ManagerAppConfig> managerAppConfig;
@@ -96,6 +100,10 @@ public class ConfigurationService implements ContainerService {
                 .filter(Files::isRegularFile)
                 .findFirst().orElse(null);
 
+        customMapTilesPath = Stream.of(getString(container.getConfig(), OR_CUSTOM_MAP_TILES_PATH, OR_CUSTOM_MAP_TILES_PATH_DEFAULT), "/deployment/map/mapdata-custom.mbtiles", "/opt/map/mapdata-custom.mbtiles", OR_CUSTOM_MAP_TILES_PATH_DEFAULT)
+                .map(Path::of)
+                .map(Path::toAbsolutePath).findFirst().orElse(null);
+
         managerConfigPath = Stream.of(getPersistedManagerConfigPath(), pathPublicRoot.resolve("manager_config.json"))
                 .map(Path::toAbsolutePath)
                 .filter(Files::isRegularFile)
@@ -110,6 +118,7 @@ public class ConfigurationService implements ContainerService {
         LOG.info("\t- manager_config.json: " + managerConfigPath);
         LOG.info("\t- mapsettings.json: " + mapSettingsPath);
         LOG.info("\t- mapdata.mbtiles: " + mapTilesPath);
+        LOG.info("\t- mapdata-custom.mbtiles: " + customMapTilesPath);
     }
 
     @Override
@@ -125,6 +134,7 @@ public class ConfigurationService implements ContainerService {
     public String toString() {
         return "ConfigurationService{" +
                 "mapTilesPath=" + mapTilesPath +
+                ", customMapTilesPath=" + customMapTilesPath +
                 ", mapSettingsPath=" + mapSettingsPath +
                 ", managerConfigPath=" + managerConfigPath +
                 '}';
@@ -162,6 +172,10 @@ public class ConfigurationService implements ContainerService {
 
     public Path getMapTilesPath() {
         return mapTilesPath;
+    }
+
+    public Path getCustomMapTilesPath() {
+        return customMapTilesPath;
     }
 
     public ManagerAppConfig getManagerConfig() {
--- a/manager/src/main/java/org/openremote/manager/map/MapResourceImpl.java
+++ b/manager/src/main/java/org/openremote/manager/map/MapResourceImpl.java
@@ -23,13 +23,11 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.openremote.container.web.WebResource;
 import org.openremote.manager.security.ManagerIdentityService;
 import org.openremote.model.http.RequestParams;
-import org.openremote.model.manager.MapRealmConfig;
+import org.openremote.model.manager.MapConfig;
 import org.openremote.model.map.MapResource;
 
 import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.core.Response;
-
-import java.util.Map;
 
 public class MapResourceImpl extends WebResource implements MapResource {
 
@@ -42,7 +40,7 @@ public class MapResourceImpl extends WebResource implements MapResource {
     }
 
     @Override
-    public Object saveSettings(RequestParams requestParams, Map<String, MapRealmConfig> mapConfig) {
+    public Object saveSettings(RequestParams requestParams, MapConfig mapConfig) {
         return mapService.saveMapConfig(mapConfig);
     }
 
--- a/manager/src/main/java/org/openremote/manager/map/MapResourceImpl.java
+++ b/manager/src/main/java/org/openremote/manager/map/MapResourceImpl.java
@@ -31,6 +31,7 @@ import org.openremote.manager.security.ManagerIdentityService;
 import org.openremote.model.http.RequestParams;
 import org.openremote.model.manager.MapConfig;
 import org.openremote.model.map.MapResource;
+import org.openremote.model.util.ValueUtil;
 
 import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.core.Response;
@@ -79,20 +80,35 @@ public class MapResourceImpl extends WebResource implements MapResource {
     @Override
     public Response uploadMap(@Context HttpServletRequest request) {
         try (InputStream stream = request.getInputStream()) {
-            boolean isSaved = mapService.saveUploadedFile(stream, "mapdata-custom.mbtiles");
+            boolean isSaved = mapService.saveUploadedFile(stream);
+            ObjectNode response = ValueUtil.JSON
+                .createObjectNode()
+                .put("map-custom", isSaved);
+
             if (isSaved) {
                 return Response.ok("File uploaded successfully").build();
-            } else {
-                return Response.serverError().entity("File upload failed").build();
             }
+            return Response.status(Response.Status.BAD_REQUEST).entity(response).build();
         } catch (IOException error) {
-            throw new WebApplicationException(Response.Status.BAD_REQUEST);
+            throw new WebApplicationException("{\"map-custom\": false}", Response.Status.BAD_REQUEST);
         }
     }
 
     @Override
-    public Response removeMap(@Context HttpServletRequest request) {
-        mapService.removeUploadedFile("mapdata-custom.mbtiles");
-        return Response.noContent().build();
+    public Response isMapCustom() {
+        ObjectNode response = ValueUtil.JSON
+            .createObjectNode()
+            .put("map-custom", mapService.isCustomUploadedFile());
+
+        return Response.ok().entity(response).build();
+    }
+
+    @Override
+    public Response deleteMap(@Context HttpServletRequest request) {
+        boolean deleted = mapService.deleteUploadedFile();
+        if (deleted) {
+            return Response.noContent().build();
+        }
+        return Response.status(400).build();
     }
 }
--- a/manager/src/main/java/org/openremote/manager/map/MapResourceImpl.java
+++ b/manager/src/main/java/org/openremote/manager/map/MapResourceImpl.java
@@ -20,6 +20,12 @@
 package org.openremote.manager.map;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.ws.rs.core.Context;
+
+import java.io.IOException;
+import java.io.InputStream;
+
 import org.openremote.container.web.WebResource;
 import org.openremote.manager.security.ManagerIdentityService;
 import org.openremote.model.http.RequestParams;
@@ -68,5 +74,25 @@ public class MapResourceImpl extends WebResource implements MapResource {
         } else {
             throw new WebApplicationException(Response.Status.NO_CONTENT);
         }
+    }
+
+    @Override
+    public Response uploadMap(@Context HttpServletRequest request) {
+        try (InputStream stream = request.getInputStream()) {
+            boolean isSaved = mapService.saveUploadedFile(stream, "mapdata-custom.mbtiles");
+            if (isSaved) {
+                return Response.ok("File uploaded successfully").build();
+            } else {
+                return Response.serverError().entity("File upload failed").build();
+            }
+        } catch (IOException error) {
+            throw new WebApplicationException(Response.Status.BAD_REQUEST);
+        }
+    }
+
+    @Override
+    public Response removeMap(@Context HttpServletRequest request) {
+        mapService.removeUploadedFile("mapdata-custom.mbtiles");
+        return Response.noContent().build();
     }
 }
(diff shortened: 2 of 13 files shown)


manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -96,6 +100,10 @@ public void init(Container container) throws Exception {
                 .filter(Files::isRegularFile)
                 .findFirst().orElse(null);
 
+        customMapTilesPath = Stream.of(getString(container.getConfig(), OR_CUSTOM_MAP_TILES_PATH, OR_CUSTOM_MAP_TILES_PATH_DEFAULT), "/deployment/map/mapdata-custom.mbtiles", "/opt/map/mapdata-custom.mbtiles", OR_CUSTOM_MAP_TILES_PATH_DEFAULT)
+                .map(Path::of)
+                .map(Path::toAbsolutePath).findFirst().orElse(null);
+
```
richturner, line 106: I don't see a test for the existence of the file

manager/src/main/java/org/openremote/manager/map/MapResourceImpl.java
```diff
@@ -71,4 +76,44 @@ public byte[] getTile(int zoom, int column, int row) {
...
+                .createObjectNode()
+                .put("map-custom", isSaved);
+
+            if (isSaved) {
+                return Response.ok("File uploaded successfully").build();
+            }
+            return Response.status(Response.Status.BAD_REQUEST).entity(response).build();
+        } catch (IOException error) {
+            throw new WebApplicationException("{\"map-custom\": false}", Response.Status.BAD_REQUEST);
+        }
```
richturner, line 98: Don't see what value the `response` object have and then in the catch you manually type out the JSON, HTTP status code is sufficient here but please log details of exceptions or known errors.

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -75,14 +85,30 @@ public class MapService implements ContainerService {
...
-    public ObjectNode saveMapConfig(Map<String, MapRealmConfig> mapConfiguration) throws RuntimeException {
+    public ObjectNode saveMapConfig(MapConfig mapConfiguration) throws RuntimeException {
         if (mapConfig == null) {
             mapConfig = ValueUtil.JSON.createObjectNode();
         }
-        mapConfig.putPOJO("options", mapConfiguration);
+
+        ObjectNode vectorTiles = ValueUtil.JSON.valueToTree(mapConfiguration.sources.get("vector_tiles"));
+        if (vectorTiles.hasNonNull("url")) {
+            if (!vectorTiles.get("url").textValue().contains("/{z}/{x}/{y}")) {
```
richturner, line 97: Some comments about this check would be useful

model/build.gradle
```diff
@@ -11,7 +11,9 @@ dependencies {
     api "org.hibernate.orm:hibernate-core:$hibernateVersion"
     api "org.hibernate.validator:hibernate-validator:$hibernateValidatorVersion"
     api "jakarta.el:jakarta.el-api:$javaxELAPIVersion"
-
+    api ("org.jboss.resteasy:resteasy-undertow:$resteasyVersion") {
+        exclude group: "io.undertow",  module: "undertow-servlet-jakarta"
+    }
```
richturner, line 16: Model should not have dependency on this

manager/src/main/java/org/openremote/manager/map/MapService.java
Ekhorn: Not entirely sure how far I should go into detail. (accidentally wrote the wrong file name in the comment, will fix after feedback)

DonWillems:
Bug: If you add the URL of a maptile server, save, then make a change to any mapsetting, it will reset the URL and show the default map.

richturner:
> Bug: If you add the URL of a maptile server, save, then make a change to any mapsetting, it will reset the URL and show the default map.
> 
> I think it would be nicer to see the filename of the mbtiles you selected instead of the image upload button. It will be the same height as the URL input then. Something like this: image

Currently I don't see any filename on test3, just the delete button.

richturner:
Sorry to hijack this issue but the appearance page eager loads the map settings panels for each realm which makes a lot of map tile requests, the panels should be lazy loaded at least to avoid un-necessary requests.

richturner:
After file upload the whole page reloads twice it seems (looks glitchy)

ui/component/or-components/src/or-collapsible-panel.ts
```diff
@@ -141,9 +169,12 @@ export class OrCollapsiblePanel extends LitElement {
...
             </div>
-            <div id="content" class="${this.expandable && this.expanded ? "expanded" : ""}">
-                <slot name="content"></slot>
-            </div>
+            ${when(!this.lazyload || this.expanded, () => html`
+                    <div id="content" class="${this.expandable && this.expanded ? "expanded" : ""}">
+                        <slot name="content"></slot>
+                    </div>
+                `
+            )}
```
Ekhorn, line 177: This code may be redundant.. Ideally we want it work through the collapsible panel without adding the event.

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -356,23 +506,50 @@ export class PageConfiguration extends Page<AppStateKeyed> {
...
+                        console.error(reason);
+                    });
+                }
                 this.requestUpdate();
                 this.loading = false;
                 this.managerConfigurationChanged = false;
                 this.mapConfigChanged = false;
                 const configURL =  (MANAGER_URL ?? "") + "/api/master/configuration/manager";
                 fetch(configURL, {cache: "reload"})
-                window.location.reload();
```
Ekhorn, line 375: What to do with this change.

We could either only reload whenever the manager configuration changed. Or we could try to sync redux with the latest manager config.

richturner (changes requested):
Please check what happens if you upload a custom map tiles file and then the system restarts, I don't see any changes to `ConfigurationService` which uses `OR_MAP_TILES_PATH`

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -488,9 +491,10 @@ public boolean deleteUploadedFile() {
 
         boolean deleted = false;
         try {
+            connection.close();
```
richturner, line 494: I think `setData()` should do this to cleanup any existing connection before switching

richturner (changes requested):

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -196,9 +232,20 @@ public void setData() throws ClassNotFoundException, SQLException, NullPointerEx
         if (!mapTilesPath.toFile().exists()) {
             return;
         }
+        Path customMapTilesPath = getCustomMapDataPath();
+        if (customMapTilesPath != null) {
+            if (delete) {
+                connection.close();
```
richturner, line 238: try catch needed and should check != null

richturner: Don't think this should be conditional on delete any change to the connection var should first close old connection

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -377,6 +441,121 @@ public byte[] getMapTile(int zoom, int column, int row) {
         }
     }
 
+    public boolean saveUploadedFile(Path path, InputStream fileInputStream) {
+        Path previous = getCustomMapDataPath();
+
+        try (OutputStream outputStream = Files.newOutputStream(path)) {
```
richturner, line 447: write output file should be done in it's own try catch so the resource is closed before you then try and open the connection on the file.

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -314,28 +463,29 @@ export class PageConfiguration extends Page<AppStateKeyed> {
         };
     }
 
-    protected async getMapConfig(): Promise<{[id: string]: any}> {
+    protected async getMapConfig(): Promise<MapConfig> {
         const response = await manager.rest.api.MapResource.getSettings();
-        return (response.data.options as {[id: string]: any});
+        const { options, sources } = response.data as MapConfig;
```
richturner, line 468: As mentioned it is possible for `MapResource.getSettings()` to return 204 no content, `response.data` then returns `""` empty string so a thruthy check could help as currently this method returns an invalid MapConfig like:
```
{
  "options": undefined,
  "sources": undefined
```

If no map config is returned then the configuration page cannot configure the map so disable this part of the UI

richturner (changes requested):

Ekhorn:
Changes reintroduced metadata getting out of sync on invalid upload

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -451,12 +459,17 @@ public boolean saveUploadedFile(Path path, InputStream fileInputStream) {
...
                     LOG.log(Level.SEVERE, "Stream continued past content-length, deleting file.");
-                    this.setData(connection, metadata, true);
+                    this.setData(true);
                     return false;
                 }
                 outputStream.write(buffer, 0, bytesRead);
                 written += bytesRead;
             }
+        } catch (IOException | ClassNotFoundException | SQLException e) {
+            LOG.log(Level.SEVERE, "Failed to save custom map file.");
```
richturner, line 469:
```suggestion
            LOG.log(Level.SEVERE, "Failed to save custom map file.");
            this.setData(true);
            return false;
```

richturner (changes requested):
Just a minor issue I think, once resolved I will checkout locally and test

richturner (changes requested):

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -263,6 +289,10 @@ protected Path getPersistedMapConfigPath() {
         return persistenceService.getStorageDir().resolve("manager").resolve("mapsettings.json");
     }
 
+    protected Path getPersistedCustomTilesPath() {
+        return persistenceService.getStorageDir().resolve("map");
+    }
```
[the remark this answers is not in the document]
Ekhorn: I think you mentioned to persist the tiles to just `tmp` in development, so should I remove this method and use `persistenceService.getStorageDir()` directly?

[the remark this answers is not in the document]
richturner: The subdirectory is fine, reduces any chance of a collision with another file

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
richturner, line 1: mapTilesPath should be initialised like managerConfigPath (See it looks in the `getPersistedManagerConfigPath()` you would need an `Objects::nonNull` check in your filter).

manager/src/main/java/org/openremote/manager/app/ConfigurationService.java
```diff
@@ -110,6 +110,10 @@ public void init(Container container) throws Exception {
         LOG.info("\t- manager_config.json: " + managerConfigPath);
         LOG.info("\t- mapsettings.json: " + mapSettingsPath);
         LOG.info("\t- mapdata.mbtiles: " + mapTilesPath);
+        Path customMapTilesPath = getCustomMapTilesPath();
+        if (customMapTilesPath != null) {
+            LOG.info("\t- " + customMapTilesPath.getFileName() + ": " + customMapTilesPath);
+        }
```
richturner, line 116: This logic shouldn't be needed, should just output mapTilesPath which should resolve to custom file if set (see file comment I have added).

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -535,30 +535,44 @@ public void saveMapMetadata(Metadata metadata) {
...
+                if (!Files.exists(parent)) Files.createDirectories(parent);
+            } catch (IOException e) {
+                LOG.log(Level.SEVERE, "Could not create parent directory for custom tiles", e);
+                return null;
+            }
+            return parent.resolve(filename);
         }
-        return null;
+        return configurationService.getCustomMapTilesPath().getParent().resolve(filename);
+    }
```
richturner: This logic should be encapsulated in the `configurationService.getCustomMapTilesPath()` now have code duplication

richturner (changes requested):

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -269,14 +352,62 @@ export class PageConfiguration extends Page<AppStateKeyed> {
...
+                                                    () => this.shadowRoot.getElementById('fileupload-elem').click()
+                                                }">
+                                                    <input id="fileupload-elem" name="configfile" type="file" accept=".mbtiles" @change="${(e) => this.uploadCustomMap(e)}"/>
+                                                </or-mwc-input>
+                                                <or-mwc-input id="filename-elem" style="width: unset" .value="${this.customMapFilename}" .label="${i18next.t("file")}" .type="${InputType.TEXT}" disabled>
+                                                </or-mwc-input>
+                                                ${when(this.customMapFilename, () => html`<or-mwc-input type="${InputType.BUTTON}" iconColor="black" icon="delete"
+                                                    @or-mwc-input-changed="${async () => await this.deleteCustomMap()}"
+                                                />`)}
```
MartinaeyNL, line 399: Minor edit here;
Sadly, web components specification requires a closing tag; https://github.com/WICG/webcomponents/issues/624
```html
<or-mwc-input ... ></or-mwc-input>
```

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -269,14 +352,62 @@ export class PageConfiguration extends Page<AppStateKeyed> {
...
+                                            .type="${InputType.URL}"
+                                            .label="${i18next.t("configuration.global.tileServerPlaceholder")}"
+                                            placeholder="https://api.example.com/tileset/{z}/{x}/{y}"
+                                            @or-mwc-input-changed="${(e: OrInputChangedEvent) => {
+                                                    this.mapConfig.sources.vector_tiles = {
+                                                        type: "vector",
+                                                        ...(e.detail.value ? { tiles: [e.detail.value], custom: true } : {}),
+                                                        custom: true
+                                                    }
+                                                    this.mapConfigChanged = true;
```
MartinaeyNL, line 375: I would move this out to a separate function, instead of inline.

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -238,17 +316,22 @@ export class PageConfiguration extends Page<AppStateKeyed> {
...
                     <div id="wrapper">
                         <div id="header-wrapper">
                             <div id="header-title">
                                 <or-icon icon="palette-outline"></or-icon>
-                                ${i18next.t("appearance")}
+                                <or-translate value="appearance"></or-translate>
                             </div>
                             <div id="header-actions">
-                                <or-mwc-input id="save-btn" .disabled="${!this.managerConfigurationChanged && !this.mapConfigChanged}" raised type="button" label="save"
+                                <or-mwc-input id="save-btn" .disabled="${!this.managerConfigurationChanged && !this.mapConfigChanged && !this.tilesForUpload && !this.tilesForDeletion}" raised type="button" label="save"
```
MartinaeyNL, line 334: For readability purposes, move this long statement from inline-JS to a small `const disabled = ... ;` line.

ui/app/manager/src/pages/page-configuration.ts
```diff
@@ -29,13 +29,14 @@ import "@openremote/or-components/or-collapsible-panel";
...
-import {ManagerAppConfig, MapRealmConfig, Realm} from "@openremote/model";
+import {ManagerAppConfig, MapConfig, Realm} from "@openremote/model";
 import {i18next} from "@openremote/or-translate";
 import "@openremote/or-components/or-loading-indicator";
 import {OrConfRealmCard} from "../components/configuration/or-conf-realm/or-conf-realm-card";
 import {OrConfPanel} from "../components/configuration/or-conf-panel";
-import { InputType } from "@openremote/or-mwc-components/or-mwc-input";
-import {DefaultHeaderMainMenu, DefaultHeaderSecondaryMenu, DefaultRealmConfig} from "../index";
+import { InputType, OrInputChangedEvent } from "@openremote/or-mwc-components/or-mwc-input";
+import {DefaultHeaderMainMenu, DefaultHeaderSecondaryMenu, DefaultRealmConfig, store} from "../index";
```
MartinaeyNL, line 38: Are these necessary? Wouldn't be nice if they are.

MartinaeyNL (commented):
A few comments on the UI; they are very minor 😉

ui/app/manager/src/pages/page-configuration.ts
Ekhorn: `store` is accidental, will remove it.

manager/src/main/java/org/openremote/manager/map/MapService.java
```diff
@@ -377,6 +489,116 @@ public byte[] getMapTile(int zoom, int column, int row) {
...
+
+    public void deleteUploadedFile() throws IOException {
+        Path previousCustomTilesPath = configurationService.getCustomMapTilesPath(true);
+
+        if (previousCustomTilesPath.toFile().isFile()) {
+
+            setData(true);
+            // TODO: Is this save necessary?
+            saveMapMetadata(metadata);
```
MartinaeyNL, line 562: Address this TODO statement; is this necessary?
Put a simple context message in the code I'd say.

MartinaeyNL:
I've asked @Ekhorn to write automated tests for it, as there is no coverage for this currently.
He will pick that up with this issue; https://github.com/openremote/openremote/issues/1834
We'll merge this for now, as extensive frontend testing has been done instead.

MartinaeyNL (approved):

PR #1722, opened 2025-03-04
fix: external attribute change becomes user change (#1711)

### LINKED ISSUE #1711: Asset Attribute linked to HTTP Agent (with PollingMillis) enables "to save" state after saving on Edit Asset Panel
**Describe the bug**

Discard asset change popup displayed after saving when changing asset attribute with an HTTP Agent connected with PollingMillis set. It keeps updating the value while in edit mode which causes the save button / discard behavior to inconsistently enable after saving your changes.

**To Reproduce**

1. Create an asset
2. Create an HTTP agent
3. Link the HTTP agent
4. Configure PollingMillis (or any other option that updates the attribute value besides user input)
5. Go to Assets modify page
6. Change an attribute value
7. Click Save
8. Shows save button inconsistently re-enabling as if changes were made again
8. Click away
9. See discard pop window

**Expected behavior**

The save button to be disabled after saving and no discard pop up to show.

**Screenshots**

Shows save button re-enabled and discard popup window.

**Desktop (please complete the following information):**
 - OS: [e.g. Linux]
 - Browser [e.g. Chromium, Firefox]
 - Version [e.g. Chromium 132.0.6834.110, Mozilla Firefox 134.0.2]

**Additional context**

The behavior is quite inconsistent because the HTTP Agent [at most requests every 5000 milliseconds](https://github.com/openremote/openremote/blob/e94e46499d0f01a6a8d952ff932dfecbe49f673c/agent/src/main/java/org/openremote/agent/protocol/http/HTTPProtocol.java#L280). If you would change the value in the code to 500 this would occur consistently.

Ekhorn:
## Description

The UI would incorrectly assume an external change for a user change upon saving asset changes. After saving the asset changes the save button would re-enable and navigating away would cause the discard popup to show. 

Any changes coming from external sources during the saving phase are now handled as external changes.

Closes #1711 

## Checklist


ui/component/or-asset-viewer/src/or-edit-asset-panel.ts
```diff
@@ -204,7 +204,19 @@ export class OrEditAssetPanel extends LitElement {
...
     shouldUpdate(changedProperties: PropertyValues): boolean {
         if (changedProperties.has("asset")) {
-            this.changedAttributes = [];
+            const oldAsset = changedProperties.get("asset") as Asset;
+            if (oldAsset?.attributes) {
+                // Handles external attribute changes post saving phase.
+                this.changedAttributes = Object.values(oldAsset.attributes).filter(({ name, timestamp: oldTimestamp, value: oldValue }) => {
+                  if (this.asset.attributes && name) {
+                    const attr = this.asset.attributes[name];
+                    return !Util.objectsEqual(oldValue, attr.value) || oldTimestamp !== attr.timestamp
```
richturner, line 213: Just looking at timestamp should be sufficient the backend ensures that any change to an attributes value must have a newer timestamp

richturner (changes requested):

ui/component/or-asset-viewer/src/or-edit-asset-panel.ts
Ekhorn: Tested it, it works :)

ui/component/or-asset-viewer/src/or-edit-asset-panel.ts
```diff
@@ -204,7 +204,18 @@ export class OrEditAssetPanel extends LitElement {
...
+                // Handles external attribute changes post saving phase.
+                this.changedAttributes = Object.values(oldAsset.attributes).filter(({ name, timestamp: oldTimestamp, value: oldValue }) => {
+                  if (this.asset.attributes && name) {
+                    return oldTimestamp !== this.asset.attributes[name].timestamp
+                  }
+                  return false;
+                }).map(({ name }) => name!);
+            } else {
+                this.changedAttributes = [];
+            }
```
richturner, line 218: Might be better to work from the new asset and compare with old as an attribute could have been added or removed and this would not detect it, something like:

```suggestion
            const oldAsset = changedProperties.get("asset") as Asset;
            if (this.asset?attributes && oldAsset?.attributes) {
                // Handles external attribute changes post saving phase.
                this.changedAttributes = Object.values(this.asset.attributes).filter(({ name, timestamp }) => {
                  return oldAsset.attributes[name]?.timestamp !== timestamp
                }).map(({ name }) => name!);
            } else {
                this.changedAttributes = [];
            }
```

richturner (changes requested):

ui/component/or-asset-viewer/src/or-edit-asset-panel.ts
```diff
@@ -204,7 +204,15 @@ export class OrEditAssetPanel extends LitElement {
 
     shouldUpdate(changedProperties: PropertyValues): boolean {
         if (changedProperties.has("asset")) {
-            this.changedAttributes = [];
+            const oldAsset = changedProperties.get("asset") as Asset;
+            if (this.asset.attributes && oldAsset?.attributes) {
```
richturner, line 208: We should probably test for the asset ID being the same as well and if it isn't then all attributes have changed

richturner (changes requested):

PR #1735, opened 2025-03-11
Add writeattribute MQTT topic to allow specifying value and timestamp

### LINKED ISSUE #1674: MQTT API: Add topic to publish an attribute with timestamp
## User Story
As a developer I want to write a value to an attribute including the timestamp I want to have set in the database, so that I can control the timestamps of the values I write and possibly align them with others.

## Description
In some cases you want several value that are published from a device to have the same value even though the aren't processed in the same millisecond by OpenRemote. This provides a cleaner dataset to potentially use in future analysis.

## Acceptance Criteria
- [ ] 1. An MQTT client can publish data and set the timestamp of it.
- [ ] 2. If the timestamp is older than another datapoint, do not write it to the asset attribute or go through rules but do write it to the datapoint table
- [ ] 3. It is still possible to publish a value without the timestamp
- [ ] 4. Documentation added

## Implementation Notes
-

richturner:
Closes #1674

### DIFF
--- a/manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
+++ b/manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
@@ -19,6 +19,7 @@
  */
 package org.openremote.manager.mqtt;
 
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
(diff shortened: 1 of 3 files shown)


richturner:
Need to also approve and merge https://github.com/openremote/documentation/pull/62

dominiquekleeven (approved):
Everything looks OK. 

I tested the new topic locally, the datapoint is properly written with the specified timestamp and value.

PR #1781, opened 2025-03-30
Mathematical Expression `ValueFilter` using Exp4j

### LINKED ISSUE #1429: Agent/attribute link parameter: add mathematical calculation to the value converter
As a user I want to transform the incoming value before writing it to the attribute, so that it fits my use case/business logic.
E.g. I want the value coming into my solar power attribute to be multiplied by -1000 as I need it in Watts and as a negative number.

Manipulations to be available:
Arithmetic action: - + / *
Exponentiation: ^
Absolute: |%VALUE%|

First verify if this can be added to the agent and attribute link parameters 'Value Converter' and 'Write Value Converter', or if a new meta items needs to be created.

Open questions:
Do we want sqrt(%VALUE%) or log(%VALUE%)? Leave them for when we see the need?

pankalog:
## Description

ValueFilter that uses Exp4j to perform mathematical calculations on an incoming AgentLink value. closes #1429 .

Users can refer to [Exp4j's documentation](https://www.objecthunter.net/exp4j/) for formatting their expression.

The ValueFilter checks if the value is null or not, then tries to cast the value into a Number, and if it's not a number by now, skip the calculation by returning null.

## Checklist


model/src/main/java/org/openremote/model/value/MathExpressionValueFilter.java
```diff
@@ -0,0 +1,49 @@
...
+
+        if (value instanceof String){
+            try{
+                value = Double.valueOf((String) value);
+            } catch (Exception ignored) {}
+        }
+
+        if (!(value instanceof Number)) {
+            return null;
+        }
```
richturner, line 41: Just use ValueUtil to coerce the value to a `Number.class`

model/src/main/java/org/openremote/model/value/MathExpressionValueFilter.java
```diff
@@ -27,23 +30,13 @@ public MathExpressionValueFilter(@JsonProperty("expression") String expression)
...
-            try{
-                value = Double.valueOf((String) value);
-            } catch (Exception ignored) {}
-        }
-
-        if (!(value instanceof Number)) {
-            return null;
-        }
+        Optional<Double> coercedValue = ValueUtil.getDoubleCoerced(value);
+        if (coercedValue.isEmpty()) return null;
```
richturner, line 34: You can use lambda `Optional.map()` here

richturner (approved):

DonWillems:
Hi Panos, did you also add some documentation for this?

PR #1782, opened 2025-03-31
Chart: Bugfix for undefined unit text on predicted datapoints

dominiquekleeven:
Minor fox for the `undefined` text that is currently shown when hovering over any predicted datapoints in the or-chart

**Before**
afbeelding

**After**
afbeelding


MartinaeyNL (approved):
Thanks!

Issue #1705, opened 2025-02-24
Research: How can we decouple the IoT data from the virtual machine and store them on a separate EBS (data) volume?


denniskuijs:
## User Story
As a system administrator, I want to store the IoT data on a separate EBS (data) volume so that backups can be more targeted, reducing the risk of data loss (during updates) by decoupling the data from the virtual machine.

## Description
**What?**
How can we decouple the IoT data from the virtual machine and store them on a separate EBS (data) volume?

**How?**
- Gap Analysis
  - Describe the current situation and give an idea about the desired situation.
- Decomposition
    - Create a flowchart that visualize the working of the current CI/CD-pipeline (provision host/account)
    - Create a architecture diagram that visualize the working of the EC2 CloudFormation template.
-  Literature Study
    -  Investigate the possibilities for storing data on an separate EBS (data) volume that is connected to the same EC2 instance.
    -  Investigate how the IoT data from OpenRemote can be stored on this EBS (data) volume.
    -  Investigate how this process can be automated and which changes need to be made.
- Prototyping
    - Develop a prototype that demonstrates a solution where the IoT data is stored on a separate EBS data volume and can be used among other OpenRemote (EC2) instances without any issues or data loss. 

**Why?**
This research is part of the development of a Proof of Concept (POC) aimed at automating the update process for the
virtual machines.

## Acceptance Criteria
- [x] An flowchart has been made that visualize the working of the current CI/CD-pipeline (provision host/account).
- [x] An architecture diagram has been made that visualize the working of the EC2 CloudFormation template.
- [x] There is an working prototype that demonstrates a solution where the IoT data is stored on a separate EBS (data) volume and can be used among other OpenRemote (EC2) instances without any issues or data loss. 
- [x] The research is fully documented.
- [x] The results are discussed with at least one team member.
- [x] The next steps are discussed.

## Implementation Notes
The solution may be based off how the data is currently stored on the virtual machines.

Issue #1757, opened 2025-03-19
CI/CD: Add CloudWatch metrics/alarms for separate EBS (data) volume


denniskuijs:
## User Story
As a system administrator, I want to monitor the performance and health of the separate `EBS` volume, so that we can quickly respond to any issues.

## Description
To monitor the performance and health of the separate `EBS` volume, several `CloudWatch` metrics/alarms need to be created and configured.

## Acceptance Criteria
- [x] 1. The `CloudWatch` metrics/alarms are created and configured.
- [x] 2. The `CloudWatch` metrics/alarms are fully tested and confirmed to be functional.
- [x] 3. The code has been reviewed by at least one team member.
- [x] 4. The implementation is fully documented.

## Implementation Notes
The implementation will be added to the existing CI/CD-pipeline `provision host`

Issue #1756, opened 2025-03-19
CI/CD: Add functionality for automatic snapshot creation


denniskuijs:
## User Story
As a system administrator, I want to automatically create snapshots (backups) of the separate `EBS` volume to ensure that the IoT data is fully backed up and can be restored in the event of data loss or other issues.

## Description
To ensure backups are created, a script for creating the `DLM (Amazon Data Lifecycle Manager)` policy needs to be implemented to automate the snapshot creation process.

## Acceptance Criteria
- [x] 1. The script for creating the `DLM (Amazon Data Lifecycle Manager)` policy is implemented.
- [x] 2. The script is fully tested and confirmed to be functional.
- [x] 3. The code has been reviewed by at least one team member.
- [x] 4. The implementation is fully documented.

## Implementation Notes
The implementation will be added to the existing CI/CD-pipeline `provision host`

Issue #1755, opened 2025-03-19
CI/CD: Add functionality for automatic attaching/detaching the separate EBS (data) volume


denniskuijs:
## User Story
As a system administrator, I want to be able to automatic attach/detach the separate `EBS` volume to prevent data loss.

## Description
To automatic attach/detach the separate `EBS` volume, a script needs to be created and added to the `SSM (Amazon Systems Manager)`

## Acceptance Criteria
- [x] 1. The script for attaching/detaching the separate `EBS` volume is implemented.
- [x] 2. The script is fully tested and confirmed to be functional.
- [x] 3. The code has been reviewed by at least one team member.
- [x] 4. The implementation is fully documented.

## Implementation Notes
The implementation will be added to the existing CI/CD-pipeline `provision host`

PR #1790, opened 2025-04-01
Feature/EBS Volume for storing/decoupling the IoT Data

denniskuijs:
## Description
Closes: #1705 #1754 #1755 #1756 #1757

## Checklist


.ci_cd/aws/cloudformation-create-ec2.yml
```diff
@@ -766,6 +776,27 @@ Resources:
...
+        prepare_volume:
+          commands:
+            01_mount_volume:
+              command: !Sub |
+                if [ -n "${SnapshotId}" ]; then
+                  sudo mount "${EBSDeviceName}" /var/lib/docker/volumes
+                else
+                  sudo mkfs -t xfs "${EBSDeviceName}"
+                  sudo mount "${EBSDeviceName}" /var/lib/docker/volumes
+                fi
```
richturner, line 789: Could you test if block device has a file system rather than using `SnapshotId`? Also see below comment about re-execution

.ci_cd/aws/cloudformation-create-ec2.yml
```diff
@@ -766,6 +776,27 @@ Resources:
...
+            02_configure_fstab:
+              command: !Sub |
+                UUID=$(sudo blkid -o value -s UUID "${EBSDeviceName}")
+                if [ -n $UUID ]; then
+                   sudo cp /etc/fstab /etc/fstab.orig
+                   sudo echo "UUID=$UUID /var/lib/docker/volumes xfs defaults,nofail 0 2" >> /etc/fstab
```
richturner, line 795: This cfn init code runs whenever a change is detected in the cloud formation stack metadata so this could result in duplicate entries in `/etc/fstab` (see `amazon-cloudwatch-agent-auto-reloader.conf`)

.ci_cd/aws/provision_host.sh
```diff
@@ -212,60 +192,315 @@ EOF
...
+  if [ -z "$INSTANCE_ID" ] && [ "$INSTANCE_STATE" != 'running' ]; then
+    echo "Failed to provision instance"
+    exit 1
+  fi
+
+  # Retrieve Volume ID and attach EBS data volume to instance.
+  echo "Instance is available .. attaching volume"
+  VOLUME_ID=$(aws ec2 describe-volumes --filters "Name=tag:Name,Values='$HOST/data'" --query "Volumes[?Tags[?Value=='$EBS_STACK_ID']].VolumeId" --output text $ACCOUNT_PROFILE 2>/dev/null)
+  VOLUME=$(aws ec2 attach-volume --device $EBS_DEVICE_NAME --instance-id $INSTANCE_ID --volume-id $VOLUME_ID --output text $ACCOUNT_PROFILE 2>/dev/null)
+  STATUS=$(aws ec2 describe-volumes --filters "Name=tag:Name,Values='$HOST/data'" --query "Volumes[?Tags[?Value=='$EBS_STACK_ID']].Attachments[].State" --output text $ACCOUNT_PROFILE 2>/dev/null)
```
richturner, line 342: Potential problem: We're attaching the volume but the ec2 cfn init script tries to format this block device but it may not be attached by that time, ec2 instance doesn't report healthy status until cfn init scripts run successfully.

.ci_cd/aws/cloudformation-create-dlm-policy.yml
richturner, line 1: Why isn't this part of the `create-ec2` stack?

.ci_cd/aws/cloudformation-create-ebs-volume.yml
richturner, line 1: Why isn't this part of the `create-ec2` stack?

.ci_cd/aws/cloudformation-create-ssm-document.yml
```diff
@@ -0,0 +1,159 @@
...
+                  runCommand: 
+                    - |
+                      VOLUME=$(aws ec2 detach-volume --volume-id {{ VolumeId }})
+                      STATUS=$(aws ec2 describe-volumes --query "Volumes[?VolumeId=='{{ VolumeId }}'].State" --output text 2>/dev/null)
+                      
+                      while [[ "$STATUS" == 'in-use' ]] do
+                        echo "Instance is still in-use .. Sleeping 30 seconds"
+                        sleep 30
+                        STATUS=$(aws ec2 describe-volumes --query "Volumes[?VolumeId=='{{ VolumeId }}'].State" --output text 2>/dev/null)
+                      done
```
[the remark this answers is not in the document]
denniskuijs: Good point! We need this counter in place, as there’s a chance something could go wrong during the volume attachment process. It’s added in the latest version.

.ci_cd/aws/cloudformation-create-ssm-document.yml
```diff
@@ -0,0 +1,159 @@
...
+                          exit 1
+                       else
+                          echo "Volume attaching is complete"
+                       fi
+                      
+              - name: MountVolume
+                action: aws:runShellScript
+                inputs:
+                  runCommand:
+                    - mount {{ EBSDeviceName }} /var/lib/docker/volumes
```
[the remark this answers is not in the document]
denniskuijs: It make sense to stop Docker first, as this will shutdown OpenRemote in case it’s running and prevent access while the volume is not attached or available.

.ci_cd/aws/provision_host.sh
```diff
@@ -212,60 +192,315 @@ EOF
...
+        exit 1
+    else
+        echo "Stack creation is complete"
+    fi
+  fi
+fi
+
+# Provision SSM Documents
+echo "Provisioning SSM Documents for attaching/detaching EBS data volume"
+STATUS=$(aws cloudformation describe-stacks --stack-name $SSM_STACK_NAME --query "Stacks[0].StackStatus" --output text 2>/dev/null)
```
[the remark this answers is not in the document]
denniskuijs: The SSM scripts only need to be provisioned once per account, as we can pass the necessary information using parameters. To target the `EC2` instance, the `IAM` role assumed by the instance must have the appropriate permissions. I’ve moved the provisioning of the `SSM` documents to the `provision account` script.

.ci_cd/aws/cloudformation-create-ec2.yml
denniskuijs: I chose to use the `SnapshotId` here because `SSM` commands execute asynchronously and don’t wait for the response of the previous command. Since the `SnapshotId` is available after the command runs, this is no longer an issue. 

To improve this solution, I updated the logic to check for an existing filesystem instead. Because this check can only succeed once the volume is attached, I’ve added a while loop to continuously check until the filesystem is detected.

.ci_cd/aws/cloudformation-create-ec2.yml
denniskuijs: In the new implementation, this code has been removed from the `cfn-scripts`, which resolves this issue. Instead, we’re using the `SSM` documents to handle the volume attachment en preparation.

.ci_cd/aws/provision_host.sh
denniskuijs: Fixed within the new implementation, volume attachment en preparation is handled by the `SSM` documents after stack creation and isn’t part of the `provision host` script anymore.

.ci_cd/aws/cloudformation-create-dlm-policy.yml
denniskuijs: It makes sense to move this part to the `create-ec2` section, as the `Lifecycle manager` is part of `EC2`. To target the correct volume I’ve chosen to use the `StackId` variable. Since there are limited tags available, using `StackId` ensures the correct volume is selected, even when multiple volumes share the same name.

When provisioning the `DLM` policy, we need to provide the `ExecutionRoleArn` While it’s technically possible to create this role within `CloudFormation`, it can already exists, and the `CLI` provides a command to create the default role if needed. Therefore, I’ve chosen to check for this role within the `provision host` script and pass its value to the template.

.ci_cd/aws/cloudformation-create-ebs-volume.yml
denniskuijs: Initially, I considered splitting the provisioning process to prevent the volume from being detached or modified during `CloudFormation` updates. 

However, after moving the volume creation to the `create-ec2` section, it turned out this wasn’t the case. Only the root volume is affected during updates, while data volumes remain the same. For added safety, I’ve included a `DeletionPolicy: Snapshot` to ensure a final snapshot is created before the volume is deleted.

Since the volume must exists in the same `Availability zone` as the instance, it will be provisioned only after the instance is fully created and the `cfn-signal` has returned success. 

I’ve removed the steps for creating the filesystem and mounting the volume from the `cfn-scripts`. 
Instead, this process is moved to the `SSM` documents, which will execute within the `provision host` script after the stack is successfully created.

.github/workflows/provision-host.yml
```diff
@@ -38,7 +38,7 @@ on:
         default: '16'
         required: false
       SNAPSHOT_ID:
-        description: 'Create EBS data volume based on snapshot'
+        description: 'Snapshot ID to create volume based of an existing snapshot'
```
richturner, line 41:
```suggestion
        description: 'Snapshot ID to create volume based off an existing snapshot'
```

.ci_cd/aws/cloudformation-create-ssm-documents.yml
richturner, line 1: I'm wondering if we can have a document to replace the contents of an existing volume with a snapshot. For example I want to deploy a clone of host A data to host B (wiping the existing data from host B)?

richturner (changes requested):

.ci_cd/aws/cloudformation-create-ssm-documents.yml
```diff
@@ -1,145 +1,146 @@
...
+                        FILESYSTEM=$(blkid -o value -s TYPE {{ DeviceName }})
+                        count=$((count+1))
+                      done
+
+                      if [ -z "$FILESYSTEM" ]; then
+                        mkfs -t xfs {{ DeviceName }}
+                        mount {{ DeviceName }} /var/lib/docker/volumes
+                      else
+                        mount {{ DeviceName }} /var/lib/docker/volumes
+                      fi
```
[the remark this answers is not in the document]
denniskuijs: Good point! I’ve improved the implementation by combining the volume attachment and filesystem check into a single command. Since the filesystem check depends on the volume being attached, running them separately can cause issues since these commands are asynchronous and don’t wait for each other to finish.

In my last commit, the volume is first attached to the machine. Once the attachment is completed successfully, the script proceeds to retrieve the filesystem and applies the appropriate logic based on that. 

- No filesystem:
   Creates a new `XFS` filesystem
   Mount volume to the `/var/lib/docker/volumes` directory

- Filesystem exists:
   Mount volume to the `/var/lib/docker/volumes` directory

When a volume is attached for the first time and has no existing filesystem, the script will create one immediately after the volume is attached. Instead of waiting 900 seconds before proceeding.

.ci_cd/aws/cloudformation-create-ssm-documents.yml
denniskuijs: Unfortunately, It’s not possible to replace the contents of an existing volume with a snapshot. Snapshots can only be specified during the initial creation of the volume. Additionally, accessing the contents of the snapshot directly is difficult because snapshots are stored in a hidden `S3` bucket managed by Amazon.

The only way to retrieve data from a snapshot is to create a new volume from it, then attach and mount that volume to an `EC2` instance to inspect and access the data via the filesystem.

To improve flexibility, we could extend our implementation to also support provisioning a host based on an existing data volume using its `VolumeId`. This would give us three different options:

-  No `SnapshotId` or `VolumeId` specified: An empty data volume is created, attached and mounted.
- `SnapshotId` specified: a new data volume is created, attached and mounted from the given snapshot.
- `VolumeId` specified: the existing volume is used directly, no new volume is created.

richturner: So would it not be possible to:

* Create new volume from snapshot
* Detach existing data volume (maybe option to delete but keep by default - remove daily backups)
* Attach new volume

Would be nice to be able to do this on an existing host to allow things like deploying prod server snapshot to staging server for testing etc.

denniskuijs:
I improved the implementation for the EBS volume creation based on your last comment in the PR. I updated the SSM documents, they are now directly interacting with the AWS API using SSM Automation. 

With this approach, we don’t have to write AWS CLI commands in bash and check their status with several loops. This is now natively handled by SSM, including the addition to wait for the previous steps to finish before continuing.

SSM automation documents are combined with normal documents for executing commands on the EC2 machines as this feature is not supported in SSM Automation.
The solution is fully tested (including cross-account), worked as expected and is less error prone.

I’ve also added a SSM Document to replace an existing volume with an snapshot. Very useful to have.

richturner (dismissed):

richturner:
@denniskuijs We should be able to run `provision_account.sh` for existing accounts so the SSM documents get provisioned do you see any issue with this?

denniskuijs:
> @denniskuijs We should be able to run `provision_account.sh` for existing accounts so the SSM documents get provisioned do you see any issue with this?

@richturner I did a quick look again, there is no check if the `or-dashboard-default` stack is already created. When the `provision_account.sh` script is executed for the existing accounts, it probably tries to create this stack again which will definitely fail since the stack name is already in use.

Maybe adding an extra check to both the `or-dashboard-default` and `or-ssm` stack in the same way as provisioning the VPC stack (row 145) would be beneficial to have.

denniskuijs:
@richturner I did a manual test, the error message appears in the logs, but isn’t breaking the script.
To be consistent, it’s still a good idea to implement a check here. These checks are already be used on several places in the `provision_host.sh` script.

Screenshot 2025-05-15 at 22 11 46

denniskuijs:
@richturner After the PR is merged, we need to perform the following tasks:

- The `docker-compose.yml` file in the `custom-project` repository needs to be updated. We need to add the `PGDATA` environment variable to the `postgresql` container to prevent any permission related issues. This also needs to be updated on the existing custom-projects once they are using this new implementation.

- The `developers-access` role in AWS needs to be updated with the following policies:
    - `dlm:CreateLifecyclePolicy`
    - `dlm:TagResource`
    - `iam:GetRole`
    - `iam:GetPolicy`
    - `iam:CreateRole`
    - `iam:AttachRolePolicy`

denniskuijs:
@richturner I tested the implementation on OpenRemote's AWS account and it's working as expected. 

I configured the required `IAM` permissions in the `developers-access-eu-west-1` role and provisioned the `or-ssm` stack for the `SSM` documents. The only thing that's left here is updating the `or-dashboard-default` stack to ensure the `EBS` data volume becomes visible on the dashboard.

The `PGDATA` variable is not required in the `docker-compose.yml` file. Apparently, it was a misconfiguration in my own test setup. 

Besides that, I added several checks to check if the `or-ssm` and `or-dashboard-default` stacks are already created and if the `EBS` data volume is already attached/mounted to the instance.

richturner (approved):

PR #1797, opened 2025-04-02
Feature/html widget

Hackerberg43:
## Description
Adds a widget in which the user is able to add custom markupin html format. Main purpose is to be able to format the dashboard into a printable report page.
 Input is sanitized using dompurify to mitigate unsafeHTML usage risks.
Double client-side sanitation is present, during widget save (to show to the user that their inserted content gets cleaned) and during render (clients can still insert malicious html via API).

## Design considerations
- Dompurify exceptions: are images and links acceptable or should it be limited to text
- Server-side sanitation seems overkill. Widget config is made by authenticated user.

## Checklist

## Future
- Custom tags to insert dates & attribute values?


DonWillems:
I have trouble running this locally. I don't see the placeholder content you've added. When I something simple like 
```
<!DOCTYPE html><html><body><p>yes</p></body></html>
```
it does save it. But I see an empty widget in modify and view mode. When inspecting the widget with the dev tools it also shows an empty html widget.
If I open the HTML editor again, it always has `""` added around the saved text.

I see you thought about sanitizing the input. Do you think it would be better to restrict it further by using Markdown instead of HTML? That might also make it a bit easier for the less proficient users.

Hackerberg43:
Hey Don, 
Did you do a full gradle build or only serve the ui from the ide? Also, which browser are you using?
I suspect the dompurify library isn't being picked up.

I've looked into Markdown, it is not safer. In the end it gets converted back to html and still requires the lit function unsafeHTML.
From a user perspective, if you think markdown is preffered then I can switch it to that no problem. There are word like online editors for both markdown and html.

DonWillems:
Hi Wouter,
I did build yes, I'll ask someone else to test too.

For the users we typically see using the insights page a markdown widget would be more accessible indeed. And for its purpose of adding a title and some explanation to an insights page that would work out nicely.
Will there also be a simple toolbar similar to what I see above this comment field? Or am I asking for too much now :)

Hackerberg43:
Allright, well when you want a shortcut bar you'll need another library because the ace-editor you are currently using does not have this feature. Could insert something old like https://markitup.jaysalvat.com/examples/markdown/ or use a different editor library for this widget all together, please discuss internally :)

PR #1801, opened 2025-04-03
UI updates: Remove now obsolete config env var from webpack serve

### LINKED ISSUE #1800: UI update: Remove now obsolete config env var from webpack serve
Webpack serve script for manager included an env var to statically copy the `manager_config.json` file into the webpack dev server but this file is now served by a HTTP endpoint in the manager HTTP API.

Seems the insights app had copy and pasted the same logic so it should be removed from each to avoid confusion.

richturner:
Closes #1800


richturner:
@MartinaeyNL I've added a bit more info to the linked issue now

MartinaeyNL (approved):

MartinaeyNL:
Thanks for the additional info. Let's get this merged 💪

PR #1806, opened 2025-04-04
Refactor identity provider role management

### LINKED ISSUE #1775: Users: if a user is created with a realm role like 'restricted user' or 'super admin' through the Manager UI, that user can not reach the Manager
**Describe the bug**
If a user is created with the realm role 'restricted user' through the Manager UI, the default role is not applied. This causes issues with accessing the manager. The default should always be applied when a new user is created.

**To Reproduce**
1. Go to the Users page in the Manager and start adding a new user
2. Select the Restricted User role for that user
3. Create the user
4. Try to log in to the manager as that user
5. Get a blank screen and 401 message

**Expected behavior**
5. Log in as the user

**OR Version**
1.3.4

**Additional context**
The default role is not applied when you follow these steps. Make sure the defaul-role gets applied even when a user is created with restricted user role selected.

richturner:
Fixes #1775

### DIFF
--- a/model/src/main/java/org/openremote/model/security/UserResource.java
+++ b/model/src/main/java/org/openremote/model/security/UserResource.java
@@ -19,16 +19,14 @@
  */
 package org.openremote.model.security;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.tags.Tag;
-import org.openremote.model.Constants;
-import org.openremote.model.http.RequestParams;
-import org.openremote.model.query.UserQuery;
-
 import jakarta.annotation.security.RolesAllowed;
 import jakarta.validation.Valid;
 import jakarta.ws.rs.*;
+import org.openremote.model.Constants;
+import org.openremote.model.http.RequestParams;
+import org.openremote.model.query.UserQuery;
 
 import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
 
@@ -39,14 +37,6 @@ import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
 @Tag(name = "User", description = "Operations on users")
 @Path("user")
 public interface UserResource {
-
-    @GET
-    @Path("{realm}/roles")
-    @Produces(APPLICATION_JSON)
-    @SuppressWarnings("unusable-by-js")
-    @RolesAllowed(Constants.READ_ADMIN_ROLE)
-    @Operation(operationId = "getRoles", summary = "Retrieve roles for a realm")
-    Role[] getRoles(@BeanParam RequestParams requestParams, @PathParam("realm") String realm);
 
     @GET
     @Path("{realm}/{clientId}/roles")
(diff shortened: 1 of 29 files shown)


richturner:
Whilst looking at the linked issue I thought it was a good opportunity to decouple the HTTP API realm and client roles from Keycloak implementation details so we just deal with simple strings when assigning realm and client roles to users.

Also got rid of un-neccessary endpoints (ones that can be replaced with other existing endpoints).

model/src/main/java/org/openremote/model/security/UserResource.java
```diff
@@ -40,14 +38,6 @@
...
 public interface UserResource {
 
-    @GET
-    @Path("{realm}/roles")
-    @Produces(APPLICATION_JSON)
-    @SuppressWarnings("unusable-by-js")
-    @RolesAllowed(Constants.READ_ADMIN_ROLE)
-    @Operation(operationId = "getRoles", summary = "Retrieve roles for a realm")
-    Role[] getRoles(@BeanParam RequestParams requestParams, @PathParam("realm") String realm);
-
```
MartinaeyNL, line 50: Don't we want to deprecate this endpoint instead of removing it?
Especially since it was calling `getClientRoles` in `UserResourceImpl` anyway.
So other than "cleaning up" it doesn't cause any maintenance overhead.

Alternatively, we could include this a minor update like 1.4.0, and notify users of the API change.

model/src/main/java/org/openremote/model/security/UserResource.java
```diff
@@ -126,62 +116,43 @@ public interface UserResource {
...
     @Path("{realm}/userRealmRoles/{userId}")
     @Produces(APPLICATION_JSON)
     @Operation(operationId = "getUserRealmRoles", summary = "Retrieve realm roles for a user in a realm")
-    Role[] getUserRealmRoles(@BeanParam RequestParams requestParams, @PathParam("realm") String realm, @PathParam("userId") String userId);
-
-    @GET
-    @Path("userRoles")
-    @Produces(APPLICATION_JSON)
-    @Operation(operationId = "getCurrentUserRoles", summary = "Retrieve client roles for the currently authenticated user")
-    Role[] getCurrentUserRoles(@BeanParam RequestParams requestParams);
```
MartinaeyNL, line 151: Same in regards to this endpoint; don't we deprecate this, or introduce it as an API update?

model/src/main/java/org/openremote/model/security/UserResource.java
```diff
@@ -126,62 +116,43 @@ public interface UserResource {
...
     @Produces(APPLICATION_JSON)
     @Operation(operationId = "getCurrentUserRealmRoles", summary = "Retrieve realm roles for the currently authenticated user")
-    Role[] getCurrentUserRealmRoles(@BeanParam RequestParams requestParams);
-
-    @PUT
-    @Path("{realm}/userRoles/{userId}")
-    @Consumes(APPLICATION_JSON)
-    @RolesAllowed(Constants.WRITE_ADMIN_ROLE)
-    @Operation(operationId = "updateUserRoles", summary = "Update client roles for a user in a realm")
-    void updateUserRoles(@BeanParam RequestParams requestParams, @PathParam("realm") String realm, @PathParam("userId") String userId, Role[] roles);
```
MartinaeyNL, line 170: Same in regards to this endpoint; do we deprecate this, or introduce it as an API update?

model/src/main/java/org/openremote/model/security/UserResource.java
MartinaeyNL, line 1: Note: these changes of payloads should be addressed in the release notes later

test/src/test/groovy/org/openremote/test/model/AssetModelTest.groovy
```diff
@@ -8,6 +8,7 @@ import org.openremote.agent.protocol.simulator.SimulatorAgent
 import org.openremote.agent.protocol.velbus.VelbusTCPAgent
 import org.openremote.container.persistence.PersistenceService
 import org.openremote.container.timer.TimerService
+import org.openremote.model.asset.impl.UnknownAsset
```
MartinaeyNL, line 11: Is this an unused import?

ui/app/manager/src/headers.ts
```diff
@@ -1,4 +1,4 @@
-import { manager } from "@openremote/core";
+import {manager, RESTRICTED_USER_REALM_ROLE} from "@openremote/core";
```
MartinaeyNL, line 1: Additional import can be removed.

ui/component/core/src/index.ts
```diff
@@ -835,7 +837,7 @@ export class Manager implements EventProviderFactory {
                 authenticated = true;
 
                 // Get user roles
-                const rolesResponse = await rest.api.UserResource.getCurrentUserRoles();
+                const rolesResponse = await rest.api.UserResource.getCurrentUserClientRoles(OPENREMOTE_CLIENT_ID);
```
MartinaeyNL, line 840: Replace `OPENREMOTE_CLIENT_ID` with `this._config.clientId || OPENREMOTE_CLIENT_ID`

ui/component/or-asset-viewer/src/index.ts
```diff
@@ -973,9 +972,9 @@ async function getLinkedUserInfo(userAssetLink: UserAssetLink): Promise<UserAsse
     const userId = userAssetLink.id!.userId!;
     const username = userAssetLink.userFullName!;
 
-    const roleNames = await manager.rest.api.UserResource.getUserRoles(manager.displayRealm, userId)
+    const roleNames = await manager.rest.api.UserResource.getUserClientRoles(manager.displayRealm, userId, OPENREMOTE_CLIENT_ID)
```
MartinaeyNL, line 975: Shouldn't we use `managerConfig.clientId` instead of `OPENREMOTE_CLIENT_ID` here as well?

model/src/main/java/org/openremote/model/security/UserResource.java
richturner: Given we have static typing of the frontend model and this endpoint is specifically only for our own `openremote` client there shouldn't be outside usage and any that there is (in custom projects for example) will quickly generate a compilation error and then a simple method rename is all that is needed.

Even if this endpoint was left there would be a breaking change as the return type is now different.

The UserResource is really intended for use by the manager which is updated in lockstep, deciding if this is an announced breaking change is something we should decide.

model/src/main/java/org/openremote/model/security/UserResource.java
richturner: See other comment

model/src/main/java/org/openremote/model/security/UserResource.java
richturner: Relates to my comment as well I think

ui/component/core/src/index.ts
richturner: Switched to use `this._config.clientId` as `normaliseConfig` will take care to set this

MartinaeyNL (approved):

PR #1821, opened 2025-04-11
Fix path based proxying issues

### LINKED ISSUE #1772: Path based proxy configuration not working correctly
Some additional issues were found while further testing path based proxying (#1669):

* The index.html of UIs use absolute URLs for loading TypeScript, (e.g. `/manager/bundle.068f51ae1ab400417aed.js`)
* The UI uses absolute URLs for some API requests instead of relative URLs
* The URLs in Keycloak responses are wrong unless it is reconfigured using the `KC_HTTP_RELATIVE_PATH` env var. So the proxy should not change paths in requests send to Keycloak.

wborn:
Fixes #1772


ui/component/or-survey/src/index.ts
MartinaeyNL, line 1: You can remove the `declare var MANAGER_URL: string;` here.

wborn:
Draft: Let's merge this after a new release.

PR #1824, opened 2025-04-15
Determine loopback IP for machines with complex network configs

ebariaux:


richturner:
@ebariaux Is this the only test that had issues as TcpTest etc. bind to `127.0.0.1` also but I guess the server and client use the same code so they resolve to the same address

richturner (approved):

Issue #1812, opened 2025-04-07
Asset lock can throw IllegalMonitorStateException


richturner:
**Describe the bug**
See the below stack trace found in a deployment log:

```
2025-04-04T06:42:27.230+01:00 java.lang.IllegalMonitorStateException
2025-04-04T06:42:27.230+01:00 at java.base/java.util.concurrent.locks.ReentrantLock$Sync.tryRelease(ReentrantLock.java:175)
2025-04-04T06:42:27.230+01:00 at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.release(AbstractQueuedSynchronizer.java:1059)
2025-04-04T06:42:27.230+01:00 at java.base/java.util.concurrent.locks.ReentrantLock.unlock(ReentrantLock.java:494)
2025-04-04T06:42:27.230+01:00 at org.openremote.model.util.LockByKey.unlock(LockByKey.java:39)
2025-04-04T06:42:27.230+01:00 at org.openremote.manager.asset.AssetStorageService.withAssetLock(AssetStorageService.java:1156)
2025-04-04T06:42:27.230+01:00 at org.openremote.manager.asset.AssetProcessingService.processAttributeEvent(AssetProcessingService.java:343)
2025-04-04T06:42:27.230+01:00 at org.openremote.manager.asset.AssetProcessingService.lambda$configure$8(AssetProcessingService.java:293)
2025-04-04T06:42:27.230+01:00 at io.micrometer.core.instrument.AbstractTimer.record(AbstractTimer.java:202)
2025-04-04T06:42:27.230+01:00 at org.openremote.manager.asset.AssetProcessingService.lambda$configure$9(AssetProcessingService.java:293)
2025-04-04T06:42:27.230+01:00 at org.apache.camel.support.processor.DelegateSyncProcessor.process(DelegateSyncProcessor.java:65)
2025-04-04T06:42:27.230+01:00 at org.apache.camel.processor.errorhandler.RedeliveryErrorHandler$RedeliveryTask.doRun(RedeliveryErrorHandler.java:808)
2025-04-04T06:42:27.230+01:00 at org.apache.camel.processor.errorhandler.RedeliveryErrorHandler$RedeliveryTask.run(RedeliveryErrorHandler.java:714)
2025-04-04T06:42:27.230+01:00 at org.apache.camel.impl.engine.DefaultReactiveExecutor$Worker.doRun(DefaultReactiveExecutor.java:199)
[...]

PR #1826, opened 2025-04-16
Asset lock enhancements

richturner:
## Description
This PR is here to capture necessary additional improvements to the asset lock where issues were initially identified in #1812, some additional goals:

- Try to revert to ReentrantLock instead of Semaphore due to these objects being instantiated frequently and the former being more lightweight 
- Memory leak protection on edge cases
- Timeout on locks

Before being able to switch back to ReentrantLock then the cause of the threading issue in #1812 needs to be identified and addressed so that `unlock()` is always called by the same thread that called `lock()`. The stacktrace in the mentioned issue suggested that camel route processing was the cause of the thread context switch between the try and finally statements but I am not so sure about that and have been trying to recreate the exception (without success) using this simple test in a branch of the old `LockByKey`:

https://github.com/openremote/openremote/blob/asset_lock_investigation/test/src/test/groovy/org/openremote/test/benchmark/AttributeEventBenchmarkTest.groovy#L192

I think we need to get to the bottom of how these `IllegalMonitorStateException` errors are occurring to be able to correctly address this issue going forward, the error can be seen in many production environments by running the following command in an SSH session:
```
docker logs or-manager-1 2>&1 | grep 'IllegalMonitorStateException'
```

Whenever this exception occurs then it is likely that no changes can be made or attribute events processed for the assetID that currently owns the lock and due to there being no timeout on lock requests all Container executor threads can become blocked.

### DIFF
--- a/container/src/main/java/org/openremote/container/message/MessageBrokerService.java
+++ b/container/src/main/java/org/openremote/container/message/MessageBrokerService.java
@@ -30,6 +30,7 @@ import org.apache.camel.impl.health.ConsumersHealthCheckRepository;
 import org.apache.camel.impl.health.ContextHealthCheck;
 import org.apache.camel.impl.health.DefaultHealthCheckRegistry;
 import org.apache.camel.impl.health.RoutesHealthCheckRepository;
+import org.apache.camel.model.RedeliveryPolicyDefinition;
 import org.apache.camel.spi.ExecutorServiceManager;
 import org.apache.camel.spi.StreamCachingStrategy;
 import org.apache.camel.spi.ThreadPoolFactory;
@@ -122,8 +123,11 @@ public class MessageBrokerService implements ContainerService {
         StreamCachingStrategy streamCachingStrategy = new DefaultStreamCachingStrategy();
         streamCachingStrategy.setSpoolThreshold(524288); // Half megabyte
         context.setStreamCachingStrategy(streamCachingStrategy);
-
-        context.getCamelContextExtension().setErrorHandlerFactory(new DefaultErrorHandlerBuilder());
+        RedeliveryPolicyDefinition redeliveryPolicy = new RedeliveryPolicyDefinition();
+        redeliveryPolicy.setAsyncDelayedRedelivery("false");
+        DefaultErrorHandlerBuilder errorHandler = new DefaultErrorHandlerBuilder();
+        errorHandler.setRedeliveryPolicy(redeliveryPolicy);
+        context.getCamelContextExtension().setErrorHandlerFactory(errorHandler);
 
         if (container.isDevMode()) {
             context.setMessageHistory(true);
--- a/manager/src/main/java/org/openremote/manager/asset/AssetProcessingService.java
+++ b/manager/src/main/java/org/openremote/manager/asset/AssetProcessingService.java
@@ -54,6 +54,8 @@ import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static org.openremote.model.attribute.AttributeWriteFailure.*;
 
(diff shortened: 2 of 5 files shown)


Bragolgirith:
Not sure if your specific threading issues are caused by this, but the `LockByKey` class from this [article](https://www.baeldung.com/java-acquire-lock-by-key) is currently definitely not thread-safe. This can be tested as follows:

```java
class ReproducerTest {
    private static final Logger logger = LoggerFactory.getLogger(ReproducerTest.class);
    public static final int THREADS_COUNT = 4;
    public static final int NUM_ITERATIONS_PER_THREAD = 100;
    public static final String TEST_LOCK_NAME = "testLockName";
    @Test
    void useLock_multithreaded() throws InterruptedException {
        final LockByKey lockByKey = new LockByKey();
        final CountDownLatch threadCountDownLatch = new CountDownLatch(THREADS_COUNT);
        // Start multiple threads
        for (int i = 0; i < THREADS_COUNT; i++) {
            final int threadId = i;
            new Thread(() -> {
                // Perform multiple iterations in each thread
                for (int iterationId = 0; iterationId < NUM_ITERATIONS_PER_THREAD; iterationId++) {
                    logger.info("Running thread {}, iteration {}", threadId, iterationId);
                    try {
                        lockByKey.lock(TEST_LOCK_NAME);
                    } finally {
                        lockByKey.unlock(TEST_LOCK_NAME);
                    }
                }
                threadCountDownLatch.countDown();
            }).start();
        }
        // Wait for all threads to complete
        threadCountDownLatch.await();
    }
}
```

which will sometimes hang and throw the following exception on the  `lockWrapper.lock.unlock();` line:
```
java.lang.NullPointerException: Cannot read field "lock" because "lockWrapper" is null
```

The problem is that the `unlock` method doesn't synchronize decrementing the counter and removing the item from the map:

```java
    public void unlock(String key) {
        LockWrapper lockWrapper = locks.get(key);
        lockWrapper.lock.unlock();
        if (lockWrapper.removeThreadFromQueue() == 0) { 
            // Multiple threads can reach this line concurrently and each may attempt to remove the lock
            // Calling Map#remove(Object key, Object value) doesn't help because the same `lockWrapper` instance is reused between the different threads.
            locks.remove(key, lockWrapper);
        }
    }
```

See [here](https://stackoverflow.com/a/77829625/3620458) for a better implementation that uses the `ConcurrentHashMap#computeIfPresent` method to decrement the counter and remove the item from the map in a synchronized manner.

richturner:
> Not sure if your specific threading issues are caused by this, but we found the `LockByKey` class from this [article](https://www.baeldung.com/java-acquire-lock-by-key) to not be thread-safe (and notified the author). This can be tested as follows:
> 
> ```java
> class ReproducerTest {
>     private static final Logger logger = LoggerFactory.getLogger(ReproducerTest.class);
>     public static final int THREADS_COUNT = 4;
>     public static final int NUM_ITERATIONS_PER_THREAD = 100;
>     public static final String TEST_LOCK_NAME = "testLockName";
>     @Test
>     void useLock_multithreaded() throws InterruptedException {
>         final LockByKey lockByKey = new LockByKey();
>         final CountDownLatch threadCountDownLatch = new CountDownLatch(THREADS_COUNT);
>         // Start multiple threads
>         for (int i = 0; i < THREADS_COUNT; i++) {
>             final int threadId = i;
>             new Thread(() -> {
>                 // Perform multiple iterations in each thread
>                 for (int iterationId = 0; iterationId < NUM_ITERATIONS_PER_THREAD; iterationId++) {
>                     logger.info("Running thread {}, iteration {}", threadId, iterationId);
>                     try {
>                         lockByKey.lock(TEST_LOCK_NAME);
>                     } finally {
>                         lockByKey.unlock(TEST_LOCK_NAME);
>                     }
>                 }
>                 threadCountDownLatch.countDown();
>             }).start();
>         }
>         // Wait for all threads to complete
>         threadCountDownLatch.await();
>     }
> }
> ```
> 
> which will sometimes hang and throw the following exception on the `lockWrapper.lock.unlock();` line:
> 
> ```
> java.lang.NullPointerException: Cannot read field "lock" because "lockWrapper" is null
> ```
> 
> The problem is that the `unlock` method doesn't synchronize decrementing the counter and removing the item from the map:
> 
> ```java
>     public void unlock(String key) {
>         LockWrapper lockWrapper = locks.get(key);
>         lockWrapper.lock.unlock();
>         if (lockWrapper.removeThreadFromQueue() == 0) { 
>             // Multiple threads can reach this line concurrently and each may attempt to remove the lock
>             // Calling Map#remove(Object key, Object value) as a workaround doesn't help because the same `lockWrapper` instance is reused between the different threads.
>             locks.remove(key, lockWrapper);
>         }
>     }
> ```
> 
> See [here](https://stackoverflow.com/a/77829625/3620458) for a variant that uses the `ConcurrentHashMap#computeIfPresent` method to decrement the counter and remove the item from the map in a synchronized manner. (Note that in this case you don't need an `AtomicInteger` for the counter, since the synchronization is done entirely outside the `LockWrapper`)

Hey @Bragolgirith thanks for taking the time to share your findings on the `LockByKey`, indeed there are a couple of issues in that original example and I ended up with a similar alteration.

PR #1842, opened 2025-04-26
Prevent or-mwc-menu from disrupting KPI attribute card layout (#1829)

### LINKED ISSUE #1829: Insights: KPI: time period selects don't overflow the widgets causing clunky behaviour
**Describe the bug**
The time period selection of the KPI open a or-mwc-menu that is very awkwardly placed. It pushes away the content, and is also partly hidden. The menu should open over and outside of the widget to find the space it needs.

**To Reproduce**
1. Create a KPI widget
2. Allow time range selection
3. In the widget change the time range, the menu pops up
4. The content is pushed away for the menu that is not placed correctly

**Expected behavior**
4. The menu opens over the widget bounds. Or find another appropriate solution

**Screenshots**

Image
Before clicking the time period:
Image
After click:
Image

OR 1.4.0

hamadhassan3:
## Description
This PR addresses the layout issue where the time period selection menu in the KPI widget (`or-attribute-card`) would disrupt the card content instead of appearing as an overlay.

To fix this:
1.  A new `fixedPosition` boolean option was added to the `getContentWithMenuTemplate` helper in `or-mwc-menu.ts`. When true, this applies `position: fixed` and a high `z-index` to the menu.
2.  This `fixedPosition: true` option was enabled for the period selector menu within `or-attribute-card.ts`.

This ensures the menu now correctly renders *over* the card content without disrupting the layout.

Fixes #1829

** Visuals **

**(Before) **
before

**(After) **
after1
after2

## Checklist


hamadhassan3:
Hi @DonWillems, I didn't encounter this issue during my testing. When I tested with the widget placed lower on the dashboard, it appeared as shown in the third screenshot in the PR description.

I'll test it again, including with the overflow removed as you suggested, and get back to you.

Thank you for having a look!

hamadhassan3:
Hi @DonWillems,

You're right that the solution breaks when the component is positioned near the bottom. I hadn't fully considered the side effects of using position: fixed in a scrollable container.

I also explored your suggestion to remove `overflow: hidden;`. However, after testing it out, I think this approach isn't sustainable in the long term. Its side effects are difficult to predict and could introduce UI inconsistencies over time (please see the image below).

Screenshot 2025-04-29 at 6 28 34 pm

After digging further into the codebase, I believe we should shift to a more flexible solution that doesn't interfere with layout flow. I’d recommend we use Portals to render the menu — this would let us dynamically position it with JavaScript at runtime without being constrained by parent styles.

Let me know what you think.

hamadhassan3:
Hi @MartinaeyNL, this looks promising—definitely worth exploring. I’ll test it out and let you know what I find.

hamadhassan3:
Hi @DonWillems, @MartinaeyNL,

As discussed earlier, I've updated this PR with an approach inspired by portals, incorporating insights from `or-mwc-dialog`. Implementing this for `or-mwc-menu` was a bit more challenging due to the nature of the component. That said, the current solution resolves all related issues, including overflow problems on smaller viewports, without introducing rigid structures.

I'd appreciate your feedback and a review of the updated implementation.

Please find below some screenshots of the updated view:
fix1
fix2
fix3

DonWillems:
Looks good to me, @MartinaeyNL ?

ui/component/or-attribute-card/src/index.ts
```diff
@@ -454,7 +454,14 @@ export class OrAttributeCard extends LitElement {
...
                                                 }),
                                                 this.period,
-                                                (value) => this._setPeriodOption(value)
+                                                (value) => this._setPeriodOption(value),
+                                                undefined,      // closedCallback
+                                                false,          // multiSelect
+                                                true,           // translateValues
+                                                false,          // midHeight
+                                                false,          // fullWidth
+                                                "kpi-menu-id",  // Menu Id
```
MartinaeyNL, line 463: Is there a need for a custom ID here? (since you're already using `ref()` for the element reference)

ui/component/or-mwc-components/src/or-mwc-menu.ts
```diff
@@ -51,20 +52,52 @@ declare global {
...
+
+export function getContentWithMenuTemplate(content: TemplateResult, menuItems: (ListItem | ListItem[] | null)[], selectedValues: string[] | string | undefined, valueChangedCallback: (values: string[] | string) => void, closedCallback?: () => void, multiSelect = false, translateValues = true, midHeight = false, fullWidth = false, menuId = "menu", fixedToHost = false): TemplateResult {
+    let menuRef: OrMwcMenu | null = null;   // Reference to the menu
 
     const openMenu = (evt: Event) => {
         if (!menuItems) {
             return;
         }
 
+        if (fixedToHost && menuRef) {
```
MartinaeyNL, line 85: Can you maybe explain in what cases `fixedToHost` needs to be used?
I know when it's set to `true`, it will use `position: fixed`, allowing overflow outside the container.
But there is probably a reason why it is `false` by default.

So a small code comment for explanation would be nice here.

MartinaeyNL:
I've tested the branch locally, and tried it out on my phone.
Is this behavior OK @DonWillems ?

image

ui/component/or-attribute-card/src/index.ts
hamadhassan3: Hi @MartinaeyNL , I used a custom ID here to avoid interfering with any existing behavior and to keep the changes scoped specifically to the KPI menu. This approach also gives us more flexibility to evolve each menu type independently in the future.

pierrekil:
@hamadhassan3 I dropped you an email some days ago. We appreciate all your work and to merge it we require a signed contributor license agreement. Maybe my email went into your spam ;-)

ui/component/or-mwc-components/src/or-mwc-menu.ts
MartinaeyNL: I noticed that, so good that you did haha

No I was just curious on **when/where** it should be used.
As in, "in what scenarios should this value be set to true?"

But the parameter is self-explanatory for me, so it's OK

ui/component/or-attribute-card/src/index.ts
MartinaeyNL: Yeah indeed. I agree on the flexibility value of modifying the element ID here.
For example using `querySelector()` in a consumer element.
Might not be necessary with these changes, but it doesn't hurt adding it. 👍

MartinaeyNL (approved):

Issue #1803, opened 2025-04-04
Slow loading time on /assets page with a lot of linked users


MartinaeyNL:
### Description
When an asset has many linked users (let's say 50+), the loading time of the asset page is very slow.
This is because the amount of HTTP requests for retrieving user roles is causing the UI to wait.
The fact that 50+ individual requests take place is not great in the first place, so this should be optimized.
Possibly by adding pagination to the table, or improving the HTTP endpoint?

### To reproduce
1. Create any type of asset
2. Link at least 50 users to that asset (permissions do not matter here)
3. View that specific asset on the /assets page. (maybe use a fresh browser window)
4. You have to wait at least 1,5 seconds before the page shows any asset data. (depending on system specs)

### Screenshots
Network logs when loading the assets page on one of our deployments;

<img src="https://github.com/user-attachments/assets/283a9952-56cc-4c61-aaa1-03f748267c8b" width="500"/>

### Expected behavior
Loading time of the full page should be much faster, less than 500 milliseconds.
Alternatively, some panels could be shown later than others. (using a loading indicator)

MartinaeyNL:
Resolved, thanks to PR #1856 🥳

PR #1856, opened 2025-05-07
Bug fix for slow loading on /assets page for many linked users

ncouture33:
## Description

/assets page takes a long time to load when there are many linked users. Pagination, an improved endpoint, or a loading indicator could be used to address this.

- Added new UserResource endpoint to retrieve batch user roles.
- Introduced UserRoles class to encapsulate role-related data in backend responses.
- Refactored frontend logic to make use of the new batch endpoint when fetching user roles.

This simplifies user role fetching by reducing multiple API calls to a single batch request, improving efficiency and addressing part of the performance concern.

While this improves the endpoint as suggested in the issue, it is a partial fix. Additional work (e.g. pagination, UI loading indicators) will be needed to fully resolve the problem.

#1803
## Checklist

### DIFF
--- a/manager/src/main/java/org/openremote/manager/security/UserResourceImpl.java
+++ b/manager/src/main/java/org/openremote/manager/security/UserResourceImpl.java
@@ -245,6 +245,34 @@ public class UserResourceImpl extends ManagerWebResource implements UserResource
     }
 
     @Override
+    public UserRoles[] getUsersRoles(RequestParams params, String realm, String clientId, String[] userIds) {
+        boolean hasAdminReadRole = hasResourceRole(ClientRole.READ_ADMIN.getValue(), Constants.KEYCLOAK_CLIENT_ID);
+        String me = getUserId();
+
+        for (String uId : userIds) {
+            if (!hasAdminReadRole && !Objects.equals(me, uId)) {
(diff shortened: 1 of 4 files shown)


ncouture33:
Note: This PR replaces PR #1855 which was automatically closed after a branch rename. Continuing the review here.

DonWillems:
Hi @ncouture33,

Thanks for your work! I've only now looked at the reported bug of the slow loading asset page and I would leave out the roles (not realm roles) altogether to improve the performance. In practice we do not see much added value in showing the roles of the linked user in that part of the interface. What do you think?

FYI, initially the roles column was added to show the composite roles of a user, instead of all roles (called permissions in the Manager UI). So a organisation using OpenRemote could have the roles (composite roles) e.g. "installer", "maintenance", "helpdesk", and you would see which role a linked user would have. But as mentioned, in practice the value of this column is not proven.

To be clear, I think the 'restricted user' column is useful.

ncouture33:
Hi @DonWillems,

Thanks for the feedback! Based on what you've said, I agree that removing the roles column makes sense for performance. I'll go ahead and make that change.

DonWillems:
Tested locally and looks good to me.

PR #1857, opened 2025-05-09
Rspack integration

MartinaeyNL:
## Description
[Rspack](https://rspack.dev) is a modern web bundler, that seamlessly replaces Webpack using a compatible API.
The compile times / performance is drastically better, and all modern features are included by default.

To migrate from Webpack to Rspack, I've simply followed the official guide from their [website](https://rspack.dev/guide/migration/webpack).
I kept the migration simple, and didn't migrate all Webpack plugins to prevent breaking code.
For example, it's using Webpack `ts-loader` like before, instead of replacing it with a RSPack-optimized plugin.

## Changelog
- Replaced Webpack with RSPack; renamed files from `webpack.config.js` to `rspack.config.js`,
Updated the NPM scripts to use `rspack` instead of the webpack counterpart. (the API is the same)
- Replaced Babel dependency with RSPack's built-in SWC loader. (It's a modern [drop-in replacement](https://swc.rs/docs/migrating-from-babel) for Babel).
- Bumped NPM package versions to `1.8.0-snapshot`

## Checklist


MartinaeyNL:
Updated description with the changelog.

I haven't found any issues during testing, nor were there problems on remote deployments.
The Webpack API is compatible with RSPack, so I'm not expecting any issues either.

Feel free to review and take a look, @Ekhorn and @richturner.
I'm happy to call / internally discuss if necessary.

ui/util/webpack.util.js
```diff
@@ -31,6 +28,7 @@ function getStandardModuleRules() {
                 test: /\.tsx?$/,
                 exclude: /node_modules/,
                 use: {
+                    // TODO: Switch to builtin:swc-loader, and remove ts-loader / webpack dependency
                     loader: "ts-loader",
                     options: {
                         projectReferences: true
```
Ekhorn, line 34: Does this not work without breaking code? Seems fairly simple? https://rspack.rs/guide/features/builtin-swc-loader#typescript-transpilation

MartinaeyNL: I didn't look fully into this, but on first sight it created a lot of errors actually 😂 
So yes, it was an proper decision to stick with `ts-loader`.

After a quick test, it looks like SWC doesn't support the `tsconfig.json` format,
so compiling with SWC throws errors for our use of `@customElement`,
since our setting of `experimentalDecorators: true` is not respected.

MartinaeyNL: I played around with this 3rd party plugin to create a SWC compatible config (see [tsconfig-to-swcconfig](https://github.com/Songkeys/tsconfig-to-swcconfig))
It worked initially, but SWC does [not support](https://github.com/swc-project/swc/discussions/2156) TypeScript composite project references,
so imports like "@openremote/model" cannot be resolved.

Ekhorn:
Tested the changes with the [update/frontend-testing](https://github.com/openremote/openremote/tree/update/frontend-testing) branch and it works!

richturner (approved):
Tried locally and all seems to work well, super quick as well!

PR #1793, opened 2025-04-01
Feature/k8s aws eks

ebariaux:
## Checklist

- [ ] 1. Acceptance criteria of the linked issue(s) are met
- [ ] 2. Tests are written and all tests pass
- [ ] 3. Changes are manually tested by you and the reviewer

PR #1876, opened 2025-05-15
Feature/k8s

ebariaux:
Supports MQTT(S), HAProxy and deployment to EKS


wborn:
Why is there also #1793 if this PR also has EKS changes?
Maybe one of them includes all changes of the other?
Please explain which PR has what changes and what should be the order in which they should be reviewed/merged.
If they depend on each other this should be mentioned in the first comment.

PR #1882, opened 2025-05-18
Asset type sorted in alphabetical order

pawanbhttr:
## Description
Asset type dropdown is sorted and displayed in ascending order

## Before Enhancement
image

## After Enhancement
image


MartinaeyNL (commented):
Thanks for the PR!

I've tested this locally, and I think this is a great addition!

Shouldn't we maybe apply this to all "allowed value constraints" ? I don't see why not.
For example, the `connection status` and `execution status` could then also be alphabetically ordered.
Same counts for custom asset types / implementations that others have done.
@DonWillems what do you think?
@DonWillems

DonWillems:
I think in most cases where you set the constraints manually (instead of the retrieved list of asset types) it is preferred to keep the order it is set in, because that makes sense for that application. So just for the asset types is fine for now.

MartinaeyNL (approved):
Thanks for signing the contributor agreement!

Let's get this merged 💪

Issue #1810, opened 2025-04-07
Account: user can not update their own data on the account page with write:user permission


DonWillems:
**Describe the bug**
The account page uses the method to update users where the write:admin permission is needed. It should be possible for a user with write:user to update its own information.
Later Keycloak settings should also be applied, see #1640 

**To Reproduce**
1. Log in as user with read:users and write:user permission
2. Go to account page
3. Fields will be readonly (because it checks for write:admin, as does the endpoint used for updating)

**Expected behavior**
3. Fields are enabled and the user can update their own data

**Additional context**
There is already a user/user GET method to retrieve the currently authenticated user. This should be extended with a PUT to also be able to update the current user with the requirement of WRITE_USER. Then we can use that on the Account page and make the readonly check based on write:user

hamadhassan3:
Hi @DonWillems, I have done some initial analysis here and would love to contribute. Working here would also help me understand the keyclock setup in more detail.

If you approve, I'll start analysing it in more detail.

DonWillems:
Hi Hamad,

Yes feel free to pick it up. If you have any questions don't hesitate to ask here or at [forum.openremote.io](https://forum.openremote.io)

Thank you,

Don

Miggets7:
In `headers.ts` the function`headerItemAccount` the `roles` property needs to be updated to `["write:user"]`
In `page-account.ts` line 179 need to be updated to `WRITE_USER` and `_getAccountRowTemplate` need to be updated to get realm information about `registrationEmailAsUsername` and also get the email of the user

Issue #1570, opened 2025-01-06
Account: add current password check when a user wants to change their password


DonWillems:
Currently when a user wants to change their password, they can do so on the Account page. This can be done without confirming your current password. This operation is sensitive enough to require an additional current password check so that it can not be done by someone else on an unattended open session. The operation is also infrequent enough that it is not that inconvenient.

- Change "Password" subheader to "Change password"
- Add "Current password" field
- Rename "password" to "new password"
- Rename "repeat password" to "repeat new password"

In the process check the current password before changing it to the new one.

Miggets7:
The user will receive an email with a "magic link" in order to update their password. This still doesn't require the user to fill in its current password, but does add some sort of 2fa by using email. So maybe that is the way to proceed.

richturner:
I tried it on staging demo and I got a login page and immediately after login I was asked to enter my new password

Miggets7:
So I would say maybe replace the inputs with a Reset Password button which sets the Keycloak action and the flow starts from Keycloak. What do you think @DonWillems ?

DonWillems:
Yes that makes sense!

MartinaeyNL:
Shared my findings on this in the linked PR #1887.

I think there's still some discussion needed on how to cope with different Keycloak configurations.
For example, how to handle users without a unique email address? What happens if no mail server is set up?
If an email could not be sent, would the user get feedback in the UI? etc.

Let me know, @Miggets7 if you need help on this.
Happy to discuss this further.

DonWillems:
@Miggets7 For now, only add the reset password button. If no email server is configured, or if the user does not have an email set, it will not work. That is a choice for now until we create a better plan for integrating the Keycloak account page.
If the user does not have an email address set, disable the reset password button and add a note "Can not send password reset email with an email address".

Miggets7:
Just discussed with @DonWillems and all the work is already done, so his last comment can be ignored.

PR #1887, opened 2025-05-21
New reset password flow

Miggets7:
openremote/openremote#1570

### DIFF
--- a/manager/src/main/java/org/openremote/manager/security/ManagerIdentityProvider.java
+++ b/manager/src/main/java/org/openremote/manager/security/ManagerIdentityProvider.java
@@ -58,6 +58,8 @@ public interface ManagerIdentityProvider extends IdentityProvider {
 
     void deleteUser(String realm, String userId);
 
+    void requestPasswordReset(String realm, String userId);
+
     void resetPassword(String realm, String userId, Credential credential);
 
     String resetSecret(String realm, String userId, String secret);
--- a/manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
+++ b/manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
@@ -31,6 +31,7 @@ import org.keycloak.admin.client.resource.RealmResource;
 import org.keycloak.admin.client.resource.UserResource;
 import org.keycloak.admin.client.resource.*;
 import org.keycloak.common.enums.SslRequired;
+import org.keycloak.models.UserModel;
 import org.keycloak.representations.idm.*;
 import org.openremote.container.message.MessageBrokerService;
 import org.openremote.container.persistence.PersistenceService;
@@ -452,6 +453,16 @@ public class ManagerKeycloakIdentityProvider extends KeycloakIdentityProvider im
             User.class,
             Collections.singletonList("attributes"),
             null);
+    }
+
+    @Override
+    public void requestPasswordReset(String realm, String userId) {
+        getRealms(realmsResource -> {
+            realmsResource.realm(realm).users().get(userId).executeActionsEmail(
+                    Collections.singletonList(UserModel.RequiredAction.UPDATE_PASSWORD.toString())
+            );
+            return null;
+        });
     }
 
     @Override
--- a/manager/src/main/java/org/openremote/manager/security/UserResourceImpl.java
+++ b/manager/src/main/java/org/openremote/manager/security/UserResourceImpl.java
@@ -167,6 +167,17 @@ public class UserResourceImpl extends ManagerWebResource implements UserResource
     }
 
     @Override
+    public void requestPasswordReset(RequestParams requestParams, String realm, String userId) {
+        try {
+            identityService.getIdentityProvider().requestPasswordReset(realm, userId);
+        } catch (ClientErrorException ex) {
+            throw new WebApplicationException(ex.getCause(), ex.getResponse().getStatus());
+        } catch (Exception ex) {
+            throw new WebApplicationException(ex);
+        }
+    }
+
+    @Override
     public void resetPassword(@BeanParam RequestParams requestParams, String realm, String userId, Credential credential) {
         try {
             identityService.getIdentityProvider().resetPassword(realm, userId, credential);
--- a/model/src/main/java/org/openremote/model/security/UserResource.java
+++ b/model/src/main/java/org/openremote/model/security/UserResource.java
@@ -102,6 +102,14 @@ public interface UserResource {
     void delete(@BeanParam RequestParams requestParams, @PathParam("realm") String realm, @PathParam("userId") String userId);
 
     @PUT
+    @Path("{realm}/request-password-reset/{userId}")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @RolesAllowed(Constants.WRITE_ADMIN_ROLE)
+    @Operation(operationId = "requestPasswordReset", summary = "Request a password reset for a user in a realm")
+    void requestPasswordReset(@BeanParam RequestParams requestParams, @PathParam("realm") String realm, @PathParam("userId") String userId);
+
+    @PUT
     @Path("{realm}/reset-password/{userId}")
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
--- a/ui/app/manager/src/pages/page-account.ts
+++ b/ui/app/manager/src/pages/page-account.ts
@@ -7,12 +7,12 @@ import {Store} from "@reduxjs/toolkit";
 import {Page, PageProvider, AppStateKeyed} from "@openremote/or-app";
 import {when} from "lit/directives/when.js";
 import {until} from "lit/directives/until.js";
-import {map} from 'lit/directives/map.js';
 import {guard} from "lit/directives/guard.js";
 import {i18next} from "@openremote/or-translate";
 import {InputType, OrInputChangedEvent, OrMwcInput} from "@openremote/or-mwc-components/or-mwc-input";
 import {ClientRole, Credential, Role, User, UserAssetLink, UserQuery} from "@openremote/model";
 import {showSnackbar} from "@openremote/or-mwc-components/or-mwc-snackbar";
+import {DialogAction, OrMwcDialog, showDialog} from "@openremote/or-mwc-components/or-mwc-dialog";
 
 export function pageAccountProvider(store: Store<AppStateKeyed>): PageProvider<AppStateKeyed> {
     return {
@@ -37,7 +37,7 @@ interface UserModel extends User {
 }
 
 @customElement("page-account")
-export class PageAccount extends Page<AppStateKeyed>  {
+export class PageAccount extends Page<AppStateKeyed> {
 
     protected static _getStyle(): CSSResult {
         return css`
@@ -47,6 +47,7 @@ export class PageAccount extends Page<AppStateKeyed>  {
                 display: flex;
                 flex-direction: column;
             }
+
             #title {
                 padding: 0 20px;
                 font-size: 18px;
@@ -63,6 +64,7 @@ export class PageAccount extends Page<AppStateKeyed>  {
                 margin-right: 10px;
                 margin-left: 14px;
             }
+
             .panel {
                 flex: 0;
                 width: 100%;
@@ -114,7 +116,7 @@ export class PageAccount extends Page<AppStateKeyed>  {
                 flex: 1 1 0;
                 gap: 20px;
             }
-            
+
             h5 {
                 margin-bottom: 0;
             }
@@ -149,15 +151,6 @@ export class PageAccount extends Page<AppStateKeyed>  {
     @state()
     protected _invalid = false;
 
-    @state()
-    protected _passwordPolicy: string[] = [];
-
-    @query("#new-password")
-    protected _passwordElem?: OrMwcInput;
-
-    @query("#new-repeatPassword")
-    protected _repeatPasswordElem?: OrMwcInput;
-
     static get styles() {
         return this._getStyle();
     }
@@ -168,11 +161,6 @@ export class PageAccount extends Page<AppStateKeyed>  {
 
     public stateChanged(_state: AppStateKeyed) {
     }
-
-    public connectedCallback() {
-        super.connectedCallback();
-        this._getPasswordPolicy();
-        }
 
     protected render(): TemplateResult | void {
 
@@ -197,22 +185,22 @@ export class PageAccount extends Page<AppStateKeyed>  {
                     <or-translate value="account">
                 </div>
                 <div id="content" class="panel">
-                    
+
                     <p class="panel-title">${i18next.t("user")} ${i18next.t("settings")}</p>
-                    
+
                     <!-- Account settings row -->
                     ${guard([this._user], () => until(
-                        this._getAccountRowTemplate(this._user, readonly, (_user, dirty, invalid) => {
-                            this._dirty = dirty;
-                            this._invalid = invalid;
-                        })
+                            this._getAccountRowTemplate(this._user, readonly, (_user, dirty, invalid) => {
+                                this._dirty = dirty;
+                                this._invalid = invalid;
+                            })
                     ))}
-                    
+
                     <!-- Actions row (such as the save button) -->
                     ${when(this._user, () => until(this._getActionsRowTemplate(this._user)))}
-                    
+
                 </div>
-                
+
                 ${when(manager.isKeycloak(), () => html`
                     <div class="panel">
                         <p class="panel-title">
@@ -220,8 +208,9 @@ export class PageAccount extends Page<AppStateKeyed>  {
                         </p>
                         <div class="row">
                             <div class="column">
-                                <or-mwc-input .type="${InputType.BUTTON}" label="${i18next.t('twoFactorConfigure')}" outlined
-                                              @or-mwc-input-changed="${() => manager.login({ action: "CONFIGURE_TOTP" })}"
+                                <or-mwc-input .type="${InputType.BUTTON}" label="${i18next.t('twoFactorConfigure')}"
+                                              outlined
+                                              @or-mwc-input-changed="${() => manager.login({action: "CONFIGURE_TOTP"})}"
                                 ></or-mwc-input>
                             </div>
                         </div>
@@ -237,7 +226,7 @@ export class PageAccount extends Page<AppStateKeyed>  {
      */
     protected async _getAccountRowTemplate(user?: UserModel, readonly = true, onchange?: (user: User, dirty: boolean, invalid: boolean) => void): Promise<TemplateResult> {
         const registrationEmailAsUsername = true;
-        if(!user) {
+        if (!user) {
             user = await this._getUser();
         }
         return html`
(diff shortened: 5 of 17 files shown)


MartinaeyNL:
As discussed internally in voice calls etc;
Resetting a password for a user without a unique email address should be impossible.

Also important to mention;
Because we have our own account page (instead of using Keycloak their account page),
we have to implement every piece of functionality ourselves. So, currently, it does not align with Keycloak configuration at all.
I assume Keycloak has configuration options for resetting password as well.
It's probably difficult to align with the Keycloak configuration, so we should consider if we'd like to force this behavior or not.

richturner:
@Miggets7 @MartinaeyNL Please discuss follow up actions on this one

richturner:
@DonWillems @MartinaeyNL can we look to get this finished and merged?

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -454,6 +455,16 @@ public void deleteUser(String realm, String userId) {
...
+    @Override
+    public void requestPasswordReset(String realm, String userId) {
+        getRealms(realmsResource -> {
+            realmsResource.realm(realm).users().get(userId).executeActionsEmail(
+                    Collections.singletonList(UserModel.RequiredAction.UPDATE_PASSWORD.toString())
+            );
+            return null;
+        });
+    }
+
```
MartinaeyNL: How does Keycloak handle users without an email address?
And if multiple users have the same email address, how does it cope with the password reset email?
In my opinion, it should be impossible to request password resets for users without a unique email address.

MartinaeyNL: Should we check if a mail servers is properly set up?
Because I can imagine users would be confused by pressing a "reset password" button, and nothing happens at all.

model/src/main/java/org/openremote/model/security/UserResource.java
```diff
@@ -101,6 +101,14 @@ public interface UserResource {
...
     void delete(@BeanParam RequestParams requestParams, @PathParam("realm") String realm, @PathParam("userId") String userId);
 
+    @PUT
+    @Path("{realm}/request-password-reset/{userId}")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @RolesAllowed(Constants.WRITE_ADMIN_ROLE)
+    @Operation(operationId = "requestPasswordReset", summary = "Request a password reset for a user in a realm")
+    void requestPasswordReset(@BeanParam RequestParams requestParams, @PathParam("realm") String realm, @PathParam("userId") String userId);
+
```
MartinaeyNL: This endpoint is for `WRITE_ADMIN` only. How would this work with a user their account page?
As in, not all users would be admins, and they'd like to request a password reset too.
Should we add a separate endpoint?

(not sure if the account page is accessible for users at all, might be a product-related decision here)

ui/app/manager/src/pages/page-account.ts
MartinaeyNL, line 1: I think there will be cases we need "new password / repeat password" fields.
For example, if no mail server is set up, or the user doesn't have an email address, you need an alternative.
(I'll leave basic authentication out of the question, since that's rarely used)
So I think we shouldn't remove them here.

manager/src/main/java/org/openremote/manager/security/UserResourceImpl.java
```diff
@@ -166,6 +166,17 @@ public void delete(RequestParams requestParams, String realm, String userId) {
...
+    @Override
+    public void requestPasswordReset(RequestParams requestParams, String realm, String userId) {
+        try {
+            identityService.getIdentityProvider().requestPasswordReset(realm, userId);
+        } catch (ClientErrorException ex) {
+            throw new WebApplicationException(ex.getCause(), ex.getResponse().getStatus());
+        } catch (Exception ex) {
+            throw new WebApplicationException(ex);
+        }
+    }
```
MartinaeyNL: We need some condition checks here to validate if the endpoint payload is valid.
Things like "is realm accessible by the authenticated user", and so on.

manager/src/main/java/org/openremote/manager/security/ManagerIdentityProvider.java
MartinaeyNL, line 1: We should really add tests for validating the functionality here.

MartinaeyNL (commented):
Summarized my findings I shared earlier, together with additional comments, in this review here.

I'm happy to call and/or discuss things I shared here.
Maybe having a check with @DonWillems, @richturner, and myself would also help to clear things up.

model/src/main/java/org/openremote/model/security/UserResource.java
richturner: This should all be defined in the requirements of the issue.
If this piece of work is for anyone to request a password reset (non-authenticated users) then it would need to be a public endpoint

Miggets7:
Not yet ready though, there's a caching mechanism in `ManagerKeycloakIdentityProvider` which only works if we update the realm through the manager and not when updating in Keycloak admin console

Miggets7:
@MartinaeyNL @richturner should be ready now to be reviewed

Miggets7:
Fix #1810

manager/src/main/java/org/openremote/manager/security/ManagerIdentityProvider.java
```diff
@@ -346,7 +349,12 @@ static List<String> getUserIds(PersistenceService persistenceService, String rea
...
         return persistenceService.doReturningTransaction(entityManager -> {
             List<Realm> realms = (List<Realm>) entityManager.createNativeQuery(
-                    "select *, (select ra.VALUE from PUBLIC.REALM_ATTRIBUTE ra where ra.REALM_ID = r.ID and ra.name = 'displayName') as displayName from public.realm r  where r.not_before is null or r.not_before = 0 or r.not_before <= extract('epoch' from now())"
+                    "select *, " +
+                    "(select ra.VALUE from PUBLIC.REALM_ATTRIBUTE ra where ra.REALM_ID = r.ID and ra.name = 'displayName') as displayName, " +
+                    "(select rsc.VALUE from PUBLIC.REALM_SMTP_CONFIG rsc where rsc.REALM_ID = r.ID and rsc.name = 'host') as smtpHost, " +
+                    "(select rsc.VALUE from PUBLIC.REALM_SMTP_CONFIG rsc where rsc.REALM_ID = r.ID and rsc.name = 'port') as smtpPort, " +
+                    "(select rsc.VALUE from PUBLIC.REALM_SMTP_CONFIG rsc where rsc.REALM_ID = r.ID and rsc.name = 'user') as smtpUser " +
+                    "from public.realm r where r.not_before is null or r.not_before = 0 or r.not_before <= extract('epoch' from now())"
```
richturner: Don't think this change is needed anymore

manager/src/main/java/org/openremote/manager/security/RealmView.java
```diff
@@ -43,6 +43,12 @@ public class RealmView {
 
     protected String displayName;
 
+    protected String smtpHost;
+
+    protected String smtpPort;
+
+    protected String smtpUser;
+
```
richturner: Not needed anymore?

model/src/main/java/org/openremote/model/security/Realm.java
```diff
@@ -127,6 +124,15 @@ public List<String> deserialize(JsonParser p, DeserializationContext ctxt) throw
...
 
+    @Formula("(select rsc.VALUE from PUBLIC.REALM_SMTP_CONFIG rsc where rsc.REALM_ID = ID and rsc.name = 'host')")
+    protected String smtpHost;
+
+    @Formula("(select rsc.VALUE from PUBLIC.REALM_SMTP_CONFIG rsc where rsc.REALM_ID = ID and rsc.name = 'port')")
+    protected String smtpPort;
+
+    @Formula("(select rsc.VALUE from PUBLIC.REALM_SMTP_CONFIG rsc where rsc.REALM_ID = ID and rsc.name = 'user')")
+    protected String smtpUser;
+
```
richturner: Not needed anymore?

test/src/test/groovy/org/openremote/test/users/UserResourceTest.groovy
```diff
@@ -292,13 +292,45 @@ class UserResourceTest extends Specification implements ManagerContainerTrait {
...
     }
+    
+    def "Self update user with WRITE_USER role"() {
+        when: "a regular user gets their own information"
+        // First get the user's ID from the authentication context
+        def users = regularUserBuildingResource.query(null, new UserQuery().realm(new RealmPredicate(keycloakTestSetup.realmBuilding.name)))
+        def currentUser = users.find { it.username == "testuser3" }
+        
+        then: "the user should be found"
+        currentUser != null
```
richturner: `KeycloakTestSetup.testUser3` variable can be used.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -94,10 +99,10 @@ public class ManagerKeycloakIdentityProvider extends KeycloakIdentityProvider im
...
     public static final String OR_KEYCLOAK_ENABLE_DIRECT_ACCESS_GRANT = "OR_KEYCLOAK_ENABLE_DIRECT_ACCESS_GRANT";
     public static final List<String> BUILT_IN_REALM_ROLES = List.of(
-        "admin",
-        "create-realm",
-        "offline_access",
-        "uma_authorization"
+            "admin",
+            "create-realm",
+            "offline_access",
+            "uma_authorization"
```
richturner: Want to avoid these formatting changes, they increase risk of merge conflicts and just go back and forth depending on people's settings. Will soon introduce code styling standards

test/src/test/groovy/org/openremote/test/users/UserResourceTest.groovy
```diff
@@ -292,13 +292,45 @@ class UserResourceTest extends Specification implements ManagerContainerTrait {
...
+        def updatedUser = regularUserBuildingResource.update(null, currentUser)
+        
+        then: "the update should succeed"
+        updatedUser != null
+        updatedUser.attributes.any { it.name == "selfUpdate" && it.value == "success" }
+        
+        when: "the user tries to update another user's information"
+        def otherUser = adminUserResource.get(null, keycloakTestSetup.realmBuilding.name, keycloakTestSetup.testuser2Id)
+        otherUser.setAttribute("unauthorizedUpdate", "shouldFail")
+        regularUserBuildingResource.update(null, otherUser)
```
richturner: @DonWillems shouldn't admin users be able to alter other users info?

test/src/test/groovy/org/openremote/test/users/UserResourceTest.groovy
```diff
@@ -292,13 +292,45 @@ class UserResourceTest extends Specification implements ManagerContainerTrait {
...
-        user = adminUserResource.update(null, keycloakTestSetup.realmBuilding.name, user)
+        user = adminUserResource.updateUser(null, keycloakTestSetup.realmBuilding.name, user)
 
         then: "the update should have succeeded"
         user.attributes.size() == 2
         user.attributes.any {it.name == EMAIL_NOTIFICATIONS_DISABLED_ATTRIBUTE && it.value == "true"}
         user.attributes.any {it.name == "test" && it.value == "testvalue"}
     }
+    
+    def "Self update user with WRITE_USER role"() {
```
richturner: Don't know how this test got into this PR as it is nothing to do with the linked issue, maybe a missing bug that should be linked?
Also I don't see tests for the functionality mentioned in this issue, even if it's just checking security logic and that correct method is called with the use of `Spy()`

ui/app/manager/src/pages/page-account.ts
```diff
@@ -197,31 +189,32 @@ export class PageAccount extends Page<AppStateKeyed>  {
...
                     <p class="panel-title">${i18next.t("user")} ${i18next.t("settings")}</p>
-                    
+
                     <!-- Account settings row -->
                     ${guard([this._user], () => until(
-                        this._getAccountRowTemplate(this._user, readonly, (_user, dirty, invalid) => {
-                            this._dirty = dirty;
-                            this._invalid = invalid;
-                        })
+                            this._getAccountRowTemplate(this._user, readonly, (_user, dirty, invalid) => {
```
richturner, line 207: Remove formatting changes (see other comment on this subject)

ui/app/manager/src/pages/page-account.ts
```diff
@@ -236,8 +229,10 @@ export class PageAccount extends Page<AppStateKeyed>  {
      * Will call {@link onchange} when an input field submits a new value.
      */
     protected async _getAccountRowTemplate(user?: UserModel, readonly = true, onchange?: (user: User, dirty: boolean, invalid: boolean) => void): Promise<TemplateResult> {
-        const registrationEmailAsUsername = true;
-        if(!user) {
+        const realmResponse = await manager.rest.api.RealmResource.get(manager.displayRealm);
+        const registrationEmailAsUsername = realmResponse.data.registrationEmailAsUsername;
+        const smtpConfigured = realmResponse.data.smtpUser && realmResponse.data.smtpHost && realmResponse.data.smtpPort;
```
richturner: @DonWillems I thought we discussed not doing this check anymore? and just check `registrationEmailAsUsername` what if this is false but user has an email address configured?

manager/src/main/java/org/openremote/manager/security/UserResourceImpl.java
richturner: @Miggets7 please don't forget @MartinaeyNL comment here about checking that the user belongs to the requested realm unless they are super admin

test/src/test/groovy/org/openremote/test/users/UserResourceTest.groovy
richturner: I see the bug is now linked but testing new fucntionality still stands (for security issues)

test/src/test/groovy/org/openremote/test/users/UserResourceTest.groovy
```diff
@@ -292,13 +292,56 @@ class UserResourceTest extends Specification implements ManagerContainerTrait {
...
+        when: "a regular user gets their own information"
+        // First get the user's ID from the authentication context
+        def users = regularUserBuildingResource.query(null, new UserQuery().realm(new RealmPredicate(keycloakTestSetup.realmBuilding.name)))
```
richturner: Get user object from `KeycloakTestSetup`

richturner (changes requested):
Still quite a few code styling changes, I find showing changes between branches in IDE useful for stepping through and confirming each change made.

Also no test for reset password endpoint to at least make sure it throws exceptions when incorrect user. Also my comment about admin resource (i.e. admin user) should be able to modify a user please confirm with Don if you are in the office together.

test/src/test/groovy/org/openremote/test/users/UserResourceTest.groovy
DonWillems: Yes, admins should be able to do that (e.g. using the Users page)

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
richturner: As we discussed in the office we don't want to become more tightly coupled to keycloak config so it is up to the instance owner to ensure correct config

test/src/test/groovy/org/openremote/test/notification/NotificationTest.groovy
```diff
@@ -615,21 +615,14 @@ class NotificationTest extends Specification implements ManagerContainerTrait {
         }
         notificationService.notificationHandlerMap.put(emailNotificationHandler.getTypeName(), mockEmailNotificationHandler)
 
-        expect: "the demo users to be created"
-        conditions.eventually {
-            def users = identityService.getIdentityProvider().queryUsers(new UserQuery().realm(new RealmPredicate(keycloakTestSetup.realmBuilding.name)).serviceUsers(false))
-            assert users.size() == 3
-            assert users.count { !TextUtil.isNullOrEmpty(it.email)} == 3
-        }
```
MartinaeyNL, line 623: Why was this test removed? I assume increasing `3` to `4` would solve the issue here?

ui/app/manager/src/pages/page-account.ts
```diff
@@ -348,52 +353,58 @@ export class PageAccount extends Page<AppStateKeyed>  {
...
             showSnackbar(undefined, "saveUserFailed");
         }
     }
 
+    /**
+     * HTML callback function for when the 'Reset Password' button is clicked.
+     */
+    protected _onResetPasswordBtnClick(_e: OrInputChangedEvent) {
+        if (this._user && !this._isInvalid()) {
+            this._requestPasswordReset()
```
MartinaeyNL: The HTTP request for resetting the password is made before the popup/modal opens?
This should be done once "OK" is pressed.

Maybe also include a small snackbar message using `showSnackbar()` with a text like "Password reset sent!"
(and "Failed to send password reset" if an error is thrown)

ui/app/manager/src/pages/page-account.ts
```diff
@@ -439,91 +450,140 @@ export class PageAccount extends Page<AppStateKeyed>  {
...
+                showSnackbar(undefined, "saveUserSucceeded");
             }
-        })
+            
+            // Update the stored user with the saved version and reset dirty flag
+            this._user = {...user};
+            this._dirty = false;
+            
+            // Force re-render to update UI elements (like the reset password button)
+            this.requestUpdate();
```
MartinaeyNL: Looking through the code, I don't see any reason that this `requestUpdate()` is necessary.
You're updating the variable in full, so the lifecycle hook is called correctly.
Let me know if I'm wrong.

ui/app/manager/src/pages/page-account.ts
MartinaeyNL: Calling this function needs an exception handler. (as the function itself doesn't handle errors)

model/src/main/java/org/openremote/model/security/UserResource.java
```diff
@@ -101,13 +109,37 @@ public interface UserResource {
...
+    @Operation(operationId = "requestPasswordReset", summary = "Request a password reset for the currently authenticated user")
+    void requestPasswordResetCurrent(@BeanParam RequestParams requestParams);
+
     @PUT
     @Path("{realm}/reset-password/{userId}")
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
     @RolesAllowed(Constants.WRITE_ADMIN_ROLE)
-    @Operation(operationId = "resetPassword", summary = "Reset the password for a user in a realm")
-    void resetPassword(@BeanParam RequestParams requestParams, @PathParam("realm") String realm, @PathParam("userId") String userId, Credential credential);
```
MartinaeyNL, line 110: We should remember ourselves to add this breaking API change in the release notes. ✍️ 
Or do we still want to support this?

MartinaeyNL:
No other comments on the API / backend related code 👍 
See my comments above on some frontend remarks.

test/src/test/groovy/org/openremote/test/notification/NotificationTest.groovy
richturner: It's a notification test no need to test for setup as ContainerTrait now ensures `startContainer()` waits for all services to be started

ui/app/manager/src/pages/page-account.ts
richturner: I haven't fully reviewed the UI code changes that Michael made

model/src/main/java/org/openremote/model/security/UserResource.java
richturner: It's not a breaking change, endpoint is still the same just method name changed for consistency

MartinaeyNL: Oh wait, my bad... You are correct.
I completely misunderstood the GitHub review viewer 😂

PR #1890, opened 2025-05-22
Update/frontend testing

### LINKED ISSUE #1815: Dynamic JSON Forms: Create frontend testing POC
## User Story
As a developer I want to know how to share a JSON Schema between the `or-json-forms` component and test tooling, so I can test the JSON forms declaratively through a JSON Schema.

## Description
The `or-json-forms`

Previous work on frontend testing has been conducted in 

https://github.com/openremote/openremote/tree/develop/frontend_testing

## Acceptance Criteria
- [ ] 1. [Criteria]

## Implementation Notes
[Relevant technical notes]

Ekhorn:
## Description

Closes #1815 

## Checklist


Ekhorn:
Hey @MartinaeyNL,

I updated the frontend testing documentation, which explains the test setup structure, how to write frontend tests, etc. See https://github.com/openremote/documentation/pull/89

Let me know if there's something to add or change :smile:

.github/workflows/ci_cd.yml
```diff
@@ -449,7 +453,7 @@ jobs:
               echo "pushRequired=true" >> $GITHUB_OUTPUT
               echo "managerDockerImage=openremote/manager:$FIRST_MANAGER_TAG" >> $GITHUB_OUTPUT
             else
-              command="docker build --build-arg GIT_COMMIT=$commitSha --platform linux/amd64,linux/aarch64 -t openremote/manager:$commitShaShort $buildPath"
+              command="docker build --load --build-arg GIT_COMMIT=$commitSha --platform linux/amd64 -t openremote/manager:$commitShaShort $buildPath && docker build --build-arg GIT_COMMIT=$commitSha --platform linux/aarch64 -t openremote/manager:$commitShaShort $buildPath"
```
richturner: What is this change for?

.github/workflows/ci_cd.yml
```diff
@@ -429,10 +429,14 @@ jobs:
...
-        id: test-ui-command
-        if: ${{ steps.skip-ui-tests.outputs.value != 'true' }}
-        run: echo ""
+      - name: Define UI component test command
+        id: test-ui-components-command
+        if: ${{ steps.skip-ui-component-tests.outputs.value != 'true' }}
+        run: echo "value=./gradlew -p ui/component npmTest" >> $GITHUB_OUTPUT
+
+      - name: Define UI end-to-end test command
+        id: e2e-test-command
```
richturner: Can we change these naming conventions please:

`test-ui-components`
`test-ui-apps`

`e2e-test-command` doesn't mention UI and also isn't very descriptive

project.gradle
```diff
@@ -353,6 +353,10 @@ task npmTest(type: Exec) {
     dependsOn getYarnInstallTask()
     commandLine npmCommand("yarn"), "run", "test"
 }
+task npmTestUI(type: Exec) {
+    dependsOn getYarnInstallTask()
+    commandLine npmCommand("yarn"), "run", "test", "--ui"
+}
```
richturner: Don't know why we need `UI` in here should just be running node tests, components like `core` don't have UI but they should still be tested

.github/workflows/ci_cd.yml
```diff
@@ -429,10 +429,14 @@ jobs:
...
-      - name: Define UI test command
-        id: test-ui-command
-        if: ${{ steps.skip-ui-tests.outputs.value != 'true' }}
-        run: echo ""
+      - name: Define UI component test command
+        id: test-ui-components-command
+        if: ${{ steps.skip-ui-component-tests.outputs.value != 'true' }}
+        run: echo "value=./gradlew -p ui/component npmTest" >> $GITHUB_OUTPUT
+
+      - name: Define UI end-to-end test command
```
richturner: These UI app tests run on every run of the CI/CD action?

.github/workflows/ci_cd.yml
Ekhorn: I need to load the manager docker image that is built, to ensure I can start the manager container with the right version for end-to-end testing.

I might be able to also just use the local production built directly instead of the container.

project.gradle
Ekhorn: The `--ui` flag is related to Playwright. It allows you to use [Playwright UI mode](https://playwright.dev/docs/test-ui-mode). I could remove it if you think it is redundant. Cd'ing into the right directory and running `npm test -- --ui` also works.

.github/workflows/ci_cd.yml
Ekhorn: Kind off, when there are backend and frontend changes, because either could break the manager.

richturner:
@Ekhorn can we save any of the old UI testing branch:

https://github.com/openremote/openremote/pull/733

A lot of time was spent defining the tests in there.

Ekhorn:
~~Looking at integrating with https://github.com/vitalets/playwright-bdd so we can have the best of worlds from #733 (cucumber-js) and the features playwright provides.~~ (Maybe for fun in my off time ;)

ui/test/config/playwright/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width,initial-scale=1.0" />
+    <title>Lit Web Components Test App</title>
```
MartinaeyNL: I'd rename this to something OpenRemote related.

ui/test/config/playwright/index.html
MartinaeyNL, line 1: Isn't it common to use a different folder name than `/playwright` ? Something like `/.playwright` or `/util` ?
Because the playwright configuration files (`component.ts` and `app.ts`) are in the parent folder.

However, I think sticking to what Playwright practices is best here.

ui/test/config/app.ts
MartinaeyNL, line 1: I'd prefer to rename this file to something like `app.config.ts` and `component.config.ts`,
since they're different environments with a similar config structure.
The name would also with identifying, since Webpack and other libraries use similar file names.
However, I'm not sure what Playwright normally practices.

ui/component/or-json-forms/playwright-ct.config.ts
MartinaeyNL, line 1: Can't we use the default `playwright.config.ts` file name, or doesn't that work for identifying component tests?

ui/app/build.gradle
MartinaeyNL, line 1: Is this file necessary? Other npm commands in Gradle work fine without a `build.gradle` in the component directory.

ui/app/manager/test/assets.test.ts
MartinaeyNL, line 1: I'm unfamiliar with what the standards of frontend testing look like, but a comment above each line sounds overdone.
I would expect, without any Playwright experience, it would look like this;
```typescript
/**
 * "Adding a new asset through the Assets page"
 * This test will login to OpenRemote, navigate to the Assets page, and create an asset through the Asset Tree.
 * After it's created, we try to update attribute values, try to disable/enable the asset, and then presses "Save".
 */
 test(`Add new asset: ${name}`, async ({ page, manager, assetsPage, components }) => {
     // ...
 }
```
Because that would increase readability by a ton. (in my opinion)
Don't know if any annotations can be useful for browsing through all tests in the repository?

*Also with this comment; take the default Playwright standards into count.*
*Your pattern might be common, and my comment doesn't make sense at all.*

ui/app/manager/test/fixtures/data/assets.ts
```diff
@@ -0,0 +1,137 @@
...
+      efficiencyExport: { name: "efficiencyExport", type: "positiveInteger" },
+      power: { name: "power", type: "number", meta: { readOnly: false } },
+      supportsImport: { name: "supportsImport", type: "boolean" },
+      powerImportMax: { name: "powerImportMax", type: "positiveNumber" },
+      energyImportTotal: { name: "energyImportTotal", type: "positiveNumber", meta: { readOnly: true } },
+      forceCharge: { name: "forceCharge", type: "executionStatus" },
+      energyExportTotal: { name: "energyExportTotal", type: "positiveNumber", meta: { readOnly: true } },
+      energyLevelPercentageMax: { name: "energyLevelPercentageMax", type: "positiveInteger" },
+      powerExportMax: { name: "powerExportMax", type: "positiveNumber" },
+      location: { name: "location", type: "GEO_JSONPoint" },
```
MartinaeyNL: Is it necessary to supply all the attributes?
Because I'd imagine you can filter it down to the attributes used in the tests.

I would like to prevent that these asset types are refactored / changed in the backend, whereafter this test breaks.

ui/app/manager/test/fixtures/data/assets.ts
```diff
@@ -0,0 +1,137 @@
...
+export type DefaultAssets = typeof assets;
+export default assets;
+
+export const thing: Asset = {
+  name: "Thing",
+  realm: "smartcity",
+  type: "ThingAsset",
+  attributes: {
+    notes: { name: "notes", type: "text", meta: {}, value: null, timestamp: 1748772049557 },
+    location: { name: "location", type: "GEO_JSONPoint", meta: {}, value: null, timestamp: 1748771034882 },
```
MartinaeyNL: Do we need timestamps to be supplied? Or is that part of the test?

ui/app/manager/test/fixtures/manager.ts
```diff
@@ -0,0 +1,489 @@
...
+      }
+    }
+  }
+
+  protected getAppUrl(realm: string) {
+    return `${new URL(this.baseURL).origin}/manager/?realm=${realm}`;
+  }
+}
+
+class AssetsPage {
```
MartinaeyNL: Maybe we should move these "page specific" ones to a separate file.
Something like `/test/fixtures/pages/assets` and `/test/fixtures/pages/realms` ?

ui/app/manager/test/fixtures/manager.ts
MartinaeyNL, line 1: I would make the comments on top of the functions more descriptive.
As in, describe in more detail what the test does, and what would happen upon errors for example.
Also, every function should have a description IMO.

It looks like they're simple utility functions, that continue the synchronized test that is running.
So I assume that any `expect()` function will throw an error, which gets caught by the test that is consuming this?
However, at other places it will write a `console.error()`, without a way for a function consumer to know it failed.
Right? or am I wrong here?

ui/app/manager/test/fixtures/manager.ts
```diff
@@ -0,0 +1,489 @@
...
+  public rules: number[] = [];
+
+  constructor(readonly page: Page, readonly baseURL: string) {
+    this.managerHost = process.env.managerUrl || "http://localhost:8080";
+    rest.initialise(`${this.managerHost}/api/master/`);
+    this.axios = rest.axiosInstance;
+  }
+
+  async goToRealmStartPage(realm: string) {
+    await this.page.goto(this.getAppUrl(realm));
```
MartinaeyNL, line 37: Out of curiosity; where does the `this.page` come from? I don't see any fields or inherited classes.
Or is this Manager class injected like a mixin into the Playwright runtime?

ui/app/manager/test/fixtures/manager.ts
```diff
@@ -0,0 +1,489 @@
...
+    data.append("grant_type", "password");
+    const { access_token } = (
+      await this.axios.post(`${this.managerHost}/auth/realms/${realm}/protocol/openid-connect/token`, data, {
+        headers: {
+          "Content-Type": "application/x-www-form-urlencoded",
+        },
+      })
+    ).data;
+    return access_token;
+  }
```
MartinaeyNL, line 111: Since we're not really using basic authentication anymore, using a Keycloak-only test is not too big of a concern.
But instead of using an Axios call, can't we use utility functions from `or-app` and `core`, like `.getKeycloakToken()` ?
It would automatically cover this function, without the necessity to build custom functions like this.
Don't know if it's possible for other utility functions as well.

ui/app/manager/test/fixtures/manager.ts
```diff
@@ -0,0 +1,489 @@
...
+   * @param user Realm user to create
+   * @param user Role to create
+   * @param user Assets to create
+   */
+  async setup(
+    realm: string,
+    { user, role, assets }: { user?: UserModel; role?: Role; assets?: Asset[] | DefaultAssets } = {}
+  ) {
+    const access_token = await this.getAccessToken("master", admin.username, admin.password);
+    const config = { headers: { Authorization: `Bearer ${access_token}` } };
```
MartinaeyNL: As additional note to my previous comment, there is a `.getAuthorizationHeader()` in `core`.

ui/app/manager/test/fixtures/manager.ts
```diff
@@ -0,0 +1,489 @@
...
+            expect(response.status).toBe(204);
+          } catch (e) {
+            console.error("Failed to reset user password", e.response.status);
+          }
+        } catch (e) {
+          console.error("Failed to update users' roles", e.response.status);
+        }
+      } catch (e) {
+        console.error("Failed to create user", e.response.status);
+      }
```
MartinaeyNL: I think we should split these requests to separate functions. Allows reusability as well.

ui/app/manager/test/fixtures/manager.ts
```diff
@@ -0,0 +1,489 @@
...
+    return `${new URL(this.baseURL).origin}/manager/?realm=${realm}`;
+  }
+}
+
+class AssetsPage {
+  constructor(private readonly page: Page, private readonly shared: Shared, private readonly manager: Manager) {}
+
+  async goto() {
+    this.manager.navigateToTab("Assets");
+  }
```
MartinaeyNL: Looks like every page has a `goto()` function, which is a good practice.
Maybe we can make an interface of it? (so that every page will have common function names and behavior)

ui/app/manager/test/insights.test.ts
MartinaeyNL, line 1: I guess we can remove this for the time being.

ui/app/manager/test/rules.test.ts
```diff
@@ -0,0 +1,161 @@
...
+});
+
+const energyRule = {
+  name: "Energy rule",
+  attribute_type: "Electricity battery asset",
+  asset: "Battery",
+  attribute_when: "Energy level",
+  attribute_then: "Energy capacity",
+  value: 50,
+};
```
MartinaeyNL: This should be in a fixture file, right?

ui/component/or-icon/test/or-icon.test.ts-snapshots/Should-render-mdi-icon-chart-areaspline-1-or-icon-linux.png
MartinaeyNL, line 1: Not a big fan of this folder name 😅

ui/component/or-json-forms/test/or-json-forms.test.ts
```diff
@@ -0,0 +1,139 @@
+import { ct } from "@openremote/test";
+
+import { OrJSONForms, StandardRenderers } from "@openremote/or-json-forms";
+
+const schemas = [
```
MartinaeyNL: Shouldn't this schema be in a fixture folder or similar?

ui/component/or-map/src/mapwidget.ts
```diff
@@ -992,4 +997,26 @@ export class MapWidget {
     protected _onGeocodeChange(geocode:any) {
         this._mapContainer.dispatchEvent(new OrMapGeocoderChangeEvent(geocode));
     }
+
+    // Source: maplibre.org/maplibre-gl-js/docs/examples/check-for-support/
+    protected isWebglSupported() {
+        if (window.WebGLRenderingContext) {
+            const canvas = document.createElement('canvas');
```
MartinaeyNL, line 1005: Nice touch!

We should build this `canvas` element within the mapwidget component, to prevent weird behavior outside of it.
Maybe `this.shadowRoot?.createElement()` would already work?

ui/test/config/app.ts
```diff
@@ -0,0 +1,73 @@
...
+  //   name: "firefox",
+  //   use: { ...devices["Desktop Firefox"] },
+  // },
+  // {
+  //   name: "webkit",
+  //   use: { ...devices["Desktop Safari"] },
+  // },
+];
+
+function createAppSetupAndTeardown(app: string) {
```
MartinaeyNL: Can you add a comment here on what this function actually does?
I have an idea, but it would be nice to know what Playwright does with this.

ui/test/config/app.ts
```diff
@@ -0,0 +1,73 @@
...
+export default (path: string) => {
+  const name = path.split("/").at(-1);
+  if (!name) return;
+
+  return defineConfig({
+    testMatch: "*.test.ts",
+    /* Fail the build on CI if you accidentally left test.only in the source code. */
+    forbidOnly: Boolean(CI),
+    /* Retry on CI only */
+    retries: CI ? 2 : 0,
```
MartinaeyNL: Do we want to retry the tests? Is this about retrying individual UI tests, or the whole UI stack?

ui/test/config/app.ts
```diff
@@ -0,0 +1,73 @@
...
+    /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+    reporter: [["html", { outputFolder: "app-test-report" }]],
```
MartinaeyNL: Not sure what we should do for the CI/CD in terms of reporting. You should ask around.
For locally testing it, it would definitely make sense btw

ui/test/fixtures/component.ts
MartinaeyNL, line 1: I'm not sure what we should do about these functions.
Because, from what I understand, these are fixtures used both in UI App tests, and UI component tests.
Don't know if this large fixture class is a solution, nor do I know if you can add them to the individual components.
Because adding them below `/component/**/test/fixtures/` would be ideal, and then reusing them in apps as well.
So, it'd be similar as with `import ... from "@openremote/or-icon"`, but it would be a dev dependency.
Not sure if that's possible at all, but we can take a look I guess.

ui/test/fixtures/shared.ts
```diff
@@ -0,0 +1,133 @@
...
+
+import { Resolve, type JsonSchema, type JsonSchema4, type CombinatorKeyword } from "@jsonforms/core";
+
+declare global {
+  interface Window {
+    _i18next: i18n;
+  }
+}
+
+// TODO: move to `@openremote/util`
```
MartinaeyNL: Yes indeed, this should be moved to the `util` package.
I think however there are already existing functions you can re-use,
so we don't have to write additional code.

ui/test/index.d.ts
MartinaeyNL, line 1: Shouldn't `.d.ts` files be ignored by Git?

ui/test/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "typesVersions": {
+    "*": {
+      "*": [
+        "./*"
+      ]
+    }
+  },
+  "license": "AGPL-3.0-or-later",
+  "devDependencies": {
+    "@jsonforms/core": "^3.5.1",
```
MartinaeyNL: Preferably a `test` package doesn't have any additional dependencies like this.
These JSON forms related dependencies for example, should be handled by `or-json-forms`.
Or was this purely specified for importing types?

ui/test/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "@openremote/util": "workspace:*",
+    "@playwright/experimental-ct-core": "1.52.0",
+    "@playwright/test": "1.52.0",
+    "css-loader": "^6.5.1",
+    "html-webpack-plugin": "^5.5.0",
+    "i18next": "^21.5.3",
+    "playwright": "1.52.0",
```
MartinaeyNL: Don't forget to update Playwright to the latest version upon merging.

ui/test/plugin/index.ts
MartinaeyNL, line 1: I would move this to a separate folder, because I think this is all related to a webpack plugin.
Preferably it's something like `/ui/test/plugins/webpack/`.
And if functions need to be reused across plugins, you could add `/ui/test/plugins/util.ts`,
which is something we tend to do across the repository.

MartinaeyNL, line 1: This looks like a lot of code for providing Webpack support. So just to clarify; is it cleaned up / finished?
If the code is based on an external source like the Playwright documentation, please provide a link and context.

ui/test/plugin/index.ts
```diff
@@ -0,0 +1,149 @@
...
+        inject: "body",
+        scriptLoading: "module",
+      }),
+    ],
+    resolve: {
+      extensions: [".js", ".jsx", ".ts", ".tsx"],
+    },
+  };
+
+  return webpackConfig;
```
MartinaeyNL: Can it reuse the Webpack configuration we already have? Or do we need a specific one for Playwright to work?
As an alternative, we can merge the config objects instead.

ui/test/plugin/register.mjs
MartinaeyNL, line 1: Does this need to be in a separate `register.mjs` file? Or is this a common practice?
Preferably you'd have functions like this exported from an `index.ts` file instead.
But if it's common practice for Playwright it's OK.

ui/test/plugin/registerSource.mjs
MartinaeyNL, line 1: This looks like a lot of confusing boilerplate. Do we need this?
Or is this a copy-paste from that community Webpack plugin we talked about in person?

ui/test/plugin/transform.cjs
MartinaeyNL, line 1: This looks like a lot of confusing boilerplate. Do we need this?
Or is this a copy-paste from that community Webpack plugin we talked about in person?

ui/test/plugin/webpackUtils.ts
MartinaeyNL, line 1: This looks like a lot of confusing boilerplate. Do we need this?
Or is this a copy-paste from that community Webpack plugin we talked about in person?

ui/test/config/playwright/index.html
Ekhorn: Playwright uses [`playwright`](https://playwright.dev/docs/test-components#how-to-get-started) for the component test app. Though we can change it to whatever we want.

ui/component/or-json-forms/playwright-ct.config.ts
Ekhorn: Sadly no.

ui/app/build.gradle
Ekhorn: As discussed I'll look at whether we can reduce the npm gradle tasks down to projects that only have a package.json (or to something that can detect whether the npm script is set).

ui/app/manager/test/fixtures/manager.ts
Ekhorn: `this.page` is passed to the manager at the bottom of the file.

```ts
export const test = base.extend<Fixtures>({
  manager: async ({ page, baseURL }, use) => await use(new Manager(page, baseURL!)),
```

https://github.com/openremote/openremote/blob/346e010bfed8002db7f1516176c280ad1aa5f68d/ui/app/manager/test/fixtures/manager.ts#L35

ui/app/manager/test/fixtures/manager.ts
Ekhorn: We'll need to separate out the DOM utilities from the core package, or like with other utilities copy it over.

ui/component/or-icon/test/or-icon.test.ts-snapshots/Should-render-mdi-icon-chart-areaspline-1-or-icon-linux.png
Ekhorn: We discussed that snapshot testing for icons should be avoided. We would ideally compare the rendered icons to SVGs (if possible).

ui/test/config/app.ts
Ekhorn: Playwright by default reports any issue in the pipeline logs, however I also made sure to export the HTML reports after the pipeline has run for when you want to view the results in more detail (includes recordings).

ui/test/index.d.ts
Ekhorn: I should try to get this combined with the `index.js` file. I don't see why we would ignore `.d.ts` files. We ignore the dist folders of the compiled typescript, which contain `.d.ts` files, but not directly no.

ui/test/package.json
Ekhorn: In case I can figure something out for https://github.com/openremote/openremote/pull/1890#discussion_r2169068006, we can get rid of some dependencies, but we definitely need all the webpack and playwright deps if we ever want to publish the test package.

ui/test/package.json
Ekhorn: Yea, I pinned it because my local setup is unusual. I'll unpin when everything else is done.

ui/test/plugin/index.ts
Ekhorn: This is a playwright plugin, that uses webpack under the hood. I don't expect other plugins will be added. In case we do want to do so we can rename the directory to plugins then.

ui/test/plugin/index.ts
Ekhorn: Will need to have a look.

ui/test/plugin/registerSource.mjs
Ekhorn: We discussed whether or not depend on the third party playwright component testing library for Lit web components. I will try to reduce whatever I can.

ui/test/config/component.ts
```diff
@@ -0,0 +1,29 @@
+import { resolve } from "node:path";
+import { defineCtConfig, Project } from "../";
+
+const { CI } = process.env;
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default (path: string) => {
+  const name = path.split("/").at(-1);
```
Ekhorn: Fix path splitting for Windows (use `path.basename(projectPath)`)

ui/test/config/app.ts
Ekhorn: The default Playwright configuration allows tests to fail twice in case they are a little flaky, which avoids the entire pipeline having to be re-run. Playwright will mention which tests are flaky and what went wrong.

I personally haven't really encountered much in my testing, but very rarely due to rare race conditions the test may timeout, which in that case I prefer to have it at least rerun once to avoid wasting time.

ui/component/or-map/src/mapwidget.ts
Ekhorn: Well `MapWidget` is not a Lit component (just a regular class), nor does a `shadowRoot` have access to the `createElement` function (it belongs to `document`). Besides `createElement` only creates an element in memory which isn't appended to the DOM, so I'm not sure what behavior it could result in outside its purpose here.

.github/workflows/ci_cd.yml
```diff
@@ -704,30 +704,30 @@ jobs:
...
-      - name: Run frontend component tests
+      - name: Run UI component tests
         if: steps.test-ui-components-command.outputs.value != ''
         run: ${{ steps.test-ui-components-command.outputs.value }}
         timeout-minutes: 20
         # continue-on-error: true
 
-      - name: Archive frontend component test results
+      - name: Archive UI component test results
         if: always()
```
richturner, line 720: Shouldn't this conditional be dependent on UI tests running? Don't know if task fails if no dirs can be found but also we don't want an empty artefact

.github/workflows/ci_cd.yml
```diff
@@ -742,19 +742,19 @@ jobs:
...
+          ${{ steps.test-ui-apps-command.outputs.value }}
         env:
           IS_MAIN_REPO: ${{ steps.is_main_repo.outputs.value }}
           MANAGER_TAG: ${{ steps.manager-docker-command.outputs.refTag }}
         timeout-minutes: 20
         # continue-on-error: true
 
-      - name: Archive e2e test results
+      - name: Archive UI app test results
         if: always()
```
richturner, line 755: Same point as component tests

build.gradle
```diff
@@ -22,9 +24,25 @@ scmVersion {
...
     version = scmVersion.version
+
+    afterEvaluate { proj ->
+        def pkgJson = file("${project.projectDir}/package.json")
+        if (!hasScriptsSection(pkgJson)) {
+            tasks.matching { it.name.startsWith('npm') }.configureEach {
+                enabled = false
+            }
+        }
+    }
```
richturner: Some consider such gradle `hacking` as something to avoid, I know we have plenty but doesn't mean we should rush for more.
Projects that want to have a specific npm* task should explicitly add it i.e. it is the responsibility of each sub project to ensure it is configured correctly.

.github/workflows/ci_cd.yml
richturner: So it looks like you want to immediately load the `amd64` image into the engine? Is this because the github test runners are `amd64`, what if that changes, see [deploy.yml](https://github.com/openremote/openremote/blob/1c3323e541365e9caaac68ee7996f5f54c46ad34/.ci_cd/deploy.sh#L157) logic for exporting the docker image. I think I initially avoided loading the images into the docker engine so they didn't end up in the github cache but we could always delete them once finished.

richturner (changes requested):

build.gradle
MartinaeyNL: Hahaha I'm responsible for him making this code. 😂 
I recommended him a more "top-level" approach. instead of listing all `npm` tasks in each gradle project individually.
It would allow executing the `npmTest` task from a top-level directory, making it easier for development work.
It was OK for me, since other tasks like `npmBuild` are now also working again.

But I can imagine it looks a bit hacky haha
Your call 👍

richturner: Hey it works and it was how I approached these gradle tasks in the first place but does feel a bit like fighting the way gradle is meant to work. To opt out we now have to create no-op test scripts and for someone not familiar with the hacks they would take some time to reverse engineer them

.github/workflows/ci_cd.yml
Ekhorn: The `actions/upload-artifact` action logs the following when no results could be found `Warning: No files were found with the provided path: ui/component/*/component-test-report/**. No artifacts will be uploaded.` so I'm leaving it at is (let me know if you want me to add the condition anyways).

.github/workflows/ci_cd.yml
Ekhorn: see https://github.com/openremote/openremote/pull/1890#discussion_r2189794247

ui/test/plugin/index.ts
Ekhorn: The only thing that overlaps are the module rules. I don't see a reason to depend on the webpack config from `generateExports` when almost every configuration needs to be adjusted.

ui/app/manager/test/fixtures/manager.ts
Ekhorn: Sadly, I cannot easily reuse this function, because the intention is that it is used after the user has logged in through the UI.

I could add an environment variable (for testing) to assign the manager instance to the window object to be able to get the access token, but that seems hacky.

Furthermore Playwright code runs in NodeJS by default so the setup and teardown are independent from the frontend, which avoids us having to wrap everything with `this.page.evaluate(async () => ...)` to run it in the browser.

ui/app/manager/test/fixtures/manager.ts
Ekhorn: See https://github.com/openremote/openremote/pull/1890#discussion_r2190389168

ui/component/or-icon/test/or-icon.test.ts-snapshots/Should-render-mdi-icon-chart-areaspline-1-or-icon-linux.png
Ekhorn: Trying `toHaveScreenshot("fixtures/map.svg")` results in the following error. I don't think there is much we can do about this.

`Error: Screenshot name "fixtures-map-or-icon-linux.svg" must have '.png' extension`

I can reduce the tests to just 1 for MDI icons and 1 for our custom icons. What do you think?

ui/test/plugin/webpackUtils.ts
Ekhorn: This is required by the Playwright Webpack plugin. I've added some comments to clarify what these functions are used for.

ui/test/plugin/registerSource.mjs
Ekhorn: Replaced most of it with a 3rd party dependency. The `ui/test/plugin/injectedScriptSource.mjs` file still contains some copied code which we cannot depend on Playwright for, because this bit of code is not exported in the `playwright-core/package.json`.

I could open an issue/PR on https://github.com/microsoft/playwright to have this be exported, though there is of course no guarantee this will merged.

ui/test/plugin/index.ts
Ekhorn: The plugin code is now refactored (see latest changes). I cannot reduce it much further.

I've added context to the functions present. I could add a permalink to the vite plugin this is based on like https://github.com/microsoft/playwright/blob/13dd4fddeb77c6802835e525684e5bea0d2f0d94/packages/playwright-ct-core/src/vitePlugin.ts.

ui/test/fixtures/component.ts
Ekhorn: That does sound like a nice idea, though the only way I can imagine this working is if we either use relative imports to these fixtures, or we adjust the tsconfig files to compile the fixture code.

I currently have this working with relative paths.

ui/app/manager/test/fixtures/data/assets.ts
Ekhorn: I've decided to use thing assets portraying other asset types to be able to remove most properties. Now only `notes` and `location` are required, besides the relevant properties within the test.

ui/test/fixtures/shared.ts
```diff
@@ -0,0 +1,108 @@
...
+      }
+    }, resources);
+  }
+
+  /**
+   * Resolves a request URL to a local filesystem path.
+   * @param url The icoming request URL to resolve
+   */
+  private urlPathToFsPath(url: string) {
+    return path.resolve(__dirname, global.decodeURI(`../../app${new URL(url).pathname}`));
```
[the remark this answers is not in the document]
Ekhorn: It is NOT incorrect ([global does have `decodeURI` in nodejs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects)), however [global is legacy](https://nodejs.org/api/globals.html#global) and yes u can just call the function directly.

.github/workflows/ci_cd.yml
```diff
@@ -447,7 +451,11 @@ jobs:
               echo "pushRequired=true" >> $GITHUB_OUTPUT
               echo "managerDockerImage=openremote/manager:$FIRST_MANAGER_TAG" >> $GITHUB_OUTPUT
             else
-              command="docker build --build-arg GIT_COMMIT=$commitSha --platform linux/amd64,linux/aarch64 -t openremote/manager:$commitShaShort $buildPath"
+              if [ "$(uname -m)" == "aarch64" ]; then
+                PLATFORM=linux/aarch64
+                OTHER_PLATFORM=linux/amd64
+              fi
+              command="docker build --build-arg GIT_COMMIT=$commitSha --platform ${PLATFORM:-linux/amd64} --load -t openremote/manager:$commitShaShort $buildPath && docker build --build-arg GIT_COMMIT=$commitSha --platform ${OTHER_PLATFORM:-linux/aarch64} -t openremote/manager:$commitShaShort $buildPath"
```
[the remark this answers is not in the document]
Ekhorn: They're defaulted and [the platform detection doesn't apply to loading a docker image](https://docs.docker.com/build/building/multi-platform/#prerequisites).

> However, you still won't be able to load the multi-platform images you build into your Docker Engine image store.

.github/workflows/ci_cd.yml
```diff
@@ -203,8 +203,8 @@ jobs:
         if: steps.backend-files-changed.outputs.any_modified != 'true'
         run: echo "value=true" >> $GITHUB_OUTPUT
 
-      - name: Set skip UI tests
-        id: skip-ui-tests
+      - name: Set skip UI component tests
+        id: skip-ui-component-tests
         if: steps.ui-files-changed.outputs.any_modified != 'true'
```
richturner, line 208: `ui-files-changed` may need to be broken into `ui-app-files-changed` and `ui-component-files-changed`

.github/workflows/ci_cd.yml
```diff
@@ -657,36 +657,16 @@ jobs:
...
-        if: always()
+        if: steps.test-backend-command.outputs.value != ''
         uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4
         with:
           name: backend-test-results
           path: test/build/reports/tests
 
       - name: Cleanup backend tests
-        if: ${{ steps.test-backend-command.outputs.value != '' && steps.test-ui-command.outputs.value != '' }}
+        if: ${{ steps.test-backend-command.outputs.value != '' && steps.test-ui-apps-command.outputs.value != '' }}
```
richturner: Looks like an error in the original logic here? This logic should probably be an `||` not `&&`

ui/test/fixtures/component.ts
richturner: Is there a conclusion to this comment?

richturner (commented):
Let's not do icon comparison, we don't want to be testing 3rd party icons. If icons aren't copied/generated then the build step should fail, that should be sufficient?

ui/test/fixtures/component.ts
MartinaeyNL: It has been resolved. Component fixtures can now be written in the package itself 👍 
(see `/component/or-json-forms/test/fixtures/` in the latest commit)

ui/test/plugin/registerSource.mjs
MartinaeyNL: Sure. You can add a comment on top of the `injectedScriptSource.mjs` file with some context.
Along with an explanation of why this file exist, you could refer to a GitHub issue in their repository,
or an external link explaining it as well.

ui/component/or-map/src/mapwidget.ts
MartinaeyNL: Oh wait, yes that makes sense. My bad haha 😅

ui/test/config/app.ts
MartinaeyNL: Oh, is the retry for individual tests? (not a full rerun of all tests)
If so, that is OK. We have the same behavior for our backend tests, so no need to change it.
Just wanted to make sure unnecessary tasks are reran when they haven't failed.

.github/workflows/ci_cd.yml
Ekhorn: I've removed the test UI condition completely, because the backend tests command is only ever run based on the backend test command value, so I see no reason to cleanup the backend tests if they have not even run.

ui/test/config/app.ts
MartinaeyNL: Yes that makes sense. Although I would ask around what we do for other CI/CD tasks.
As in, if no report of backend tests is stored, alongside the logs, it doesn't make sense to add it here too.

I'm also not sure what this report adds on top of logs?

ui/test/package.json
Ekhorn: I bumped the dependencies, but we discussed not to unpin playwright, because we now depend on https://github.com/sand4rt/playwright-ct-web, which is not yet part of Playwright meaning that there is a brief period after each playwright release where they may not be up to date yet breaking the setup.

ui/test/config/app.ts
Ekhorn: Yes it retries (flaky) tests that failed because of a timeout. So in CI we allow a flaky test to be rerun twice, before we consider it a failure. I'll expand the comment in the configuration.

ui/test/config/app.ts
Ekhorn: > As in, if no report of backend tests is stored, alongside the logs, it doesn't make sense to add it here too.

The backend results are archived (see artifacts at https://github.com/openremote/openremote/actions/runs/16196063548). The backend results also use an html report which you can open in your browser.

image

So it would make sense to me include a report for frontend tests as well then.

> I'm also not sure what this report adds on top of logs?

The logs only mention what failed with a stacktrace how it failed and how long it took to run all tests. The report include videos like I mentioned and additional tracing information.

image

ui/test/plugin/registerSource.mjs
Ekhorn: I made the issue and referred to it in the comment.

ui/test/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "css-loader": "^6.5.1",
+    "html-webpack-plugin": "^5.5.0",
+    "i18next": "^21.5.3",
+    "playwright": "1.53.1",
+    "playwright-core": "1.53.1",
+    "ts-loader": "~9.5.2",
+    "typescript": "~5.8.3",
+    "webpack": "^5.90.1",
+    "webpack-cli": "^5.1.4",
+    "webpack-dev-server": "^4.15.1"
```
Ekhorn: Write to rspack

ui/test/package.json
Ekhorn: Try pinning on patch.

Ekhorn: Doesn't work, https://github.com/sand4rt/playwright-ct-web throws an error on diverging patches.

richturner:
@Ekhorn LGTM!

PR #1894, opened 2025-05-23
Add Keycloak IdP APIs

wborn:
## Description

Adds APIs for configuring identity providers in Keycloak.

## Checklist


model/src/main/java/org/openremote/model/security/User.java
```diff
@@ -186,6 +186,16 @@ public User setAttribute(String key, String value) {
...
+    public User setAttribute(String key, List<String> values) {
+        if (attributes == null) {
+            attributes = new ArrayList<>();
+        } else {
+            attributes.removeIf(attr -> attr.getName().equals(key));
+        }
+        values.forEach(value -> attributes.add(new UserAttribute(key, value)));
+        return this;
+    }
+
```
MartinaeyNL, line 198: Can't this function be `values.forEach(value -> setAttribute(key, value)` ?

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -1021,13 +1021,92 @@ public ClientRepresentation createUpdateClient(String realm, ClientRepresentatio
...
         }));
     }
 
+    public Map<String, String> getIdentityProviderImportConfig(String realm, Map<String, Object> importData) {
+        if (importData == null || importData.isEmpty()) {
+            throw new IllegalArgumentException("Import data is null or empty");
+        }
+
+        return getRealms(realmsResource -> realmsResource.realm(realm).identityProviders().importFrom(importData));
+    }
```
MartinaeyNL, line 1051: Can you describe what this function does using a comment?
The `importData` parameter seems very generic, so it'd like to know what the input should look like.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -1021,13 +1021,92 @@ public ClientRepresentation createUpdateClient(String realm, ClientRepresentatio
...
+
+    public void createUpdateIdentityProvider(String realm, String alias, String providerId, Map<String, String> config) {
+        IdentityProviderRepresentation representation = new IdentityProviderRepresentation();
+        representation.setAlias(alias);
+        representation.setProviderId(providerId);
+        representation.setConfig(config);
```
MartinaeyNL, line 1057: I think we should add method overloading would be nice here;
```java
public void createUpdateIdentityProvider(IdentityProviderRepresentation representation) {
```
Abstracting this function makes sense to me as well.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -1021,13 +1021,92 @@ public ClientRepresentation createUpdateClient(String realm, ClientRepresentatio
...
+        getRealms(realmsResource -> {
+            IdentityProvidersResource identityProvidersResource = realmsResource.realm(realm).identityProviders();
+
+            if (identityProvidersResource.findAll().stream().anyMatch(ipr -> alias.equals(ipr.getAlias()))) {
+                identityProvidersResource.get(alias).update(representation);
+            } else {
+                try (Response response = identityProvidersResource.create(representation)) {
+                    if (response.getStatusInfo().getFamily() != Response.Status.Family.SUCCESSFUL) {
+                        LOG.warning("Failed to create identity provider response=" + response.getStatusInfo().getStatusCode() + ": " + representation);
+                        return null;
```
MartinaeyNL: Don't know what would work best here; returning `null` or throwing an exception instead.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -1021,13 +1021,92 @@ public ClientRepresentation createUpdateClient(String realm, ClientRepresentatio
...
+                    }
+                }
+            }
+
+            return identityProvidersResource.get(alias).toRepresentation();
+        });
+    }
+
+    public void createUpdateIdentityProviderMapper(String realm, String idpAlias, String mapperName, String idpMapper, Map<String, String> mapperConfig) {
```
MartinaeyNL, line 1076: Same here as L1031, in regards to applying method overloading. (with `IdentityProviderMapperRepresentation`)

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -1021,13 +1021,92 @@ public ClientRepresentation createUpdateClient(String realm, ClientRepresentatio
...
+                        LOG.warning("Failed to create identity provider mapper '" + mapperName +  "' response=" + response.getStatusInfo().getStatusCode() + ": " + response.getStatusInfo().getReasonPhrase());
+                    }
+                }
+            }
+
+            return null;
+        });
+    }
+
+    public void addAuthenticationExecutionConfig(String realm, String flowAlias, String executionProviderId, String authenticatorAlias, Map<String, String> authenticatorConfig) {
```
MartinaeyNL, line 1101: Same here as L1031, in regards to applying method overloading. (with `AuthenticatorConfigRepresentation`)

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -1021,13 +1021,92 @@ public ClientRepresentation createUpdateClient(String realm, ClientRepresentatio
...
+                    AuthenticatorConfigRepresentation config = new AuthenticatorConfigRepresentation();
+                    config.setAlias(authenticatorAlias);
+                    config.setConfig(authenticatorConfig);
+
+                    try (Response response = authenticationManagementResource.newExecutionConfig(execution.getId(), config)) {
+                        if (response.getStatusInfo().getFamily() != Response.Status.Family.SUCCESSFUL) {
+                            LOG.warning("Failed to create execution config '" + authenticatorAlias + "' response=" + response.getStatusInfo().getStatusCode() + ": " + response.getStatusInfo().getReasonPhrase());
+                        }
+                    }
+                    return null;
```
MartinaeyNL: Shouldn't there be a `catch` statement here?

model/src/main/java/org/openremote/model/security/User.java
wborn: No because then it will no longer be possible to set attributes with the same key but different values.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
wborn: I tried to prevent using any Keycloak classes in the interface as much as possible. This way code will less likely break when the Keycloak APIs change.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
wborn: I tried to prevent using any Keycloak classes in the interface as much as possible. This way code will less likely break when the Keycloak APIs change.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
wborn: I tried to prevent using any Keycloak classes in the interface as much as possible. This way code will less likely break when the Keycloak APIs change.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
wborn: It's a [try-with-resouces statement](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) to make sure the resource is automatically closed.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
wborn: I added some documentation to explain how it it works. 
It's very generic because it uses a very generic Keycloak API. This allows for using the same method for importing configurations of any identity provider.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
wborn: It will throw exceptions now so the method caller can decide how to handle this.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
MartinaeyNL: My bad, I didn't read properly 😂

richturner (approved):

PR #1898, opened 2025-05-23
Feature/array value validator

ebariaux:
I left a question in the comments about the path to use to report the error.

### DIFF
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -35,6 +35,17 @@ dependencies {
     api "com.fasterxml.uuid:java-uuid-generator:$uuidGeneratorVersion"
 
     api "fr.neolegal:exp4j:$exp4jVersion"
+
+    dependencies {
+        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.1'
+        testImplementation 'org.mockito:mockito-core:5.12.0'
+        testRuntimeOnly 'org.glassfish:jakarta.el:4.0.2'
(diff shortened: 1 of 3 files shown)


richturner:
Using spock specification (see VelbusBasicTest) for the gherkin like syntax is quite a nice feature to keep rather than simple junit tests, also functionality provided by groovy can be valuable for more readable tests.

Currently all tests reside in the test module as we have a monolith structure and mostly integration tests I propose we stick with that for now. There is a test file called `AssetModelTest` maybe these tests can be included in there or in a test spec next to this file?

If you prefer java and plain junit then I have no problem with that

ebariaux:
I appreciate the value of Gherkin style tests for integration tests and I'm using them extensively to validate rules and additional end points in custom project.
I looked at the VelbusBasicTest, it's good that it does not start a container but still running that class took 21s on my machine. AssetModelTest runs much faster but requires a dev stack to be running.

I strongly believe we need both integration and unit tests. We need tests that run fast and are "local" to a specific functionality we want to test (not saying 1-1 mapping between a test and a class but more specific than a whole feature).

As to where tests should live, I agree that integration tests should live in their own module, as they're mostly testing the whole solution.
But with unit tests we want to test specific code. We have a "monolith" only as far as code lives in a single repo and is deployed in a single JVM but the project is still organised in different independent gradle modules.
I see no reason tests exercising code specific to that module should be made external to the module. It blurs the lines of dependencies and will make things harder in the future if we want to further split things.

This is just explaining the rationale behind my thinking.
Ideally we should align the whole team on what tests to write, using what technology and where they live.

PR #1905, opened 2025-05-27
Bugfix/MQTT Wildcard subscription does not take into account the service user being a restricted user 

### LINKED ISSUE #1902: MQTT Wildcard subscription does not take into account the service user being a restricted user
### Summary

When subscribing to a MQTT wildcard topic as a service user with the `restricted user` realm role, the client receives events from **assets that are not directly linked** to the service user.

### Affected Topics (Confirmed)

- `asset/#`
- `attribute/+/#`

### Expected Behavior

Restricted service users **should only receive events from assets they are directly linked to**, even when using MQTT wildcard subscriptions.

### Actual Behavior

The restricted service user **receives events from unrelated assets**, violating the expected access control behavior.

### Steps to Reproduce

1. Create a service user with the `restricted user` realm role and link it to a specific set of assets.
2. Subscribe to a wildcard topic like `asset/#` or `attribute/+/#`.
3. Trigger the respective event from a non-linked asset, and see that the event is received by the MQTT client.

### Impact

This may lead to unintended data leakage when using a service user with the `restricted user` realm role.

dominiquekleeven:
Closes #1902 

***

This adds an additional step to the `onSubscribe` function and explictely updates the built AssetFilter with the `userAssetIds` based on the assets linked to the restricted service user.

This ensures the published events are filtered by the assets the service user is linked to (when restricted). See `AssetFilter.java`  https://github.com/openremote/openremote/blob/e9d3cba436cd23f7097d33c83ac0e7b4c98a0cc1/model/src/main/java/org/openremote/model/asset/AssetFilter.java#L212-L214

### DIFF
--- a/manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
+++ b/manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
@@ -295,12 +295,15 @@ public class DefaultMQTTHandler extends MQTTHandler {
             return;
         }
    
-        // If service user is restricted, then restrict subscription to only linked assets
+        // Add an additional filter for restricted service users so that events are additionally filtered to only include events for linked assets
         if (isRestricted) {
+            filter.setRestrictedEvents(true);
+
             List<String> userAssetIds = assetStorageService.findUserAssetLinks(authContext.getAuthenticatedRealmName(), authContext.getUserId(), null)
                 .stream()
                 .map(userAssetLink -> userAssetLink.getId().getAssetId())
                 .toList();
+
             filter.setUserAssetIds(userAssetIds);
         }
 
--- a/manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
+++ b/manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
@@ -295,10 +295,9 @@ public class DefaultMQTTHandler extends MQTTHandler {
             return;
         }
    
-        // Add an additional filter for restricted service users so that events are additionally filtered to only include events for linked assets
+        // Specify userAssetIds for restricted service users so that events are additionally filtered to only include events for these assets
         if (isRestricted) {
-            filter.setRestrictedEvents(true);
-
+            // filter.setRestrictedEvents(true); -- This restricts the events to only attributes with the restricted meta configuration
             List<String> userAssetIds = assetStorageService.findUserAssetLinks(authContext.getAuthenticatedRealmName(), authContext.getUserId(), null)
                 .stream()
                 .map(userAssetLink -> userAssetLink.getId().getAssetId())
--- a/manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
+++ b/manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
@@ -47,6 +47,7 @@ import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import static org.openremote.manager.asset.AssetProcessingService.ATTRIBUTE_EVENT_PROCESSOR;
 import static org.openremote.manager.mqtt.MQTTBrokerService.connectionToString;
@@ -280,6 +281,8 @@ public class DefaultMQTTHandler extends MQTTHandler {
     @SuppressWarnings({"rawtypes", "unchecked"})
     @Override
     public void onSubscribe(RemotingConnection connection, Topic topic) {
+        AuthContext authContext = getAuthContextFromConnection(connection).orElse(null);
+        boolean isRestricted = identityProvider.isRestrictedUser(authContext);
 
         boolean isAssetTopic = isAssetTopic(topic);
         String subscriptionId = topic.getString(); // Use topic as unique subscription ID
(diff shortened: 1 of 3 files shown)


manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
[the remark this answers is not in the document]
richturner: I need to dig into the code but I used to retrieve user assets at the time the subscription was created and then disconnect them if the user asset links change, thus allowing caching as part of the initial subscription, seems like a regression here but will look into the code again.

manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
```diff
@@ -75,6 +76,13 @@ public class DefaultMQTTHandler extends MQTTHandler {
         .expireAfterWrite(300000, TimeUnit.MILLISECONDS)
         .build();
 
+
+    // Intermediary cache for authorized event subscriptions created during the canSubscribe method to be used later in the onSubscribe method
+    // Uses a short TTL to avoid holding stale subscriptions in the cache
+    protected final Cache<String, EventSubscription<?>> eventSubscriptionCache = CacheBuilder.newBuilder()
+        .expireAfterWrite(5000, TimeUnit.MILLISECONDS)
```
richturner: Let's go with a 30s expiry to not be too aggressive as I assume there is some scheduled task to do this work, we can manually evict also (see below)

manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
```diff
@@ -218,6 +226,11 @@ public boolean canSubscribe(RemotingConnection connection, KeycloakSecurityConte
             return false;
         }
 
+        String subscriptionId = topic.getString();
```
richturner: Cache key needs to be linked to the user so use `topic + authContext.getUserId()`

manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
```diff
@@ -280,26 +293,19 @@ public boolean canPublish(RemotingConnection connection, KeycloakSecurityContext
...
 
-        if (filter == null) {
-            LOG.info("Invalid event filter generated for topic '" + topic + "': " + connectionToString(connection));
+        // Get the authorized event subscription from the intermediary cache
+        EventSubscription<?> subscription = eventSubscriptionCache.getIfPresent(subscriptionId);
+        if (subscription == null) {
+            LOG.info("Subscription not found in intermediary cache: " + subscriptionId);
             return;
         }
+        String sessionKey = getSessionKey(connection);
```
richturner: Is this unused?

manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
```diff
@@ -280,26 +293,19 @@ public boolean canPublish(RemotingConnection connection, KeycloakSecurityContext
...
+        
         String subscriptionId = topic.getString(); // Use topic as unique subscription ID
-        AssetFilter filter = buildAssetFilter(topic);
-        Class subscriptionClass = isAssetTopic ? AssetEvent.class : AttributeEvent.class;
-        String sessionKey = getSessionKey(connection);
 
-        if (filter == null) {
-            LOG.info("Invalid event filter generated for topic '" + topic + "': " + connectionToString(connection));
+        // Get the authorized event subscription from the intermediary cache
+        EventSubscription<?> subscription = eventSubscriptionCache.getIfPresent(subscriptionId);
```
richturner, line 300: Annoyingly you cannot get and evict in a single call but I think we should evict here

test/src/test/groovy/org/openremote/test/mqtt/MqttBrokerTest.groovy
```diff
@@ -659,18 +662,168 @@ class MqttBrokerTest extends Specification implements ManagerContainerTrait {
...
+        }
+
+
+        and: "the event should only contain attributes that are read restricted"
+        conditions.eventually {
+            assert restrictedReceivedEvents.size() == 1
+            def event = restrictedReceivedEvents.get(0)
+            assert event instanceof AssetEvent
+            def assetEvent = event as AssetEvent
+            assert assetEvent.getAsset().getAttribute("notes").isEmpty()
```
richturner: use `hasAttribute()`

richturner (approved):

manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
dominiquekleeven: Changed it to be 30 seconds

manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
dominiquekleeven: Changed it for both canSubscribe and onSubscribe. Now uses the topic and userId for the key

manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
dominiquekleeven: Used for here

```
Map<String, Consumer<? extends Event>> subscriptionConsumers = sessionSubscriptionConsumers.computeIfAbsent(sessionKey, (s) -> new HashMap<>());
```

Its consistent with onUnsubscribe

manager/src/main/java/org/openremote/manager/mqtt/DefaultMQTTHandler.java
dominiquekleeven: Added the invalidate call after the null check

test/src/test/groovy/org/openremote/test/mqtt/MqttBrokerTest.groovy
dominiquekleeven: Changed to hasAttribute

PR #1912, opened 2025-06-01
[Modbus] Add support for custom register read lengths, to accomodate larger datatypes

pankalog:
Now actually supports `REAL`, `DINT`, `DWORD`, etc.

## Description
A Modbus register has a length of 16 bits or 2 bytes, which is not enough to represent certain data types, like floats, double-integers, etc. Hence, for those data types, we need to tell PLC4X to read that amount of data registers, that it can then use to parse the data type correctly.

It also allows users to set their own, custom amount of registers to be read, in case it's more/less, like `WCHAR` etc.

There seems to be an issue with the UI processing primitive AgentLink parameters as required, but that is not the case for something like that length, the write address, the read address, etc. I would like to eventually get this solved.

## Checklist

### DIFF
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -35,6 +35,17 @@ dependencies {
     api "com.fasterxml.uuid:java-uuid-generator:$uuidGeneratorVersion"
 
     api "fr.neolegal:exp4j:$exp4jVersion"
+
+    dependencies {
+        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.1'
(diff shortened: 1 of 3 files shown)


pankalog:
@richturner Id like to make some more modifications to the `ModbusAgentLink` to remove the need for required agent link values, so please review but maybe don't merge until that part is pushed as well.

richturner (approved):

pankalog:
@richturner can be merged now, hopefully before release 🤩

richturner (approved):

PR #1916, opened 2025-06-03
Improve container trait ruleset re-synchronisation

richturner:
`ContainerTrait.startContainer()` tries to speed up tests by reusing the container and restoring the initial state between tests, part of that logic captures the rulesets after first startup and then on next startContainer call purges all rulesets and re-inserts the captured ones, for this to work post recent hibernate/JPA updates we need to clear the version and ID fields.

### DIFF
--- a/test/src/main/groovy/org/openremote/test/ContainerTrait.groovy
+++ b/test/src/main/groovy/org/openremote/test/ContainerTrait.groovy
@@ -281,9 +281,21 @@ trait ContainerTrait {
             }
 
             // Track rules (very basic)
-            TestFixture.globalRulesets = getRulesets(GlobalRuleset.class)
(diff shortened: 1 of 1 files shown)


ebariaux:
This does not work for me, I got a NPE later in the startContainer code
Line 357
```
            while (i < 100 && TestFixture.realmRulesets.stream().filter { it.enabled }.any { !rulesService.realmEngines.containsKey(it.realm) || !rulesService.realmEngines.get(it.realm).deployments.containsKey(it.id) }) {
```
the id of the ruleset is used in the containsKey check

Moving it to the bottom of the method makes it pass the first time but fail on the next occurrences.

richturner:
@ebariaux now working correctly tested in custom project and working as expected

PR #1917, opened 2025-06-03
MQTT agent message match filter bug fix

### LINKED ISSUE #1886: Inbound MQTT Agent messages are not filtered by means of messageMatch settings
**Describe the bug**
Inbound MQTT Agent messages are not filtered by means of agentLink messageMatchFilter and messageMatchPredicate parameter settings.

**To Reproduce**
- Create MQTT Agent
- Create an asset attribute with an agentLink
- Configure the agentLink with subscriptionTopic, valueFilter, messageMatchFilter and messageMatchPredicate
- Publish messages to the MQTT topic
- The asset attribute is always updated no matter if the messageMatchFilter/messageMatchPredicate filter criteria is met or not

**Expected behavior**
Asset attribute value should be updated only if messageMatchFilter/messageMatchPredicate filter criteria is met.

**Cause**
The messageMatchFilter and messageMatchPredicate parameter settings are not processed if the asset attribute is linked to a MQTT Agent.

rainerhitz:
Fixes #1886


test/src/test/groovy/org/openremote/test/protocol/mqtt/MQTTClientProtocolTest.groovy
```diff
@@ -70,13 +79,47 @@ import java.security.PrivateKey
...
+import static org.openremote.model.value.ValueType.BOOLEAN
 import static org.openremote.model.value.ValueType.NUMBER
 
 class MQTTClientProtocolTest extends Specification implements ManagerContainerTrait {
 
+    @Shared
+    int mqttBrokerPort
+
+    @Shared
+    Server mqttBroker
```
richturner: If you're running the whole manager then you can just connect to our own broker

rainerhitz: I've made the change and replaced the Moquette test broker with the internal OR broker.

PR #1928, opened 2025-06-10
Enhancement/EBS Volume

denniskuijs:
## Checklist

### DIFF
--- a/.ci_cd/aws/cloudformation-create-ec2.yml
+++ b/.ci_cd/aws/cloudformation-create-ec2.yml
@@ -881,6 +881,7 @@ Resources:
 
   EBSDataVolume:
     Type: AWS::EC2::Volume
+    DeletionPolicy: Snapshot
     Properties:
       AvailabilityZone: !GetAtt EC2Instance.AvailabilityZone
       Size: !Ref DataDiskSize
@@ -891,7 +892,6 @@ Resources:
           Value: !Sub ${Host}-data
         - Key: Type
           Value: !Sub ${Host}-or-data-not-in-use
-    DeletionPolicy: Snapshot
 
   # DLM Policy for automatic snapshot creation
 
@@ -900,7 +900,12 @@ Resources:
     DependsOn: EBSDataVolume
     Properties:
       ExecutionRoleArn: !Ref DLMExecutionRoleArn
-      Description: 'or-data'
+      Description: 
+        !Join
+          - '-'
+          - !Split
+            - '.'
+            - !Sub ${Host}-or-data
       State: ENABLED
       PolicyDetails:
         PolicyLanguage: STANDARD
@@ -912,13 +917,14 @@ Resources:
           - VOLUME
         Schedules:
           - Name: Daily Backup
+            CopyTags: true
             CreateRule:
               Interval: 24
               IntervalUnit: HOURS
               Times:
-              - '05:00'
+                - '05:00'
             RetainRule:
-              Count: 5
+              Count: 1
       Tags:
         - Key: Name
           Value: !Ref Host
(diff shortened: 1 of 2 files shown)


richturner:
Where are we on this PR, no issues linked and description is limited

richturner:
Not sure one snapshot of each host is sufficient we should decide the correct retention policy

richturner (dismissed):
Unless costs are very prohibitive let's go with 1 week of daily backups and 4 weeks of weekly backup

denniskuijs:
I've moved this PR to draft for now. We first want to make a cost breakdown of the total snapshot costs for all our instances, especially the largest ones.

denniskuijs:
@richturner I've created a cost breakdown for storing EBS snapshots for all our instances based on a daily (7 copies) and weekly (4 copies) retention policy. 

You can find the overview [here](https://docs.google.com/spreadsheets/d/1FFxPP8ua45PykmNhoy24LWU35S8uxRxg9f9IShWiL6Y/edit?usp=drive_link)

Points to take in mind:
- The costs are calculated without compression. Since AWS uses compression to reduce storage usage, the actual costs may be lower.
- The calculations are based on the total used capacity on each volume. When migrating instances to the new `EBS` volume implementation, we will only create snapshots of the data volume that stores the `/var/lib/docker/volumes` directory. This means the actual stored data might (and costs) may be lower.
- The first snapshot is always a `full snapshot` which includes all the data on the volume. Every subsequent snapshot is `incremental`, meaning AWS only stores the changes compared to the previous snapshot.
  
  The costs assume an average daily data change rate of 0.5%. Some instances are lightly used and will change less, while others may have a higher rate. You can adjust the percentage in the spreadsheet to see the effect on total costs.

We should monitor costs regularly and adjust the retention policy if needed once all the instances are migrated. For now, I think the proposed retention policy is fine, as the estimated costs are relatively low.

ebariaux:
Indeed costs seem minimal compare to the rest of our infra.
The 0.5% value on incremental backups has an important impact though and I can't judge if that's realistic or not.
Monitoring after this is in place will anyway be the way to go.

.ci_cd/aws/cloudformation-create-ec2.yml
```diff
@@ -912,18 +915,25 @@ Resources:
           - VOLUME
         Schedules:
           - Name: Daily Backup
+            CopyTags: true
             CreateRule:
               Interval: 24
               IntervalUnit: HOURS
               Times:
-              - '05:00'
+                - '03:00'
```
ebariaux: Using Interval and co here and CronExpression for weekly backup.
Maybe align on always using CronExpression for consistency ?

.ci_cd/aws/cloudformation-healthcheck-alarm.yml
```diff
@@ -1,8 +1,8 @@
 # ---------------------------------------------------------------------------------------------------------------------
-# Creates an alarm for Route53 health check - must be run in the us-east-1 region
+# Creates an alarm for Route53 HealthCheck - must be run in the us-east-1 region
```
ebariaux: Health Check ?
We're not very consistent in our files on how we write it:
HealthCheck
Health Check
Health check

I would think the later to be the most aligned with AWS doc but seems we're going for Health Check (just as done 3 lines below)

ebariaux (changes requested):
Minor cosmetic changes, code looks fine to me

.ci_cd/aws/cloudformation-create-ec2.yml
denniskuijs: Good point! Using the `CronExpression` for both schedules makes sense as it’s more flexible than the `Interval` option.

.ci_cd/aws/cloudformation-healthcheck-alarm.yml
denniskuijs: I try to align the comments with the AWS docs as much as possible. But Health Check is fine here.

ebariaux (approved):

PR #1941, opened 2025-06-16
Enable direct rule navigation via rule ID in URL

### LINKED ISSUE #1520: Rules: use rule ID in the URL so you can navigate to a rule directly

hamadhassan3:
## Description
This PR adds support for deep linking to individual rules.

- Users can now navigate directly to a rule using its ID in the URL (e.g., /rules/123). 

- The URL now updates automatically when a user selects a rule in the UI.

### Requesting Feedback

1. The `_fetchRuleset(rulesetId: number)` method seems a bit hacky to me, as it tries to fetch the ruleset first from the realm API and then from global. However, this was unavoidable because we do not have any further information in the URL to distinguish between the two. Please let me know if this is acceptable.

2. I tried to follow the same flow as the Assets and Insights pages, but the implementation on the rules page is somewhat different. After spending some time trying to pass values from parent to children, I realized that the best way to handle this was to use the existing `_onNodeSelectionChanged` method in `or-rules`. This avoids unnecessary dependencies between the components and lets us decouple the Realm/Global logic from this flow.

### Execution
https://github.com/user-attachments/assets/408b8624-f5ce-4c94-b32d-8d4281f29425

Resolves #1520

## Checklist

### DIFF
--- a/ui/app/manager/src/pages/page-rules.ts
+++ b/ui/app/manager/src/pages/page-rules.ts
@@ -142,6 +142,9 @@ export class PageRules extends Page<AppStateKeyed>  {
                 this._ruleId = newRuleId;
                 this._updateRoute(newRuleId.toString());
             }
+        } else {
+            this._ruleId = undefined;
+            this._updateRoute();
         }
     }
 
--- a/ui/app/manager/src/pages/page-rules.ts
+++ b/ui/app/manager/src/pages/page-rules.ts
@@ -1,10 +1,10 @@
 import {css, html} from "lit";
 import {customElement, property, query} from "lit/decorators.js";
 import "@openremote/or-rules";
-import {OrRules, RulesConfig} from "@openremote/or-rules";
+import {OrRules, OrRulesSelectionEvent, RulesConfig, RulesetNode} from "@openremote/or-rules";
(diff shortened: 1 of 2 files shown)


MartinaeyNL (commented):
Hi! Thanks for the PR!

I'd overall take a different approach of applying the `/rules/<id>` URL.
Let me explain.

For the Rules page, the `or-rule-tree` component is responsible for fetching rulesets with the HTTP API.
It will fetch **all** rules from the realm, and puts them in the tree afterwards.
Once a user selects them, the `or-tree-select` event is dispatched from the `or-rule-tree` component.
`or-rules` listens for that event, get the ruleset object from the event payload, and updates the ruleset inside `or-rule-viewer`.

I think the best approach here, is to pass the `<id>` string parameter further down the element tree.
As in. pass the URL parameter -> `page-rules` -> `or-rules` -> `or-rule-tree`.
Then, inside `or-rule-tree` you can add a condition after the rules are fetched.
"Does this new list include `<id>`? Then select that ruleset."

Let me know what you think!

hamadhassan3:
Hi @MartinaeyNL, thanks for the detailed feedback. That makes more sense. My first instinct was also to follow the approach that you have suggested. I'll refactor the code accordingly.

ebariaux:
No follow-up

PR #1942, opened 2025-06-17
Dashboards with gateway widgets are visible for users without asset access

### LINKED ISSUE #1932: Insights: page that contains Gateway widgets to which a restricted user does not have access is shown
**Describe the bug**
Pages that contain widgets with asset attribute data the logged in restricted user may not see (is not linked to) should not be shown in the list. This check whether a restricted user is linked to the asset on the page does not take into account the gateway used in a Gateway widget. If a restricted user is not linked to the gateway used in the widget, that page should not show up in the list for that user.

**To Reproduce**
1. Create a restricted user with the read role
2. Create a gateway
3. Create an Insights page with a Gateway widget that uses the gateway you created
4. Log in as the restricted user and navigate to the Insights page
5. See the Insights page you created in the list

**Expected behavior**
5. The Insights page with the Gateway widget does not appear in the list since you are not linked to it.

**Screenshots**

**OR version**
Tested on 1.6.0

MartinaeyNL:
## Description
Fixes #1932 

## Changelog
- Now only showing dashboards where user has gateway asset access,
by adding the `attributeRefs` field to the gateway widget configuration.
- Added SQL migration script to add the `attributeRefs` array to each existing gateway widget.
- Added `AssetWidgetConfig` TypeScript interface, to encourage use of the `attributeRefs` field in other widgets.
- Fix for uncaught errors within the gateway widget.

## Checklist

### DIFF
--- a/manager/src/main/resources/org/openremote/manager/setup/database/V20250617_01__GatewayWidgetRefs.sql
+++ b/manager/src/main/resources/org/openremote/manager/setup/database/V20250617_01__GatewayWidgetRefs.sql
@@ -9,7 +9,7 @@ SET template = jsonb_set(
                                         widget,
                                         '{widgetConfig,attributeRefs}',
                                         CASE
-                                            WHEN widget -> 'widgetConfig' ->> 'gatewayId' <> '' THEN
+                                            WHEN widget -> 'widgetConfig' ->> 'gatewayId' != '' THEN
                                                 jsonb_build_array(
                                                         jsonb_build_object(
                                                                 'id', widget -> 'widgetConfig' ->> 'gatewayId',


MartinaeyNL:
Updated the changelog, and the code is ready to review 👍

MartinaeyNL:
Requested a review from @Ekhorn.

@DonWillems It might be tricky to test locally, so I've deployed it to a test server.
Feel free to check out if you'd like 👍

manager/src/main/resources/org/openremote/manager/setup/database/V20250617_01__GatewayWidgetRefs.sql
Ekhorn, line 1: When I create a gateway dashboard without connecting a gateway asset it will still show to the restricted user, which is fine?

When I connect the gateway dashboard to the asset it doesn't show to the restricted user anymore after applying the migration (just like the issue describes).

When I connect the gateway asset to a dashboard after the migration is applied, it does indeed not show. It stores the following value in the database. That is without `name: null`. Is there a reason to set `name -> NULL`?

```
"attributeRefs": [
  {
    "id": "5Wvnc6lLFZxi4T7rzgsPCX"
  }
]
```

FYI, you can write `<>` as `!=` (bit more readable).

---

When creating a gateway dashboard and connecting it to a gateway asset on master, and starting the manager with `OR_SETUP_RUN_ON_RESTART: false` on this branch, then it does apply the migration properly.

```
"attributeRefs": [
 {
            "id": "5nGXtdrAWcf4FKIvvWYQpX",
            "name": null
 }
]
```

MartinaeyNL: > When I create a gateway dashboard without connecting a gateway asset it will still show to the restricted user, which is fine?

Yes. I assumed that Gateway widgets without a linked asset, are still included in an HTTP response body to the user.
It would be an empty widget at that point, but the frontend UI could decide that themselves.
.
> It stores the following value in the database. That is without `name: null`. Is there a reason to set `name -> NULL`?

I don't think it'll matter. Whether it's not set, or if it is NULL, it should behave the same.
The frontend handles both scenarios perfectly, and the backend doesn't read the `name` field at all.
.

 > FYI, you can write <> as != (bit more readable).
 
 Thanks! I'll update that. ChatGPT recommended me this, so I had no clue 😂

MartinaeyNL:
Addressed the review comments. Let me know @Ekhorn what you'd like to do with this.

MartinaeyNL:
🥳

PR #1945, opened 2025-06-20
Flow node titles have weird formatting

DonWillems:
## Description
Node titles are formatted as e.g. "b o o l e a n i n p u t" because the humanLike function that is used can't handle the all caps for the original nicely "BOOLEAN_INPUT".
So I made the original lowercase first, and then applied the humanLike, now it comes out nicely.
However, there might very well be a more sophisticated way to deal with this...

Additional thought:
This title is especially useful for the processors where only a sign is used to indicate its function. For some input for example its a bit inconsistent as the title will not be exactly the same as the label of the node. I would say this is fine for now.


MartinaeyNL:
To understand this PR, I've pulled the branch locally to try it out.
I agree with your additional thoughts. No need to do additional improvements, this is OK👍 

Let's get this merged 💪

PR #1946, opened 2025-06-20
Logs: deselect all categories by default, making filtering 1 category easier

DonWillems:
## Description
The category multiselect allows the user to filter the messages shown in the logs.
When I use the logs I often only want to see 1 log category, then its annoying to have to deselect all other categories to end up with the one you want. This change has all categories deselected by default (still showing all logs), so that you only have to click the one category you want to see.

I realize another use case would be that you want to see all logs, except for one of two that are filling the page that you don't need. @richturner , is that why it was implemented like this you think?

### DIFF
--- a/ui/component/or-log-viewer/src/index.ts
+++ b/ui/component/or-log-viewer/src/index.ts
@@ -289,7 +289,7 @@ export class OrLogViewer extends translate(i18next)(LitElement) {
                         ${hideCategories ? `` : getContentWithMenuTemplate(
                             html`<or-mwc-input .type=${InputType.BUTTON} raised ?disabled="${disabled}" .label="${i18next.t("categories")}" icontrailing="chevron-down"></or-mwc-input>`,
                             this._getCategoryMenuItems(),
-                            this.categories,
+                            undefined,
                             (v) => this._onCategoriesChanged(v as Model.SyslogCategory[]),
                             () => this._onCategoriesClosed(),
                             true


MartinaeyNL:
I'll pick up the review for this, and try it out locally. 👍 
I've assigned @richturner as well, to address the question Don asked in the PR description.

ui/component/or-log-viewer/src/index.ts
```diff
@@ -289,7 +289,7 @@ export class OrLogViewer extends translate(i18next)(LitElement) {
                         ${hideCategories ? `` : getContentWithMenuTemplate(
                             html`<or-mwc-input .type=${InputType.BUTTON} raised ?disabled="${disabled}" .label="${i18next.t("categories")}" icontrailing="chevron-down"></or-mwc-input>`,
                             this._getCategoryMenuItems(),
-                            this.categories,
+                            undefined,
```
MartinaeyNL: So in short, you'd like to;
- If no category is selected, it will show all logs from all categories.
- No category will be checked/selected by default.

If that's true, then we should change the initial categories that are selected.
In the `shouldUpdate` function at line 238, the default value can be removed;
```
if (this.config && this.config.initialCategories) {
    this.categories = [...this.config.initialCategories];
} else {
    // This condition can be removed
    this.categories = Object.keys((Model as any)["SyslogCategory"]) as Model.SyslogCategory[];
}
```

MartinaeyNL (changes requested):
Placed a small comment, which should be an easy fix.
Should I commit that change myself @DonWillems ?

richturner:
Not sure I put much thought into the initial implementation, I think logging in general could do with revisiting but this makes sense, maybe a deselect all option could suffice

MartinaeyNL (approved):

PR #1956, opened 2025-06-24
Remove OR_SSL_PORT from Readme

denniskuijs:
## Description
We're receiving an increasing number of questions from users encountering issues when starting OpenRemote using the Quick Start Guide.

Currently, the documentation includes a startup command that overrides the `OR_SSL_PORT` environment variable. However, when this variable is configured, OpenRemote may not be accessible via the hostname defined in the `OR_HOSTNAME` variable.
If the `OR_SSL_PORT` is not specified, OpenRemote uses the standard SSL port `443`, which is fine for most use cases.

To avoid confusion, it's better to remove the `OR_SSL_PORT` variable from the documentation.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ The quickest way to get your own environment with full access is to make use of 
 [OpenRemote Stack](https://raw.githubusercontent.com/openremote/openremote/master/docker-compose.yml) (Right click 'Save link as...')
 3. In a terminal `cd` to where you just saved the compose file and then run:
 ```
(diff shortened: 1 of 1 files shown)


richturner:
Yeah this probably should have been updated when we introduced the `-1` default; for history there are problems with keycloak issuer validation when 443 is explicitly specified because `https://mydomain.com` is (in the version of keycloak at the time of testing) not treated the same as `https://mydomain.com:443`

PR #1957, opened 2025-06-24
Prevent parallel rule execution

ebariaux:


test/src/test/groovy/org/openremote/test/rules/RulesEngineFiringTest.groovy
```diff
@@ -0,0 +1,98 @@
...
+        AtomicBoolean parallelExecutionOccurred = new AtomicBoolean(false)
+        mockRulesEngine.fireAllDeployments() >> {
+            if (rulesExecuting.getAndSet(true)) {
+                parallelExecutionOccurred.set(true)
+            }
+            callRealMethod()
+            rulesExecuting.set(false)
+        }
+        rulesService.realmEngines.put(Constants.MASTER_REALM, mockRulesEngine)
+        rulesService.start(container)
```
richturner, line 83: Nice way to capture this in a test

PR #1964, opened 2025-06-26
UI bugfix: Remove the ability to edit gateway service user in UI

Hackerberg43:
## Description
- Removes the possibility to edit gateway-service user rights via the users-page. The reason is to increase clearity.
- You can still activate/deactivate the user.

The gateway service user is treated differently in the backend and has its own set of permissions on top of the ones you can set. The gateway service user is not ment to be used as an API user, so additional rights should not be required for standard use.
The option of deleting the user is also hidden, since this is done automatically when deleting the Gateway Asset.

Ofcourse this is just a front-end thing, you can still alter the user rights/delete it directly via the api.


MartinaeyNL:
Thanks for the PR!
I'll try to take a look at this tomorrow, as it's relatively easy to review.
Would be great to get this bug resolved 😉

ui/app/manager/src/pages/page-users.ts
```diff
@@ -912,6 +918,7 @@ export class PageUsers extends Page<AppStateKeyed> {
                                           .value="${user.secret}"
                                           .type="${InputType.TEXT}"></or-mwc-input>
                             <or-mwc-input ?readonly="${!user.id || readonly}"
+                                          ?disabled="${isGatewayServiceUser}"
```
MartinaeyNL, line 922: I was wondering whether the "Regenerate secret" button should be disabled.
Regenerating when the secret has been compromised sounds like a good option to have.
What's your opinion?

MartinaeyNL (commented):
Pushed a small commit to ensure that `gateway-***` users can not be created.
Code looks great, only one small comment on the regenerate secret button 👍

ui/app/manager/src/pages/page-users.ts
Hackerberg43: I agree, but that means you break the gateway asset/manager interconnect aswell. So maybe recreation of the asset is required? I was not able to test this, because the regenerate secret button does currently not work, at least not on our staging deployment and a local test.

MartinaeyNL: I took a quick look, and you are correct haha
The regenerate secret button was broken, which I just committed a fix for. 👍 

In regards to disabling the "Regenerate secret" button;
Generating a new secret for a service user wouldn't reflect back to the attribute on the Gateway asset.
So it makes sense to disable it for now.

MartinaeyNL:
Pushed a hotfix for the "Regenerate secret" button, which was not working for service users.
I'll check with @DonWillems on Monday what he thinks, and continue from there.

MartinaeyNL (approved):

MartinaeyNL:
Don is OK with the changes, so I'm happy to merge this 👍

Issue #1767, opened 2025-03-21
Dynamic JSON Forms: Investigate how to make the solution more user friendly


Ekhorn:
## User Story
As an integrator I want the UI for configuration items to be simplified, so that I can more easily configure my assets.

## Description

**What?**

1. The MVP currently requires every single parameter to be selected one by one.

![Image](https://github.com/user-attachments/assets/3cea077e-533f-4229-805d-9cc7b0b1587a)

2. The constraints configuration item has constraints that should be unique; once added should not show again in the select dialog.
3. Show only compatible constraints in the UI.

![Image](https://github.com/user-attachments/assets/c9a95a40-be0c-42a0-8376-ad81954e9d8d)

4. Don't show duplicate titles

**Why?**

1. It would make sense for some inputs to show by default (even if they're not required?) 
2. A constraint can only be applied once anyways.
3. Not all constraints can be applied to every attribute type.
4. Yes.

**How?**

1. May need a custom renderer implementation based on the JSON Schema using the `@InjectJsonSchema` annotation.
2. Investigate how to restrain this through the JSON Schema definition.
3. Investigate if certain constraints can be restricted down by attribute type.
4. Investigate ;)

## Acceptance Criteria

- [ ] We'll need to check in with users what they think of the MVP.

Ekhorn:
Things to resolve after meeting with end user:
- [x] Missing descriptions for constraint parameters
> ![Image](https://github.com/user-attachments/assets/2c7f0662-332b-42e4-b407-f281f5339968)
- [x] Add examples of allowed values to parameter descriptions in dialog
> The time based constraints were very unclear as expected the attribute update to be used
> ![Image](https://github.com/user-attachments/assets/f5bd3e8c-8d58-4e62-b33a-65a9976039e0)
> Examples like this would be helpful:
> | Type                                  | Value                         |
> | ------------------------------------- | ----------------------------- |
> | Date Only                             | 2025-04-18                    |
> | Date and Time (UTC)                   | 2025-04-18T14:30:00Z          |
> | Date and Time with Time Zone          | 2025-04-18T14:30:00+02:00     |
> | Date and Time with Fractional Seconds | 2025-04-18T14:30:00.123Z      |
> | Combined Date and Time with Time Zone | 2025-04-18T14:30:00.456-05:00 |
> | Milliseconds Timestamp                | 1713442200456                 |
- [ ] Add information buttons or helper text that describe the parameters of configuration items once they're already added.
[...]

Issue #1766, opened 2025-03-21
Dynamic JSON Forms: Add the ability to translate the JSON Schema descriptions


Ekhorn:
## User Story

As an integrator I want to be able to see descriptions of configuration item parameters in the language of choice, so I can understand what they do and how their used.

## Description

The Dynamic JSON Forms MVP can found on the [dynamic-json-forms/value-constraints](https://github.com/openremote/openremote/tree/dynamic-json-forms/value-constraints) branch, which implements an endpoint that generates the schemas for the requested configuration item, which currently support the Agent Link and Constraints configuration items.

![Image](https://github.com/user-attachments/assets/7e3a6452-957e-42ee-99c5-5f04722c47e1)

The solution should be able to be used with any language by allowing for translations of those descriptions to be added.

## Acceptance Criteria

- [ ] The solution must continue to work flawlessly when adjusting model definitions.
- [ ] The descriptions are available in English and Dutch. 

**Configuration item example**
- [ ] 1. Given I'm on the modify assets page.
- [ ] 2. When I click `+ Add configuration item`, select and create any configuration item.
- [ ] 3. When I click `+ Add parameter` and select any parameter.
- [ ] 4. Then it should show the descriptions of parameter in English.
- [ ] 5. Given I select Dutch.
- [ ] 6. Then it should show the descriptions in Dutch.

## Implementation Notes

See https://jsonforms.io/docs/i18n

Ekhorn:
1. Enabling translations in jsonforms/core is a matter of setting the `i18n` property in the context in `ui/component/or-json-forms/src/index.ts` with the following values

```ts
                i18n: {
                    // TODO: is this the right way to consume the currently configured locale??
                    locale: manager.language,
                    translate: (id, defaultMessage, values) => {
                        return i18next.t("schema.item." + id);
                    },
                    // TODO: Look into what to do with errors
                    // translateError: (error, translate, uischema) => null,
                }
```

2. Though to make the setup a little nicer. We could make a script that extracts all possible labels, descriptions, titles etc. into the translations files so we don't need to add those keys manually.

My idea would be to run this during the same phase as the transpile Java model task, so `npm run serve` would automatically create all keys you'll need.

Ekhorn:
> **Implementation details**
> 
>     1. Enabling translations in jsonforms/core is a matter of setting the `i18n` property in the context in `ui/component/or-json-forms/src/index.ts` with the following values
> 
> 
>                 i18n: {
[...]

Issue #1747, opened 2025-03-18
Value Converter & Write Value Converter descriptions are missing in the UI


Ekhorn:
**Describe the bug**

The 

**To Reproduce**
1. Go to 'Asset page'
2. Create an asset
3. Add an agent link
4. Click on agent link parameters
5. Click on either `Value Convertor` or `Write Value Convertor`

**Expected behavior**

The description box to display the descriptions on:

https://github.com/openremote/openremote/blob/99212f9d446000e091b98ede2b81c9872d5eb92d/model/src/main/java/org/openremote/model/asset/agent/AgentLink.java#L53

https://github.com/openremote/openremote/blob/99212f9d446000e091b98ede2b81c9872d5eb92d/model/src/main/java/org/openremote/model/asset/agent/AgentLink.java#L61

**Expected behavior**

Shows no description.

**Screenshots**

![Image](https://github.com/user-attachments/assets/4a609abd-57a0-4d8d-9763-56ce4bb9c62a)

![Image](https://github.com/user-attachments/assets/06bfe5e5-6d22-4b28-850f-c18a0a8f6af0)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

This may have happened at same time #1731 occurred?

Ekhorn:
The following line

https://github.com/openremote/openremote/blob/99212f9d446000e091b98ede2b81c9872d5eb92d/model/src/main/java/org/openremote/model/asset/agent/AgentLink.java#L57

causes the JSON Schema not to contain the description that would otherwise be shown in the UI.

![Image](https://github.com/user-attachments/assets/31629062-e7b3-4096-90a1-544f1fae15fa)

The documentation in https://github.com/mbknor/mbknor-jackson-jsonSchema/blob/883805554abed514aefe7dd5aa863a289c259438/README.md#flexible

mentions the `merge = false` will replace the entire description with the provided description.

> The default behaviour of @JsonSchemaInject is to merge the injected json into the generated JsonSchema. If you want to have full control over it, you can specify @JsonSchemaInject.merge = false to replace the generated jsonSchema with the injected json.

which is why the description doesn't show and by default it merges injected JSON with the description.

Ekhorn:
Before 



After

Issue #1704, opened 2025-02-24
Dynamic JSON Forms: Investigate what is needed to make the MVP


Ekhorn:
## Description

**What?**

In order to create a unified solution for dynamically generating JSON forms, validating their output and generating documentation, the following must be done:

**How?**

1. Decomposition -> What is involved in the current JSON Forms (Agent Linking) implementation? How is this divided within the code base?
2. Gap Analysis -> What is still missing in the current implementation? What is needed to make the solution work?
3. Available Product Analysis -> What technologies and standards can be used to fill these gaps?
4. Architecture Design -> The design the intended solution using a component diagram.

**Why?**

So that actionable steps can be taken to build a first prototype (MVP) that replaces the current Constraints configuration item textbox.

## Acceptance Criteria

- [x] Required technologies have been documented
- [x] The architecture design describes
  - [x] What is used as single source of truth
  - [x] How this source can be used to generate forms
  - [x] How this source can be used to validate the form output
  - [x] How this source can be used to generate documentation
- [x] Documenting what actionable steps to take next and why
  - [x] Create or update the MVP issue
  - [x] The actionable steps describe how to integrate the solution with the Constraints configuration item

## Implementation Notes

The solution may be based off how linking Agents to Asset Attributes is currently implemented.

richturner:
@Ekhorn worth looking at AssetModelTest in particular the `ValueUtil.getSchema(AgentLink.class)` currently when I need to update the agent link schema baked into `or-attribute-input` I put a breakpoint at this line and copy/paste the value but this method is `ValueUtil` is where you can utilise dynamic/ondemand schema generation.

richturner:
Need to deal with `i18n` for schema, might need to look to use keys in the schema which can be looked up by translation runtime

Ekhorn:
### Description

Current findings show that the Agent linking implementation already uses technologies that are viable for the Dynamic JSON Forms. Though the code is solely focused on supporting Agent Linking and should be made more generic to support any possible configuration.

The JSON Schema standard is widely adopted and many code editors use it to validate data. Using JSON Schema as single source of truth is already in use and it would be a waste to step away from that. Validation on the frontend is already done using Ajv and on the backend https://beanvalidation.org/, replacing the bean validation may be done with a dedicated JSON schema validator that supports POJOs as Rich pointed out.

### Standards & Technologies

**Already used**

- JSON Schema -> Single source of truth
- Jackson -> Generate JSON schemas
[...]

PR #1965, opened 2025-06-27
Feature: Make JSON Forms dynamic and improve UX

### LINKED ISSUE #1768: JSON Forms: Investigate performance issues
## User Story
As a integrator I want the UI to feel snappy, so that I can quickly configure the assets I want.

## Description

The JSON Forms re-renders for each input present even though only 1 value is changed.

## Acceptance Criteria
- [ ] 1. The frontend inputs render faster

Ekhorn:
## Description

Closes #1768 #1767 #1766 #1747 #1704

## Checklist


ui/component/or-attribute-input/src/index.ts
```diff
@@ -118,120 +114,78 @@ export function getHelperText(sending: boolean, error: boolean, timestamp: numbe
...
-                    id: "",
-                    type: "DefaultAgentLink"
-                };
+                // Need to have a value so that the agent ID picker is shown
+                if (valueDescriptor.name === "agentLink") {
+                    value = {
+                        id: "",
+                        type: "DefaultAgentLink"
+                    };
+                }
```
Ekhorn: It would be nice if this could be set as `default` in the root of the JSON Schema to be able to remove this here and make it generically applicable. The only issue is that I haven't found a way to get this to work on polymorphic types.

ui/component/or-attribute-input/src/index.ts
```diff
@@ -118,120 +114,78 @@ export function getHelperText(sending: boolean, error: boolean, timestamp: numbe
...
-                    if (!matchedAgent) {
-                        console.warn("Agent link: linked agent could not be found");
-                        link.id = "";
-                    }
-                } else {
-                    link.id = "";
-                }
+            const cacheKey = `${valueDescriptor.type}:${valueDescriptor.arrayDimensions}`;
+            if (!schema && !schemas.has(cacheKey)) {
+                const response = await manager.rest.api.AssetModelResource.getValueDescriptorSchema({
```
Ekhorn, line 156: Make requesting independent from the component like the cache itself so we can block all UI elements and request each schema only once (it now sometimes makes a second request for the same schema).

manager/src/main/java/org/openremote/manager/asset/AssetModelResourceImpl.java
```diff
@@ -64,4 +71,16 @@ public Map<String, ValueDescriptor<?>> getValueDescriptors(RequestParams request
     public Map<String, MetaItemDescriptor<?>> getMetaItemDescriptors(RequestParams requestParams, String parentId) {
         return assetModelService.getMetaItemDescriptors(parentId);
     }
+
+    @Override
+    public Response getValueDescriptorSchema(RequestParams requestParams, String version, String descriptorType, Integer arrayDimensions) {
```
Ekhorn, line 76: Make sure schemas are "Gzipped" (at a certain size). Work in progress currently, trying to compare impact on the server and impact on high bandwidth vs low bandwidth connections.

Ekhorn: I think GZIP compression is currently the best, because off wide browser adoption, faster compression (than brotli) and decent compression ratios (see https://speedvitals.com/blog/zstd-vs-brotli-vs-gzip/).

We currently use gzip on the deployment folder and mbtiles, but not really anywhere else. For the deployment folder we implement a servlet response class with a private `GzipThresholdOutputStream` sub class which would be nice if we could somehow reuse this for any ordinary endpoint like here for instance.

model/src/main/java/org/openremote/model/value/ForecastConfigurationWeightedExponentialAverage.java
```diff
@@ -49,13 +51,17 @@ public class ForecastConfigurationWeightedExponentialAverage extends ForecastCon
     @NotNull
     @JsonSerialize(using = ToStringSerializer.class)
     @JsonDeserialize(converter = PeriodAndDurationConverter.class)
+    @JsonSchemaInject(merge = false, jsonSupplierViaLookup = JSONSchemaUtil.SCHEMA_SUPPLIER_NAME_STRING_TYPE)
+    // TODO: consider @JsonSchemaFormat("duration") requires new or-mwc-input type
```
Ekhorn: Consider creating an issue in case we don't want to implement this field in this PR.

model/src/main/java/org/openremote/model/value/ValueConstraint.java
```diff
@@ -70,6 +72,16 @@ public abstract class ValueConstraint implements Serializable {
     public static final String NOT_EMPTY_MESSAGE_TEMPLATE = "{ValueConstraint.NotEmpty.message}";
     public static final String NOT_BLANK_MESSAGE_TEMPLATE = "{ValueConstraint.NotBlank.message}";
     public static final String NOT_NULL_MESSAGE_TEMPLATE = "{ValueConstraint.NotNull.message}";
+    public static final String TIME_CONSTRAINT_EXAMPLES = """
```
Ekhorn: Should consider only listing values.

ui/component/or-json-forms/src/controls/control-array-element.ts
```diff
@@ -303,7 +304,13 @@ export class ControlArrayElement extends ControlBaseElement {
                         <div id="type-list" class="col">
                             <or-mwc-list @or-mwc-list-changed="${(evt: OrMwcListChangedEvent) => {if (evt.detail.length === 1) onParamChanged((evt.detail[0] as ListItem).data as CombinatorInfo); }}" .listItems="${listItems.sort((a, b) => a.text!.localeCompare(b.text!))}" id="parameter-list"></or-mwc-list>
                         </div>
-                        <div id="parameter-desc" class="col"></div>
+                        <div id="parameter-desc" class="col">
+                            <or-translate id="parameter-title" value="${selectedItemInfo?.title}"></or-translate>
+                            <p>${selectedItemInfo?.description}</p>
+                            ${when(selectedItemInfo && selectedItemInfo.examples.length > 0, () => html`<ul>
+                                <or-mwc-list .type="${ListType.PLAIN}" .listItems="${selectedItemInfo?.examples.sort((a, b) => a.localeCompare(b)).map((s) => ({ text: s }))}" id="examples-list"></or-mwc-list>
```
Ekhorn: No need to use `or-mwc-list` to render these items. Use standard HTML list items (`li`).

ui/app/shared/locales/ar/or.json
Ekhorn, line 1: Script should not affect existing translation values

ui/app/shared/locales/cn/or.json
Ekhorn, line 1: (that will only work if we can load the asset model separate from the server)

model/src/main/java/org/openremote/model/value/ValueConstraint.java
```diff
@@ -732,6 +728,10 @@ public String toString() {
         }
     }
 
+    @JsonSchemaInject(json =
+    """
+        { "i18n": "schema.meta.constraint.message" },
+    """)
     protected String message;
```
Ekhorn: A custom annotation to set i18n keys would be nice.

ui/app/shared/locales/ar/or.json
```diff
@@ -212,8 +212,8 @@
   "buttonTextConfirm": "نص لزر التفعيل",
   "buttonTextDecline": "نص لزر عدم التفعيل",
   "custom": "إدخال يدوي",
-  "realm": "نطاق",
-  "realm_plural": "النطاقات",
+  "realm": "عالم دنيا",
+  "realm_plural": "عالم دنيا",
```
Ekhorn, line 216: The script should not be affecting existing values.

ui/app/shared/locales/cn/or.json
```diff
@@ -225,8 +225,8 @@
   "buttonTextConfirm": "确认按键文本",
   "buttonTextDecline": "拒绝按键文本",
   "custom": "自定义参数",
-  "realm": "租户",
-  "realm_plural": "租户",
```
Ekhorn, line 229: The script should not be affecting existing values.

model/src/main/java/org/openremote/model/value/ValueConstraint.java
Ekhorn: A custom annotation like `JSONSchemaExamples` would be nice (but doesn't currently work).

Ekhorn: These examples are relevant for the attribute value not the meta item value. That might be a bit confusing determine how to deal with this.

model/src/main/java/org/openremote/model/value/ValueConstraint.java
Ekhorn: Planning on simple title and description annotations that a processors can hook into that generates translate keys base on the fqcn.

PR #1967, opened 2025-06-27
Feature: add user tag

Hackerberg43:
## Description
As an intermediate solution until the teams feature will be developed, this PR adds the possibility to give a "Tag" to every user.
This can be used to store which company the account belongs to for example and sort/search based on this tag.
Very simple implementation, but will help us alot while we wait for teams :)

afbeelding

It uses the allready existing attributes property from keycloak:

afbeelding

### DIFF
--- a/ui/component/or-icon/build.gradle
+++ b/ui/component/or-icon/build.gradle
@@ -13,6 +13,12 @@ tasks.register('clean') {
     }
 }
 
+tasks.register('npmAnalyze', Exec) {
+    mustRunAfter npmClean
+    dependsOn getYarnInstallTask()
+    commandLine npmCommand("yarn"), "run", "analyze"
+}
+
 tasks.register('prepareUi') {
     dependsOn clean, npmAnalyze, npmPrepare
 }
--- a/manager/src/main/java/org/openremote/manager/web/ManagerWebService.java
+++ b/manager/src/main/java/org/openremote/manager/web/ManagerWebService.java
@@ -94,10 +94,11 @@ public class ManagerWebService extends WebService {
     public static final String API_PATH = "/api";
     public static final String MANAGER_APP_PATH = "/manager";
     public static final String INSIGHTS_APP_PATH = "/insights";
+    public static final String STORYBOOK_APP_PATH = "/storybook";
     public static final String SWAGGER_APP_PATH = "/swagger";
     public static final String SHARED_PATH = "/shared";
 
-    public static final List<String> APP_PATHS = List.of(MANAGER_APP_PATH, INSIGHTS_APP_PATH, SWAGGER_APP_PATH, SHARED_PATH);
+    public static final List<String> APP_PATHS = List.of(MANAGER_APP_PATH, INSIGHTS_APP_PATH, STORYBOOK_APP_PATH, SWAGGER_APP_PATH, SHARED_PATH);
 
     public static final String UI_CLASSPATH_PREFIX = "org/openremote/web";
 
--- a/project.gradle
+++ b/project.gradle
@@ -355,10 +355,6 @@ tasks.register('npmClean', Exec) {
     dependsOn getYarnInstallTask()
     commandLine npmCommand("yarn"), "run", "clean"
 }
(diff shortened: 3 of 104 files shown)


MartinaeyNL:
Something outside of the box I'm thinking of when I see this;

Couldn't we expand this functionality by supporting multiple Keycloak user attributes?
For example, you specify is an array of attributes (like `['tag']`) in the `manager_config.json` file,
whereafter it will create a column for each attribute in that config;

| Username | First name | Surname       | Email            | Status  | Attribute1 | Another attribute | Tag attribute |
|----------|------------|---------------|------------------|---------|-------------|-------------------|---------------|
| admin    | System     | Administrator |                  | Enabled |             |                   | Company1      |
| user2    | Second     | User          | custom@email.com | Enabled |             | Special value     |               |
| user3    | Third      | User          |                  | Enabled |             |                   |               |

Because of the limited width on mobile phones, you'd leave those attribute columns out.
And a setup without Keycloak would not add any of these columns.

Hackerberg43:
Well yes, if you see an added value in more fields, shouldn't be too much effort.

ebariaux:
For now, decided that we'll only add the "tag" attribute from Keycloak, not going for anything more generic at this stage.

Hackerberg43:
@Ekhorn General user frontend tests were added. These verify:
-the search functionality in both regular and service user panels
-adding tags and filtering by them
-the inability to edit system managed gateway service users (which was introduced recently)

Fixed two bugs that the tests exposed.

Should be ready now.

Ekhorn:
Great! I'll have a look this afternoon.

ui/component/or-icon/build.gradle
```diff
@@ -3,7 +3,7 @@ buildDir = "dist"
 tasks.register('copyMdiFont', Exec) {
     dependsOn getYarnInstallTask()
     // Using yarn throws an error relating to node_module resolution
-    commandLine npmCommand("npm"), "run", "copyMdiFont"
+    commandLine 'bash', '-c', 'npm run copyMdiFont'
```
Ekhorn: Forcing bash would not work on Windows (unless using WSL 😉). Any reason for this? Or can it be reverted?

```suggestion
    commandLine npmCommand("npm"), "run", "copyMdiFont"
```

ui/app/manager/test/users.test.ts
```diff
@@ -0,0 +1,111 @@
+import { expect } from "@openremote/test";
+import { adminStatePath, test as setup, userStatePath } from "./fixtures/manager";
```
Ekhorn: Please rename to `test` as this isn't part of the setup.

```suggestion
import { adminStatePath, test, userStatePath } from "./fixtures/manager";
```

ui/app/manager/test/users.test.ts
```diff
@@ -0,0 +1,111 @@
...
+import { users } from "./fixtures/data/users";
+import permissions from "./fixtures/data/permissions";
```
Ekhorn: Looks unused

```suggestion
import { adminStatePath, test } from "./fixtures/manager";
```

ui/app/manager/test/users.test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+setup(`Login as "admin" user`, async ({ page, manager, context }) => {
+  await manager.goToRealmStartPage("master");
+  await manager.login("admin");
+  await page.waitForURL("**/manager/**");
+  await page.waitForTimeout(2000);
+  await context.storageState({ path: adminStatePath });
+});
+
+setup.describe(async () => {
```
Ekhorn: `Login as "admin" user` shouldn't be needed as it is already part of the `test.setup.ts` file. Also the describe block can be removed.


```suggestion
```

ui/component/or-icon/build.gradle
Hackerberg43: Oops, that a personal tweak ;)

MartinaeyNL (approved):

PR #1968, opened 2025-06-27
Configuration, script and documentation to deploy under AWS EKS

ebariaux:


kubernetes/eks-setup.sh
```diff
@@ -0,0 +1,176 @@
+#!/bin/bash
```
wborn, line 1: It might be helpful to use `set -eo pipefail` with the Bash scripts. That way issues will be easier to debug as the execution will stop whenever unexpected errors occur.

kubernetes/eks-setup-haproxy.sh
```diff
@@ -0,0 +1,131 @@
...
+
+# TODO: --name $CLUSTER_NAME -> in cluster.yaml
+# TODO: region is duplicated in cluster.yaml
+eksctl create cluster -f cluster.yaml --profile or
+
+# TODO: maybe extract cluster name after cluster creation so source of truth is in cluster.yaml, not other way around
+# Not straightforward
+
+# TODO: it might be possible to replace some of the operations below by configuration options in cluster.yaml
+oidc_id=$(aws eks describe-cluster --profile or --name $CLUSTER_NAME --query "cluster.identity.oidc.issuer" --output text | cut -d '/' -f 5)
```
wborn, line 13: It might be a good idea to also replace all the other backtick \`commands\` in the scripts with  `$(commands)` as well, see:

[Why is $(...) preferred over \`...\` (backticks)?](http://mywiki.wooledge.org/BashFAQ/082)

kubernetes/eks-setup.sh
ebariaux: Implemented

kubernetes/eks-setup-haproxy.sh
ebariaux: Implemented

kubernetes/eks-setup.sh
```diff
@@ -0,0 +1,150 @@
...
+HOSTED_ZONE_ID=$(aws elbv2 describe-load-balancers --profile or --query "LoadBalancers[?DNSName=='$DNS_NAME'].CanonicalHostedZoneId | [0]")
+
+aws route53 change-resource-record-sets \
+    --hosted-zone-id /hostedzone/Z08751721JH0NB6LLCB4V \
+    --change-batch \
+     '{"Changes": [ { "Action": "UPSERT", "ResourceRecordSet": { "Name": "'$MQTT_FQDN'", "Type": "A", "AliasTarget":{ "HostedZoneId": '$HOSTED_ZONE_ID',"DNSName": '\"$DNS_NAME\"',"EvaluateTargetHealth": false} } } ]}' \
+     --profile dnschg
+
+# We're re-directing the MQTTS_FQDN to its network load balancer
+DNS_NAME=$(kubectl get svc manager-mqtts -o jsonpath="{.status.loadBalancer.ingress[0].hostname}")
```
wborn: While testing the `DNS_NAME` was an empty string causing the subsequent while loop to never exit:

`Waiting for network load balancer to be created for ...`

To fix that you might want to change the code so it waits until the command is not empty or add a quick fix like `sleep 3s`

PR #1970, opened 2025-07-02
Add JVM metrics to prometheus endpoint

ebariaux:


container/src/main/java/org/openremote/container/Container.java
```diff
@@ -32,13 +32,31 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Handler;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
-import static java.lang.System.Logger.Level.*;
+import static java.lang.System.Logger.Level.ERROR;
+import static java.lang.System.Logger.Level.INFO;
+import static java.lang.System.Logger.Level.WARNING;
```
richturner: We need to enforce code styling rules (branch pending) but for now let's not have files with only import/style changes

richturner (changes requested):

PR #1972, opened 2025-07-03
MQTT agent wildcard topic support 

### LINKED ISSUE #2014: Add MQTT Agent wildcard topic subscription support
## User Story
As an MQTT agent user it would be nice to specify wilcard topics to subscribe to and then attributes that are linked to the agent would be able to consume matching messages from these wildcard subscriptions in order to minimise the number of subscriptions created.

## Acceptance Criteria
- MQTT agent should have an attribute to supply an array of wilcard topics to subscribe to
- New MQTT subscriptions are not created if the subscription topic of a linked attribute matches one of the MQTT agent wildcard topics. If the topic doesn't match the MQTT subscription is created
- Must be able to have multiple linked attributes with the same topic that match a wildcard topic

rainerhitz:
Closes #2014


agent/src/main/java/org/openremote/agent/protocol/mqtt/AbstractMQTT_IOClient.java
```diff
@@ -285,26 +289,26 @@ public void sendMessage(MQTTMessage<S> message) {
...
+        addMessageConsumer("#", true, null, messageConsumer);
     }
 
     public void addMessageConsumer(String topic, Consumer<MQTTMessage<S>> messageConsumer) {
-        addMessageConsumer(topic, subscribeQos, messageConsumer);
+        addMessageConsumer(topic, true, subscribeQos, messageConsumer);
     }
 
-    public void addMessageConsumer(String topic, MqttQos qos, Consumer<MQTTMessage<S>> messageConsumer) {
+    public void addMessageConsumer(String topic, boolean isSubscription, MqttQos qos, Consumer<MQTTMessage<S>> messageConsumer) {
```
richturner: I can see why you've done it this way but I'm not keen on this `isSubscription` logic being pushed down to the IO client, this seems like a higher level concept that should be dealt with by users of this IO client (i.e. the protocol)

agent/src/main/java/org/openremote/agent/protocol/mqtt/MQTTAgent.java
```diff
@@ -52,6 +57,7 @@ public class MQTTAgent extends IOAgent<MQTTAgent, MQTTProtocol, MQTTAgentLink> {
     public static final AttributeDescriptor<String> LAST_WILL_TOPIC = new AttributeDescriptor<>("lastWillTopic", ValueType.TEXT);
     public static final AttributeDescriptor<String> LAST_WILL_PAYLOAD = new AttributeDescriptor<>("lastWillPayload", ValueType.TEXT);
     public static final AttributeDescriptor<Boolean> LAST_WILL_RETAIN = new AttributeDescriptor<>("lastWillRetain", ValueType.BOOLEAN);
+    public static final AttributeDescriptor<String> WILDCARD_SUBSCRIPTION_TOPICS = new AttributeDescriptor<>("wildcardSubscriptionTopicList", ValueType.TEXT, new MetaItem<>(MetaItemType.MULTILINE));
```
richturner: You can have array types i.e. `ValueType.TEXT.asArray()` see `MetaItemType.UNITS`

agent/src/main/java/org/openremote/agent/protocol/mqtt/MQTTAgent.java
```diff
@@ -179,4 +185,26 @@ public MQTTAgent setSubscribeQos(int subscribeQos) {
...
+    public MQTTAgent setWildcardSubscriptionTopicList(List<String> wildcardList) {
+        String wildcards = Optional.ofNullable(wildcardList)
+            .orElseGet(List::of)
+            .stream()
+            .map(String::trim)
+            .filter(s -> !s.isEmpty())
+            .collect(Collectors.joining("\n"));
+        getAttributes().getOrCreate(WILDCARD_SUBSCRIPTION_TOPICS).setValue(wildcards);
+        return this;
+    }
```
richturner: Use array type

agent/src/main/java/org/openremote/agent/protocol/mqtt/MQTTProtocol.java
```diff
@@ -64,7 +64,8 @@ protected void doLinkAttribute(String assetId, Attribute<?> attribute, MQTTAgent
                     updateLinkedAttribute(new AttributeRef(assetId, attribute.getName()), msg.payload);
                 }
             };
-            client.addMessageConsumer(topic, Optional.of(agentLink.getQos().orElse(agent.getSubscribeQoS().orElse(0))).map(qos -> qos > 2 || qos < 0 ? null : qos).map(MqttQos::fromCode).orElse(MqttQos.AT_MOST_ONCE), messageConsumer);
+            boolean isSubscription = !matchWildcardTopicList(topic);
```
richturner: `isSubscription` rename to `matchesWildcardSubscription` and then you don't need to call `client.addMessageConsumer` but add to a topic consumer hashmap here in the protocol and handle routing in the wildcard messageConsumer.

agent/src/main/java/org/openremote/agent/protocol/mqtt/MQTTProtocol.java
```diff
@@ -161,4 +176,42 @@ protected MQTTMessage<String> createWriteMessage(MQTTAgentLink agentLink, Attrib
...
+                continue;
+            }
+
+            if (!w.equals(topicLevels[i])) {
+                return false;
+            }
+        }
+
+        return i == topicLevels.length;
+    }
```
richturner: Maybe this could be added as a helper method to the existing `Topic` class?

agent/src/main/java/org/openremote/agent/protocol/mqtt/MQTTProtocol.java
```diff
@@ -161,4 +176,42 @@ protected MQTTMessage<String> createWriteMessage(MQTTAgentLink agentLink, Attrib
     public String getProtocolName() {
         return "MQTT Client";
     }
+
+    private boolean matchWildcardTopicList(String topic) {
+        return getAgent().getWildcardSubscriptionTopicList()
```
richturner: Store `List<Topic>` (see below comment about reusing the Topic class)

rainerhitz: Currently, you cannot use the 'org.openremote.manager.mqtt.Topic' class because it would create a circular dependency.

richturner:
Hi @rainerhitz wasn't sure if you were on holiday and wanted to get this finished off so I made a few changes to re-use the `Topic` class (moved into model) and also got rid of the WildcardSubscriptionManager to keep things simpler, please feel free to look at my changes and feedback always welcome.

richturner (approved):

PR #1973, opened 2025-07-03
Fix for missing ppm unit translation (English & Dutch)

Hackerberg43:
ppm units where displayed as units.ppm because of missing translation in or.json


MartinaeyNL:
Thanks for resolving this! 👍

Issue #1686, opened 2025-02-19
Services: Develop API for services


DonWillems:

dominiquekleeven:
Feature has been implemented in the `feature/services-page` branch, branch needs some additional clean up before opening up the PR as it currently contains docker-compose changes for deployment purposes.

PR #1979, opened 2025-07-04
Feature/Added services page and resource API

dominiquekleeven:
Draft PR for #1686 and #1686

- Added registry mechanism for microservices/external services
- Added services page
- Added integration tests for the microservice resource and registry service

***
Branch still needs clean up (contains some docker changes for testing, which need to be removed/reset)

WIP


dominiquekleeven:
Will deploy this to `test3` for further testing

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            LOG.warning("Failed to send heartbeat to microservice: " + e.getMessage());
+            throw new InternalServerErrorException("Failed to send heartbeat to microservice");
+        }
+    }
+
+    @Override
+    public Response deregisterService(RequestParams requestParams, String serviceId, String instanceId) {
+        if (!isServiceAccount()) {
+            LOG.warning("Only service accounts can deregister services");
+            throw new ForbiddenException("Only service accounts can deregister services");
```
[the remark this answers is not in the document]
dominiquekleeven: Other resources throw a forbidden exception in this scenario, so leaving as is to ensure consistency

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        return Arrays.stream(services)
+                .map(MicroserviceInfo::fromMicroservice)
+                .toArray(MicroserviceInfo[]::new);
+    }
+
+    @Override
+    public Response sendHeartbeat(RequestParams requestParams, String serviceId, String instanceId) {
+        if (!isServiceAccount()) {
+            LOG.warning("Only service accounts can send heartbeats");
+            throw new ForbiddenException("Only service accounts can send heartbeats");
```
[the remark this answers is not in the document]
dominiquekleeven: Other resources throw a forbidden exception in this scenario, so leaving as is to ensure consistency

dominiquekleeven:
Close to being ready for review, cleaning up the branch tomorrow to not include any deployment additions (e.g. docker-compose changes, .env changes).

dominiquekleeven:
Couple remaining issues regarding the front-end implementation:

- [x] Services page should be responsive similar as to how assets/insights are handled
  - Added media queries and dynamic rendering based on whether the client has a mobile width. It behaves exactly the same as insights and assets in terms of displaying the sidebar and selecting a tree entry + navigating back.
- [x] Refresh button placement in the header of the services page sidebar isn't properly centered and aligned
- [x] Unavailable services should not be able to be selected, or should provide some form of error message in the iframe area when selected
- [x] Directly navigating to the services page with the specified service-id as part of the URL does not work as intended, it should open the services page and have the service selected if its present.
- [x] Some explanation test when no services are present, e.g. link to docs with guide on adding a service/registering a service.
- [x] Missing translation for page title
- [x] The `or-iframe` component `onerror` event does not get triggered when the host is fully unreachable, e.g. URL doesn't lead anywhere, this causes the `or-iframe` component to get stuck in its loading state.
  - Resolved by adding a timeout mechanism and also ensured the iframe contents are hidden until actually loaded to prevent showing a the browsers HTTP/Host unreachable page.

dominiquekleeven:
Deployed on `test3` alongside the ML Forecast Service (https://github.com/openremote/service-ml-forecast/pull/40 version)

ui/component/or-services/src/index.ts
```diff
@@ -0,0 +1,368 @@
...
+    const event = new CustomEvent("service-selected", {
+      detail: {
+        service: service,
+        serviceId: service.serviceId
+      }
+    });
+    this.dispatchEvent(event);
+    
+    // On mobile, hide the tree when a service is selected
+    if (window.matchMedia("(max-width: 600px)").matches) {
```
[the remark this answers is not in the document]
dominiquekleeven: Same as or-dashboard, just be fine to leave it as is.

ui/component/or-services/src/or-service-tree.ts
```diff
@@ -0,0 +1,144 @@
...
+    return html`
+      <div id="tree-header">
+        <h3 id="tree-header-title">
+          <or-translate value="services.title"></or-translate>
+        </h3>
+        <div class="hideMobile">
+          <or-mwc-input
+            type=${InputType.BUTTON}
+            icon="refresh"
+            title="Refresh services"
```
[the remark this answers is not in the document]
dominiquekleeven: Resolved

ui/app/manager/src/headers.ts
```diff
@@ -37,6 +37,15 @@ export function headerItemInsights<S extends AppStateKeyed, A extends AnyAction>
     };
 }
 
+export function headerItemServices<S extends AppStateKeyed, A extends AnyAction>(orApp: OrApp<S>): HeaderItem {
+    return {
+        icon: "puzzle",
+        href: "services",
+        text: "Services",
+        roles: () => !manager.hasRealmRole("restricted_user")
```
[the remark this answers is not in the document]
dominiquekleeven: Resolved

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
```diff
@@ -0,0 +1,125 @@
...
+        super(timerService, identityService);
+        this.microserviceRegistry = serviceRegistry;
+    }
+
+    @Override
+    public Microservice registerService(RequestParams requestParams, Microservice microservice) {
+        if (!isServiceAccount()) {
+            LOG.warning("Service registration not allowed for non-service users");
+            throw new WebApplicationException("Service registration not allowed for non-service users",
+                    Response.Status.FORBIDDEN);
```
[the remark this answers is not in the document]
dominiquekleeven: Indeed, its a credential type mismatch, therefore considered an authentication check.

I've changed the response codes to 401 (UNAUTHORIZED)

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,272 @@
...
+        registry.values().stream()
+                .flatMap(List::stream)
+                .filter(entry -> entry.getLeaseInfo().isExpired(currentTime)
+                        && entry.getStatus() == MicroserviceStatus.AVAILABLE)
+                .forEach(entry -> entry.setStatus(MicroserviceStatus.UNAVAILABLE));
+    }
+
+    protected void cleanupExpiredUnavailableInstances() {
+        long currentTime = timerService.getCurrentTimeMillis();
+        long purgeThreshold = currentTime + DEFAULT_PURGE_UNAVAILABLE_MS;
```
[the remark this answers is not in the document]
dominiquekleeven: Yes, indeed a bug. 

Logic here:
- get current timestamp
- subtract the purge threshold e.g. 24 hours
- iterate through the services and collect any that are unavailable and have an expiration timestamp that is older than the purge timestamp.
- de-register the collected services

dominiquekleeven:
Will update `test3` tomorrow with the latest changes, cleaning up the PR e.g. remove any unrelated files, and changing to status to Ready for review

dominiquekleeven:
Deployed on `test3` alongside the ML Forecast Service (https://github.com/openremote/service-ml-forecast/pull/40 version)

dominiquekleeven:
@MartinaeyNL Requested you as a reviewer, mainly for the front-end additions/changes. Feel free to reach out if I need to elaborate something.

ui/component/or-services/src/or-service-tree.ts
```diff
@@ -0,0 +1,164 @@
...
+    // Only select if ONE is selected
+    const newSelected = !nodes || nodes.length > 1 ? undefined : nodes[0];
+
+    if (newSelected?.service) {
+      // Dispatch a custom event that the page can listen to
+      const event = new CustomEvent("service-selected", {
+        detail: {
+          service: newSelected.service,
+        },
+      });
```
MartinaeyNL: We tend to create interfaces for dispatching a `CustomEvent`, and use the `or-` prefix as well.
You'd have something similar as the "select" event for `or-tree-menu`;
```typescript
/**
 * Model for the SELECT event that {@link OrTreeMenu} can dispatch.
 * Once a node is selected, a list of all the selected nodes will be shared with the consumer elements.
 */
export class OrTreeSelectEvent extends CustomEvent<TreeNode[]> {

    public static readonly NAME = "or-tree-select";

    constructor(nodes: TreeNode[]) {
        super(OrTreeSelectEvent.NAME, {
            bubbles: true,
            composed: true,
            detail: nodes
        });
    }
}
```
Mainly meant for type-safety within our own codebase.

For documentation purposes, can you also add an `@event` decorator on the component?
It is recognized by the `custom-elements.json` generator we use, and can be interpreted by documentation sites.
For example;
```typescript
/**
 * @event {OrTreeSelectEvent} or-tree-select - Triggers upon selecting a node, and dispatches a list of the nodes selected.
 * @event {OrTreeDragEvent} or-tree-drag - Triggers upon dragging a node to a new group, and dispatches a list of dragged nodes, the group node, and the updated list of all nodes.
 */
@customElement("or-tree-menu")
export class OrTreeMenu extends LitElement {
// ...
```

ui/component/or-services/src/or-service-tree.ts
```diff
@@ -0,0 +1,164 @@
...
+            title="${i18next.t("services.refresh")}" 
+            @or-mwc-input-changed=${this._onRefreshServices}
+          ></or-mwc-input>
+        </div>
+      </div>
+    `;
+  }
+
+  protected _onRefreshServices(): void {
+    this.dispatchEvent(new CustomEvent("refresh-services", { detail: {} }));
```
MartinaeyNL: Same comment here in regards to a custom event

ui/component/or-services/src/index.ts
```diff
@@ -0,0 +1,417 @@
...
+    or-iframe {
+      flex: 1;
+      max-width: calc(100vw - 300px);
+    }
+  }
+`;
+
+export interface ServicesConfig {
+  // Configuration options for the services component
+}
```
MartinaeyNL: This is unused I think?

ui/component/or-components/src/or-iframe.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  @property({ type: String })
+  public src?: string;
+
+  @property({ type: Number })
+  public timeout = 10000; // 10 seconds default timeout
+
+  @property({ type: String })
+  public loadErrorMessage?: string;
```
MartinaeyNL: I think we should replace this HTML attribute with a `<slot>` element.
So, instead of a `<span>${this.loadErrorMessage}</span>`, you'd use `<slot name="onerror"></slot>`.

It gives the consumer flexibility what to display when an error occurs.
Whether it's a `<span>`, or a button with a "retry" behavior, it can be defined like this;
```
<or-iframe src="">
    <span slot="onerror">My custom error message</span>
</or-iframe>

// or

<or-iframe src="">
    <button slot="onerror" @click="...">Try again</span>
</or-iframe>
```

ui/component/or-components/src/or-iframe.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  @state()
+  private loading = true;
+
+  @state()
+  private loadingError = false;
```
MartinaeyNL: I'd keep this generic and use `error` instead of `loadingError`.

ui/component/or-components/src/or-iframe.ts
```diff
@@ -0,0 +1,217 @@
...
+      src: this.src,
+      error:
+        type === OrIFrameEventType.ERROR || type === OrIFrameEventType.TIMEOUT
+          ? type === OrIFrameEventType.TIMEOUT
+            ? "Timeout loading iframe"
+            : "Error loading iframe"
+          : undefined,
+    };
+
+    this.dispatchEvent(new CustomEvent("or-iframe-event", { detail }));
```
MartinaeyNL: Similar comment to the one in the `or-services` component, in regards to documenting the event and making it type-safe.

ui/component/or-components/src/or-iframe.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    switch (type) {
+      case OrIFrameEventType.LOADED:
+        this.loadingError = false;
+        this.loading = false;
+        this.requestUpdate();
```
MartinaeyNL: I don't think this `requestUpdate()` is necessary,
as you're already updating the state with `this.loading = false;` and `this.loadingError = false;`.
We only tend to use it when modifying a part of the state;
```typescript
this.loading = false; // full update, no requestUpdate() necessary
this.list.push(...); // not caught as an update, we tend to add requestUpdate() here.
```

model/src/main/java/org/openremote/model/microservices/MicroserviceResource.java
```diff
@@ -0,0 +1,118 @@
...
+     */
+    @POST
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @RolesAllowed({ Constants.WRITE_SERVICES_ROLE })
+    @Operation(operationId = "registerService", summary = "Register an external service/microservice", responses = {
+            @ApiResponse(responseCode = "200", description = "Service registered successfully", content = @Content(mediaType = "application/json", schema = @Schema(implementation = Microservice.class))),
+            @ApiResponse(responseCode = "400", description = "Invalid microservice object"),
+            @ApiResponse(responseCode = "409", description = "Microservice instance already registered"),
+    })
```
MartinaeyNL: I see that these also show up in Swagger documentation, great!
Reminds me that I should keep an eye out, that other endpoints are also documented like this.

ui/component/or-services/src/index.ts
MartinaeyNL, line 1: Can you use an indentation of 4 n `.ts` files, like the other components?
(an automatic linter is being worked on 😂 )

MartinaeyNL, line 1: I was testing the UI, and i tend to press the "reload services" button, just "to double check if everything is loaded".
When everything is OK, there is nothing happening in the UI.

Maybe we should show a simple snackbar saying; "Loaded X services!", to indicate that 'something' is going on?

MartinaeyNL (commented):
I'm overall very happy with the code 👍 
Mostly went over the frontend related code, but also had a quick look at the other code.
Let me know if you want to go over stuff in a call.

I also think some slight changes in CSS could improve the looks of this page.
But I think you have a chat scheduled on that next week, I believe?

ui/component/or-services/src/index.ts
dominiquekleeven: Yes unused, forgot to remove.

ui/component/or-components/src/or-iframe.ts
dominiquekleeven: Replaced it with a slot element :)

ui/component/or-components/src/or-iframe.ts
dominiquekleeven: Changed to just error

ui/component/or-components/src/or-iframe.ts
dominiquekleeven: Removed it

model/src/main/java/org/openremote/model/microservices/MicroserviceResource.java
dominiquekleeven: Yea I think I saw these being used in another resource interface so basically used that as inspiration :)

ui/component/or-services/src/index.ts
dominiquekleeven: They all should be indented using 4 spaces now, I updated the preffier config we had in the repo to enforce so for those who use prettier it'll automatically enforce 4 spaces for identation when formatting.

dominiquekleeven: Good idea, I've added a snackbar for this and the translations

It's now `{{count}} service(s) retrieved`  / `{{count}} service(s) opgehaald`

ui/component/or-services/src/or-service-tree.ts
dominiquekleeven: Added interfaces for all the `new CustomEvent` instances, also added the @event decorator to the components

ui/component/or-services/src/or-service-tree.ts
dominiquekleeven: Added interfaces for all the `new CustomEvent` instances, also added the @event decorator to the components

ui/component/or-components/src/or-iframe.ts
dominiquekleeven: Added interfaces for all the `new CustomEvent` instances, also added the @event decorator to the components

dominiquekleeven:
@MartinaeyNL Thanks for the review!

I've addressed your comments and `test3` has been updated with the changes, feel free to review again when you have the time.

dominiquekleeven:
> I'm overall very happy with the code 👍 Mostly went over the frontend related code, but also had a quick look at the other code. Let me know if you want to go over stuff in a call.
> 
> I also think some slight changes in CSS could improve the looks of this page. But I think you have a chat scheduled on that next week, I believe?

Yes I will bring this up next week, we can maybe sit down and brainstorm about a couple changes that could be made to improve the looks, it feels quite empty/barebones at the moment.

dominiquekleeven:
Additional feedback from today
**Back-end:**
- [x] Registry should take into account multi-tenancy in the following way:

  - A service is globally registered if its registered on the master realm via a super user. These services are listed on the services page for all realms.
  - A service is realm specific if its registered for a specific realm with the write:service role. These services are only listed on the services page of the realm its registered for.

- [x] Service registry should also publish `MicroserviceEvent`s that for when a new service is registered, a lease is expired, a lease is refreshed, a service is deregistered.
  
**Front-end:**
- [x] The Services navigation entry should be moved after Rules.

- [x] The refresh iframe and open in new tab should be removed from the header when having a service selected

- [x] The `no services found` text should include a href to the documentation on how to add one

- [x] The refresh button for the services tree can be removed, we should use WebSockets instead for real-time status/new service registrations.

- [x] Add the ability to sort the list on `name` or `status`

***

dominiquekleeven:
Most of the feedback from monday has been addressed, integration tests need to be updated with the new multi tenancy logic for the registry and the event handling.

dominiquekleeven:
`test3` updated with latest changes

dominiquekleeven:
@richturner I’ve requested a review from you for this PR. The description contains the most notable changes along with some remarks and questions. I’d be happy to schedule a call to discuss or go over anything once you’ve had a chance to take a look

ui/app/manager/src/pages/page-services.ts
```diff
@@ -0,0 +1,344 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+import { css, html, TemplateResult, PropertyValues } from "lit";
+import { customElement, state } from "lit/decorators.js";
+import { Page, PageProvider, router } from "@openremote/or-app";
+import { AppStateKeyed } from "@openremote/or-app";
```
MartinaeyNL: You can merge these into one statement 😉

ui/app/manager/src/pages/page-services.ts
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    protected async _loadData(silent = false): Promise<void> {
+        if (!silent) {
+            this._loading = true;
+        }
+
+        try {
+            const realmServices = await manager.rest.api.MicroserviceResource.getServices({ realm: this.realmName });
+            const globalServices = await manager.rest.api.MicroserviceResource.getGlobalServices();
```
MartinaeyNL: I'd prefer to use something like `Promise.all()` which executes these HTTP requests in parallel.
On another note; if the global request fails, it "fails to load services", although realm services were fetched correctly.
Would be nice to do some minor tweaks here.

setup/src/integration/java/org/openremote/setup/integration/KeycloakTestSetup.java
```diff
@@ -83,6 +85,8 @@ public KeycloakTestSetup(Container container) {
     public void onStart() throws Exception {
         super.onStart();
 
+        ManagerIdentityProvider identityProvider = identityService.getIdentityProvider();
+
         // Realms
         realmMaster = identityService.getIdentityProvider().getRealm(Constants.MASTER_REALM);
```
MartinaeyNL: Seems like a duplicate pattern; we should between
- Caching `identityProvider` in an variable, or..
- use `identityService.getIdentityProvider()` throughout the setup code.

MartinaeyNL (commented):
Reviewed the UI code, and took a short look at the backend code.
Only had a few minor comments. Great stuff! 👍

setup/src/integration/java/org/openremote/setup/integration/KeycloakTestSetup.java
dominiquekleeven: I made sure the identityProvider variable is now used throughout the setup code, I didn't notice it before.

ui/app/manager/src/pages/page-services.ts
dominiquekleeven: I opted for `Promise.allSettled` according to docs it allows it to handle the individual promise failures, so it now runs the requests in parallel and handles the individual failures accordingly. It is very unlikely that one succeeds and the other fails, but just in case theres now error handling for that.

Let me know if this is alright :)

ui/app/manager/src/pages/page-services.ts
dominiquekleeven: Done :)

MartinaeyNL (commented):
The comments I had have been resolved. So I'm OK with these changes 👍 
Let's have Rich take a deeper look into the manager code and the overall solution.

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+
+            @SuppressWarnings("unchecked")
+            EventSubscription<MicroserviceEvent> subscription = (EventSubscription<MicroserviceEvent>) eventSubscription;
+            subscription.setFilter(event -> {
+                Microservice eventMicroservice = event.getMicroservice();
+                boolean isGlobalService = eventMicroservice.getIsGlobal();
+                boolean realmMatches = Objects.equals(eventMicroservice.getRealm(),
+                        authContext.getAuthenticatedRealmName());
+
+                return isGlobalService || realmMatches ? event : null;
```
richturner: This logic seems wrong, any user can receive events from global services which means data could be crossing realms?
Should regular users be able to subscribe to global services?
Shouldn't global services only be available to super users?

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+     *                     admin
+     *                     service user with global access.
+     */
+    public Microservice registerService(Microservice microservice, boolean isGlobal) {
+        LOG.info("Registering microservice: " + microservice.getServiceId() + ", instanceId: "
+                + microservice.getInstanceId());
+
+        if (microservice.getInstanceId() == null) {
+            microservice.setInstanceId(UniqueIdentifierGenerator.generateId());
+        }
```
richturner: Should be moved before logging statement

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+    protected static final long DEREGISTER_UNAVAILABLE_INTERVAL_MS = 3600000;
+
+    protected TimerService timerService;
+    protected ScheduledExecutorService scheduledExecutorService;
+    protected ManagerIdentityService identityService;
+    protected ClientEventService clientEventService;
+
+    // serviceId: <instanceId: Microservice>
+    // Nested ConcurrentHashMaps for segmenting, segment locks and fast lookups
+    protected ConcurrentHashMap<String, ConcurrentHashMap<String, Microservice>> registry;
```
richturner: Do we need the nesting if synchronisation is always done on the outer `ConcurrentHashMap`?

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+
+            return authContext.hasResourceRole(ClientRole.READ_SERVICES.getValue(),
+                    Constants.KEYCLOAK_CLIENT_ID);
+        });
+    }
+
+    @Override
+    public void start(Container container) throws Exception {
+        markExpiredInstancesAsUnavailableFuture = scheduledExecutorService.scheduleAtFixedRate(
+                this::markExpiredInstancesAsUnavailable, 0,
```
richturner: Do we need initialDelay 0, there's nothing to do?

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+                MARK_EXPIRED_INSTANCES_INTERVAL_MS, TimeUnit.MILLISECONDS);
+
+        deregisterExpiredUnavailableInstancesFuture = scheduledExecutorService.scheduleAtFixedRate(
+                this::deregisterLongExpiredInstances, 0,
```
richturner: Do we need initialDelay 0, there's nothing to do?

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * Register a microservice instance
+     *
+     * @param microservice The microservice to register
+     * @param isGlobal     Whether the service should be available to all realms.
+     *                     This should only be used for services that use a super
+     *                     admin
+     *                     service user with global access.
+     */
+    public Microservice registerService(Microservice microservice, boolean isGlobal) {
```
richturner: Microservice has a global property also not clear why we need a method param as well

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+                });
+
+        // publish the register event as a new microservice has been added to the
+        // registry
+        clientEventService.publishEvent(new MicroserviceEvent(microservice, MicroserviceEvent.Cause.REGISTER));
+
+        LOG.info("Successfully registered microservice: " + microservice.getServiceId() + ", instanceId: "
+                + microservice.getInstanceId() + ", isGlobal: " + microservice.getIsGlobal());
+
+        return microservice;
```
richturner: What purpose does returning the microservice object achieve? It's always the same instance, the null check in callers is also redundant as it cannot be null

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+
+    /**
+     * Get a specific microservice registration.
+     * 
+     * @param serviceId  The serviceId of the microservice to get
+     * @param instanceId The instanceId of the microservice to get
+     * @return The microservice
+     */
+    public Microservice getService(String serviceId, String instanceId) {
+        return registry.values().stream()
```
richturner: Not sure this is that efficient, a `get()` would be quicker with the hashcode lookup, wrap with `Optional.ofNullable()` if you want a functional approach.

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+     */
+    protected void markExpiredInstancesAsUnavailable() {
+        long now = timerService.getCurrentTimeMillis();
+        LOG.fine("Marking expired microservice instances as unavailable");
+
+        registry.entrySet().stream()
+                .flatMap(serviceEntry -> serviceEntry.getValue().entrySet().stream())
+                .map(Map.Entry::getValue)
+                // filter by lease expired and current status is still available
+                .filter(ms -> ms.getLeaseInfo().isExpired(now)
```
richturner: rename `isUnavailable(now)`

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+
+        long now = timerService.getCurrentTimeMillis();
+        long deregisterThreshold = now - DEFAULT_DEREGISTER_UNAVAILABLE_MS;
+
+        registry.entrySet().stream()
+                .flatMap(serviceEntry -> serviceEntry.getValue().entrySet().stream())
+                .map(Map.Entry::getValue)
+                // filter by unavailable + meets threshold
+                .filter(ms -> ms.getStatus() == MicroserviceStatus.UNAVAILABLE
+                        && ms.getLeaseInfo().getExpirationTimestamp() < deregisterThreshold)
```
richturner: use `isExpired(now)` method like unavailable handler

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
```diff
@@ -0,0 +1,383 @@
...
+            }
+
+            return instanceMap;
+        });
+    }
+
+    /**
+     * Get all registered services and microservices and their instances for the
+     * given
+     * realm.
```
richturner:
```suggestion
     * given realm.
```

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+
+    @Override
+    public Microservice registerService(RequestParams requestParams, Microservice microservice) {
+        if (!isServiceAccount()) {
+            LOG.warning("Service registration not allowed for non-service users");
+            throw new WebApplicationException("Service registration not allowed for non-service users",
+                    Response.Status.UNAUTHORIZED);
+        }
+
+        if (!isRealmActiveAndAccessible(microservice.getRealm())) {
```
richturner: Realm should be extracted from `requestParams` and then set on the microservice

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+            throw new WebApplicationException(
+                    "Realm '" + microservice.getRealm() + "' is nonexistent, inactive or inaccessible",
+                    Response.Status.FORBIDDEN);
+        }
+
+        try {
+            Microservice registeredMicroservice = microserviceRegistry.registerService(microservice);
+
+            if (registeredMicroservice == null) {
```
richturner: See previous comment about this `null` check

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+            throw new WebApplicationException("Service registration is only available for service users",
+                    Response.Status.UNAUTHORIZED);
+        }
+
+        if (!isSuperUser()) {
+            throw new WebApplicationException("Global services can only be registered by super users",
+                    Response.Status.FORBIDDEN);
+        }
+
+        if (!microservice.getRealm().equals(MASTER_REALM)) {
```
richturner: Again use realm from `requestParams`

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+            throw new WebApplicationException("Global services must have the realm set to the master realm, got: "
+                    + microservice.getRealm(),
+                    Response.Status.BAD_REQUEST);
+        }
+
+        try {
+            Microservice registeredMicroservice = microserviceRegistry.registerService(microservice, true);
+
+            if (registeredMicroservice == null) {
```
richturner: Same comment about `null` check

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+
+    @Override
+    public void heartbeat(RequestParams requestParams, String serviceId, String instanceId) {
+        if (!isServiceAccount()) {
+            LOG.warning("Service heartbeat is only available for service users");
+            throw new WebApplicationException("Service heartbeat is only available for service users",
+                    Response.Status.UNAUTHORIZED);
+        }
+
+        Microservice service = microserviceRegistry.getService(serviceId, instanceId);
```
richturner: Track which user ID registers a service and only allow that user to heartbeat and deregister, protects against rogue services affecting other services

manager/src/main/java/org/openremote/manager/setup/AbstractKeycloakSetup.java
```diff
@@ -52,11 +52,13 @@ public abstract class AbstractKeycloakSetup implements Setup {
...
         ClientRole.READ_INSIGHTS,
         ClientRole.READ_ALARMS,
+        ClientRole.READ_SERVICES,
         ClientRole.WRITE_ASSETS,
         ClientRole.WRITE_ATTRIBUTES,
         ClientRole.WRITE_RULES,
         ClientRole.WRITE_INSIGHTS,
-        ClientRole.WRITE_ALARMS
+        ClientRole.WRITE_ALARMS,
+        ClientRole.WRITE_SERVICES
```
richturner, line 61: Need to look at creating a java code migration file for flyway to add these roles in exsiting deployments (i.e. not a clean install)

model/src/main/java/org/openremote/model/microservices/Microservice.java
```diff
@@ -0,0 +1,259 @@
...
+    /**
+     * The unique identifier of the service (e.g., "energy-service").
+     * 
+     * This ID is used to group related service instances and must be consistent
+     * across all instances of the same service type.
+     */
+    @Size(min = 3, max = 255, message = "{Microservice.serviceId.Size}")
+    @NotEmpty(message = "{Microservice.serviceId.NotEmpty}")
+    @JsonProperty("serviceId")
+    protected String serviceId;
```
richturner: Could the username (without service user prefix) be used as the serviceId (would help with tracking users and their services)

model/src/main/java/org/openremote/model/microservices/Microservice.java
```diff
@@ -0,0 +1,259 @@
...
+    /**
+     * Unique identifier for this specific instance within the scope of the
+     * serviceId.
+     * 
+     * Each running instance of a service must have a unique instanceId to
+     * distinguish
+     * it from other instances of the same service type.
+     */
+    @JsonProperty(value = "instanceId")
+    protected String instanceId;
```
richturner: Could we use an incrementing counter for each serviceId?

model/src/main/java/org/openremote/model/microservices/Microservice.java
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * The version of the service.
+     * 
+     * Can be optionally set, to determine the version of the service, and interpret
+     * which
+     * version of the service is running and registered.
```
richturner:
```suggestion
     * which version of the service is running and registered.
```

model/src/main/java/org/openremote/model/microservices/MicroserviceResource.java
```diff
@@ -0,0 +1,190 @@
...
+import jakarta.ws.rs.*;
+
+import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
+
+/**
+ * REST resource for managing microservices and external services.
+ * 
+ * This resource provides endpoints for service discovery, registration, and
+ * management
+ * of external services and microservices through the OpenRemote manager.
```
richturner:
```suggestion
 * management of external services through the manager.
```

model/src/main/java/org/openremote/model/microservices/MicroserviceResource.java
```diff
@@ -0,0 +1,190 @@
...
+ * 
+ * Registered services are made available via the OpenRemote manager's Web UI
+ * and API,
+ * enabling centralized service management and monitoring.
```
richturner:
```suggestion
 * and API, enabling centralized service management and monitoring.
```

setup/src/integration/java/org/openremote/setup/integration/KeycloakTestSetup.java
```diff
@@ -133,6 +137,23 @@ public void onStart() throws Exception {
...
+        superServiceUser = new User()
+            .setServiceAccount(true)
+            .setEnabled(true)
+            .setUsername("superserviceuser");
+        superServiceUser = keycloakProvider.createUpdateUser(MASTER_REALM, superServiceUser, UniqueIdentifierGenerator.generateId("superserviceusertest"), true);
+        keycloakProvider.updateUserClientRoles(
+            MASTER_REALM,
+            superServiceUser.getId(),
+            Constants.KEYCLOAK_CLIENT_ID,
+            Stream.of(ClientRole.READ_ASSETS, ClientRole.WRITE_ASSETS, ClientRole.WRITE_ATTRIBUTES, ClientRole.WRITE_SERVICES, ClientRole.READ_SERVICES).map(ClientRole::getValue).toArray(String[]::new)
```
richturner: Does this user need asset and attribute roles?

ui/app/manager/src/headers.ts
```diff
@@ -37,6 +46,8 @@ export function headerItemInsights<S extends AppStateKeyed, A extends AnyAction>
     };
 }
 
+
+
```
richturner:
```suggestion
```

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: I've set it to the interval

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: I've set it to the interval

ui/component/or-tree-menu/src/index.ts
```diff
@@ -73,7 +73,7 @@ const styles = css`
         display: flex;
         justify-content: space-between;
         align-items: center;
-        padding: 0 15px;
+        padding: 0px 5px 0px 15px;
```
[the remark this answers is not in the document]
dominiquekleeven: This fixes inconsistent padding in the header of the tree menus across pages. This is because some pages use a different tree header structure.

ui/.prettierrc.json
```diff
@@ -1,5 +1,7 @@
 {
   "trailingComma": "es5",
   "printWidth": 120,
-  "singleQuote": false
+  "singleQuote": false,
+  "tabWidth": 4,
+  "useTabs": false
```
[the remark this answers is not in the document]
dominiquekleeven: Should be fine, we don't apply formatting yet and 4 spaces is what we use for indentation.

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: Fair, I've removed the overload method and set the flag directly in the resource method

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: Yep totally missed that

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
dominiquekleeven: Removed the redundant null check

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
dominiquekleeven: Removed the redundant null check

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: Moved before logging

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: The idea here is that a global service is available to all realms, its basically a service that supports multi-tenancy, such as the ML Forecast Service. So events for these services should be published so that the front-end can react to them e.g. add a new service to the list, update the state etc.

Realm specific services are only available for the realm they are registered in and are supposed to be single-tenant instances.

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: Made it a direct get

setup/src/integration/java/org/openremote/setup/integration/KeycloakTestSetup.java
dominiquekleeven: Should've been only the services roles, changed it

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
dominiquekleeven: Now uses `getRequestRealmName()` and sets it on the external service

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
dominiquekleeven: Now uses `getRequestRealmName()` and sets it on the external service

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
dominiquekleeven: Its being tracked as part of the leaseInfo now, as `registrarUserId` which allows to keep track of it in the back-end. I added checks for both deregister and heartbeat.

model/src/main/java/org/openremote/model/microservices/Microservice.java
dominiquekleeven: `serviceId` shows up in the front-end and is the parent key for all instances (based on Eureka), so I wouldn’t tie it to a username. Since it would get exposed to the front-end. 

I did add `registrarUserId` to the leaseInfo though(internal only prop), so we can still track who registered what, and also add additional security checks such making sure a service instance can't be de-registered by someone who didn't register it in the first place.

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: Now uses the isExpired helper

model/src/main/java/org/openremote/model/microservices/Microservice.java
dominiquekleeven: Changed instanceId to be an integer and now uses an incremental counter. Each serviceId has its own counter and gets incremented when a new instance is registered.

Gets cleared on service stop, and the counter for a serviceId gets cleared/reset when all instances of the service have been de-registered.

manager/src/main/java/org/openremote/manager/microservices/MicroserviceResourceImpl.java
dominiquekleeven: Also added test cases for these

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: You’re right, the nested concurrent map was unnecessary for synchronization. I’ve changed it to a regular map and made sure that any lookups using the instanceId are done on the key.

manager/src/main/java/org/openremote/manager/microservices/MicroserviceRegistryService.java
dominiquekleeven: Not sure what you meant here,

Do you mean renaming the isExpired() method? I think isExpired() is more appropriate since it’s directly related to the expiry of the lease/registration.

dominiquekleeven:
> Sorry for the delay in getting to this, looking good just some minor feedback.
> 
> Also maybe ExternalService rather than Microservice would be more appropriate, we then have ContainerService for within JVM


@richturner I've addressed all review comments, besides the Flyway migration one. I've also renamed things from `Microservice`  to `ExternalService`

As mentioned this morning, I'd appreciate a re-review, in the mean time I'll continue with the migration and docs.

dominiquekleeven:
Added the flyway migration for the new roles (placed in the `org/openremote/manager/setup/database` directory since Flyway was already configured to look there).

I tested the migration manually by removing the client roles via the Keycloak administration, and then running the migration to confirm that the roles were re-added. I verified this for both single-realm and multi-realm setups. In both cases, the role was correctly added to the realm:  `openremote` client roles.

dominiquekleeven:
> Do we need to carefully undo each code migration step if we get a failure along the way, given we don't have a transaction to provide that for us?

So from my testing the following happens:

If a migration fails (due to any unhandled exception), it prevents the manager from starting and will run the migration again on the next startup/restart. This means a bad migration can block the manager from starting. Any changes made during this migration are persisted, so for example, some realms may have the roles added while others may not.

We could add logic to try to undo any changes, but depending on the exception or issue encountered, that might be impossible.

Unsure how to exactly handle failures in this case.
***

How do we usually handle this when new roles have been added? e.g. `write:alarms` / `read:alarms` I can't seem to find a migration related to it?

PR #1586, opened 2025-01-13
Storybook UI documentation

MartinaeyNL:
Changelog:
- Added the new Storybook app as `/ui/app/workshop`, meant for local and staging deployments.
- Created a new `docs` component containing generated documentation like the Custom Elements manifest.
- ...

**This is pull request is very WIP.** 

(but you can ignore the `temp` in the branch name) 👍

PR #1984, opened 2025-07-07
UI fix for attribute picker not working in Flow Rules.

### LINKED ISSUE #1828: Rules: Flow: Asset attribute selector broken
**Describe the bug**
When changing the asset attribute of an input or output, the asset select list is broken.

**To Reproduce**
1. In the Flow rules add an input node and select an asset attribute
2. Click the node again to change its attribute
3. Try to select assets on the left (you need multiple)
4. Try to select the original asset
5. Click around a bit and see how wonky it is.

**Expected behavior**
3. You can select any asset in the list to see its available attributes

**Screenshots**
https://github.com/user-attachments/assets/82f2cb07-01ab-4dc5-8126-0a33c9f18429

OR 1.4.0

Has not been recently introduced.

MartinaeyNL:
## Description
Previously, you could not modify the selected asset/attribute within the Flow Rules UI.
A workaround was to delete the node, and recreate it with a different asset.
This PR resolves this issue.

Fixes #1828 

I simply removed the references of selected assets / attributes.
This might sound like a weird solution, by just removing it, but it makes the most sense to me.
Other pages like the 'Data Export', and widgets on the Insights, do not pre-select any asset/attributes upon opening the menu.
So, for me the easiest solution was to remove them.

Any feedback is appreciated! 👍 

## Checklist

## Screenshot

image

### DIFF
--- a/ui/component/or-rules/src/flow-viewer/components/internal-picker.ts
+++ b/ui/component/or-rules/src/flow-viewer/components/internal-picker.ts
@@ -1,6 +1,6 @@
 import { LitElement, html, css, TemplateResult } from "lit";
 import {customElement, property} from "lit/decorators.js";
-import { Node, PickerType, AttributeInternalValue, Asset, NodeDataType, AttributeRef, AssetModelUtil } from "@openremote/model";
+import { Node, PickerType, AttributeInternalValue, Asset, NodeDataType, AssetModelUtil } from "@openremote/model";
 import { nodeConverter } from "../converters/node-converter";
 import { InputType, OrInputChangedEvent } from "@openremote/or-mwc-components/or-mwc-input";
 import rest from "@openremote/rest";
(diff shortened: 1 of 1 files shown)


MartinaeyNL:
Will talk with product owners internally to review this functionality.

MartinaeyNL:
Thanks!
To add what I have found during the investigation process;

There is a rendering / lifecycle issue where the selected asset is updated, but is reverted shortly after.
So an update with the **new** asset takes place, whereafter a new lifecycle takes place with the **old** value.
I'm not sure where this is from, but it seems to be `or-asset-tree` specific,
or one of the consumers is injecting data through HTML attributes.

I would assume that other integrations of `or-asset-tree`, like the Insights and Data Export should experience the same.
There might be some 'higher-up' influence, so that's why we should test/reproduce it at those different pages first.
Isolated testing using Storybook (#1586) would be ideal for situations like this, to easily identify the cause of this issue.

For now, I simply disabled this "preselect asset / attribute" functionality.
You can repost this text above when a new issue has been created.

PR #1996, opened 2025-07-11
UI updates: Added Bar Chart widget

MartinaeyNL:
## Description
See original PR by @Hackerberg43 in #1935.
This PR is a continuation of his work to get the bar chart added to the Manager UI.

## Changelog
- Added bar chart widget
- *work in progress*

## Checklist

### Design notes
- Created a separate function for interval data, where every set of bars represents an interval. This is a highly customized interpretation and not a classic echarts bar-chart. It uses a time axis instead of category to clearly indicate the start and end times of the interval.
- The background is generated based on the interval data, to even more clearly show which timeframe the data refers to. When no data is present, the background is also hidden, this means gaps in the data do actually make sense, which is definitely not the case otherwise.
- Bar widths are defined by a custom calculation, to even better fit them to the relevant timeframe
- Took the reworked time controls from new or-chart as a basis
- Added the possibility to select an interval, which also has an auto functionality and the user input is checked on validity. You cant select Interval: 1 Month with timeframe This Day for example.
- Added more aggregation methods in AssetDataPointIntervalQuery and made them available in this widget. 

| Method | Number                                       | Boolean                            |
|--------|----------------------------------------------|------------------------------------|
| COUNT  | Amount of datapoints                         | Amount of datapoints               |
| SUM    | Sum of all values         | Amount of Trues                   |
| MODE   | Most frequent number                         | Predominant state                  |
| MEDIAN | Middle number                                |Kind of same result as mode  |

- You can select more than one aggregation method per attribute

### Screenshots

afbeelding
afbeelding
afbeelding


MartinaeyNL:
I will continue on this after PR #1997 has been merged, as some functionality can be reused for this PR as well.

ui/component/or-dashboard-builder/src/widgets/barchart-widget.ts
```diff
@@ -0,0 +1,222 @@
...
+                <div style="height: 100%; display: flex; justify-content: center; align-items: center; text-align: center;">
+                    <span><or-translate .value="${this._error}"></or-translate></span>
+                </div>
+                
+            `, () => {
+                return html`
+                    <or-attribute-barchart .assets="${this.loadedAssets}" .assetAttributes="${this.assetAttributes}"
+                                           .attributeColors="${this.widgetConfig?.attributeColors}" .attributeConfig="${this.widgetConfig?.attributeSettings}"
+                                           ?showLegend=${this.widgetConfig?.showLegend} ?stacked=${this.widgetConfig?.stacked}
+                                           .timestampControls="${!this.widgetConfig?.showTimestampControls}"
```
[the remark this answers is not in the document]
MartinaeyNL: It's a bit unfortunate naming, but it is consistent with the chart widget.
So I'm leaving this for now to prevent breaking changes or inconsistency across widgets.

MartinaeyNL:
Updated the changelog with the latest additions 👍 
Ready to review!

I'll assign @Ekhorn to review this.

ui/component/or-attribute-barchart/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "main": "dist/umd/index.bundle.js",
+  "module": "lib/index.js",
+  "exports": {
+    ".": "./lib/index.js",
+    "./*": "./lib/*.js"
+  },
+  "types": "lib/index.d.ts",
+  "scripts": {
+    "test": "echo \"No tests\" && exit 0",
+    "prepack": "npx webpack"
```
Ekhorn:
```suggestion
    "prepack": "npx rspack"
```

ui/component/or-dashboard-builder/src/widgets/barchart-widget.ts
```diff
@@ -0,0 +1,222 @@
...
+    // We either refresh the datapointQuery or the full widgetConfig depending on the force parameter.
+    // TODO: Improve this to a more efficient approach, instead of duplicating the object
+    public refreshContent(force: boolean) {
+        if(!force) {
+            const datapointQuery = JSON.parse(JSON.stringify(this.widgetConfig.datapointQuery)) as AssetDatapointQueryUnion;
+            datapointQuery.fromTimestamp = undefined;
+            datapointQuery.toTimestamp = undefined;
+            this.datapointQuery = datapointQuery;
+        } else {
+            this.widgetConfig = JSON.parse(JSON.stringify(this.widgetConfig)) as BarChartWidgetConfig;
```
Ekhorn:
```suggestion
            const datapointQuery = structuredClone(this.widgetConfig.datapointQuery) as AssetDatapointQueryUnion;
            datapointQuery.fromTimestamp = undefined;
            datapointQuery.toTimestamp = undefined;
            this.datapointQuery = datapointQuery;
        } else {
            this.widgetConfig = structuredClone(this.widgetConfig) as BarChartWidgetConfig;
```

see https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone

The same can be done for the `chart-widget`.

ui/component/or-dashboard-builder/src/settings/barchart-settings.ts
```diff
@@ -0,0 +1,568 @@
...
+            }
+        } else {
+            if (type === "min") {
+                this.widgetConfig.chartOptions.options.scales.y1.min = value;
+            } else {
+                this.widgetConfig.chartOptions.options.scales.y1.max = value;
+            }
+        }
+        this.notifyConfigUpdate();
+    }
```
Ekhorn:
```suggestion
    protected setAxisMinMaxValue(axis: "left" | "right", type: "min" | "max", value?: number) {
        if (axis === "left") {
            this.widgetConfig.chartOptions.options.scales.y[type] = value;
        } else {
            this.widgetConfig.chartOptions.options.scales.y1[type] = value;
        }
        this.notifyConfigUpdate();
    }
```

ui/component/or-dashboard-builder/src/settings/barchart-settings.ts
Ekhorn, line 1: Seems like most methods here are duplicated from `chart-settings.ts`. Would be nice to extend those through a `BaseChartSettings` component.

ui/component/or-attribute-barchart/src/index.ts
```diff
@@ -0,0 +1,1303 @@
...
+    millis: number;
+}
+
+/**
+ * List of interval options
+ */
+export enum BarChartInterval {
+    AUTO = "auto", NONE = "none", ONE_MINUTE = "1Minute", FIVE_MINUTES = "5Minutes", THIRTY_MINUTES = "30Minutes", ONE_HOUR = "hour",
+    ONE_DAY = "day", ONE_WEEK = "week", ONE_MONTH = "month", ONE_YEAR = "year"
+}
```
Ekhorn: Where is prettier when u need it! ;)

```suggestion
export enum BarChartInterval {
    AUTO = "auto",
    NONE = "none",
    ONE_MINUTE = "1Minute",
    FIVE_MINUTES = "5Minutes",
    THIRTY_MINUTES = "30Minutes",
    ONE_HOUR = "hour",
    ONE_DAY = "day",
    ONE_WEEK = "week",
    ONE_MONTH = "month",
    ONE_YEAR = "year"
}
```

ui/component/or-attribute-barchart/src/index.ts
```diff
@@ -0,0 +1,1303 @@
...
+            }
+            this._latestError = "errorOccurred";
+        }
+    }
+
+    protected async _loadAttributeData(asset: Asset, attribute: Attribute<any>, color: string, from: number, to: number, formula: AssetDatapointIntervalQueryFormula, faint = false, label?: string, options?: any, _unit?: any): Promise<BarChartData> {
+        const dataset = this._getDefaultDatasetOptions(label ?? "", formula, color, faint);
+
+        if (asset.id && attribute.name && this.datapointQuery) {
+            const datapointQuery = JSON.parse(JSON.stringify(this.datapointQuery)); // recreating object, since the changes shouldn't apply to parent components; only or-attribute-barchart itself.
```
Ekhorn:
```suggestion
            const datapointQuery = structuredClone(this.datapointQuery); // recreating object, since the changes shouldn't apply to parent components; only or-attribute-barchart itself.
```

see https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone

test/src/test/groovy/org/openremote/test/assets/AssetDatapointQueryTest.groovy
```diff
@@ -313,6 +314,136 @@ class AssetDatapointQueryTest extends Specification implements ManagerContainerT
...
+        assetDatapointService.upsertValue(asset.getId(), attributeName, 90d, dateTime.minus(1, ChronoUnit.MINUTES))
+
+        and: "requesting interval datapoints using MODE with a full interval"
+        def modeDatapoints2 = assetDatapointService.queryDatapoints(
+                asset.getId(),
+                asset.getAttribute(attributeName).orElseThrow({ new RuntimeException("Missing attribute") }),
+                new AssetDatapointIntervalQuery(queryTime30, queryTimeNow, "60 minutes", AssetDatapointIntervalQuery.Formula.MODE, true)
+        )
+
+        then: "MODE datapoints will should return 90, because it is most common within the interval"
```
Ekhorn:
```suggestion
        then: "MODE datapoints should return 90, because it is most common within the interval"
```

test/src/test/groovy/org/openremote/test/assets/AssetDatapointQueryTest.groovy
```diff
@@ -313,6 +314,136 @@ class AssetDatapointQueryTest extends Specification implements ManagerContainerT
...
+
+        when: "requesting interval datapoints using MODE"
+        def modeDatapoints1 = assetDatapointService.queryDatapoints(
+                asset.getId(),
+                asset.getAttribute(attributeName).orElseThrow({ new RuntimeException("Missing attribute") }),
+                new AssetDatapointIntervalQuery(queryTime30, queryTimeNow, "5 minutes", AssetDatapointIntervalQuery.Formula.MODE, true)
+        )
+
+        then: "MODE datapoints should be correct"
+        assert modeDatapoints1.size() == 6 // TODO: Fix!! Because 30/5=6, plus an extra datapoint for the spare milliseconds of time.
```
Ekhorn: I'm guessing you're referring to the fact that MODE with no repeating numbers should result in "no mode" instead of the direct datapoint value. It would also make sense to test the happy flow here then (which might be what u are referring to).

Hackerberg43:
Hey Martin, I see you did quite some work there. Here are my usage related findings:

- Need to find a consistent naming of difference/delta method, now both are used. I think delta might be a bit too technical for the average person.

- I see you removed the "one" interval and replaced it by timeframe=interval eg. week per week. However, in most cases this displays two intervals instead of one. For example last week per week, shows last week and the week before that:
afbeelding


- I think the tooltip could use a | between attributeref and method:
afbeelding


- I see that you changed the position of the method label to the first positive number. However, when more than one method is selected, it is still confusing:
afbeelding

Would at least suggest to hide the label when only one method is active:
line 1256:
`show: !this.stacked && Object.entries(this.attributeConfig || {})
                    .filter(([key]) => key.includes("method"))
                    .length > 1,`
I made the background to help with the visibility to see which bar belongs to which interval. In this configuration selecting multiple methods is pretty useless and should be disabled imho. 

 - The xAxis labels have become quite inconsistant at times:
afbeelding

ui/component/or-dashboard-builder/src/settings/barchart-settings.ts
MartinaeyNL: Yes that is correct. And you're making a fair point here.
However, unfortunately I don't think it's generic enough to create a `BaseChartSettings` class.
As in, each chart will always have their own implementation, with slightly different options.
So, you could technically make a base class, but every chart settings would mutate the function anyway.

test/src/test/groovy/org/openremote/test/assets/AssetDatapointQueryTest.groovy
MartinaeyNL: Yeah I didn't spend enough time to investigate this, so I haven't found it out yet 😂 
Still can't get a grip on where it is going wrong.
Because it shouldn't differ per method/formula how many datapoints would return.
Requesting 6 hours of data, with an interval of 1 hour, should always return 7 datapoints.

I'll investigate it when I find the time 😅

Hackerberg43: the calculation method in this mode query results in null rows being deleted, so that behavior is as expected :)

MartinaeyNL: Thanks! I've removed the comment in code 👍

MartinaeyNL:
Thanks for taking a look @Hackerberg43 !

I adjusted several things, whereof many are the issues/bugs you addressed! 👍 
The x-axis labels are now shown in a 45 degree angle when there is too many data.
I've also refactored DELTA to DIFFERENCE, and improved the 'AVG/MIN/MAX' label positioning.

I also addressed some of @pierrekil his requests.

I'll deploy it to a test server of ours, and let @Ekhorn take a look again 😉 
Thanks everyone!

MartinaeyNL:
Addressed the feedback @pierrekil gave me in person, and deployed these changes on the same test server again 👍 
I'll try it out myself as well, to see if I can break it somehow 😂

MartinaeyNL:
The recent commit of "Fix for selecting more than 10 attributes not working", also fixes #2035 👍

PR #2002, opened 2025-07-14
Fix lint warnings on Dockerfile

ebariaux:
I had the lint warnings when building with buildx

### DIFF
--- a/kubernetes/README-AWS.md
+++ b/kubernetes/README-AWS.md
@@ -1,0 +1,187 @@
+# Kubernetes deployment under AWS (EKS)
+
+## TL;DR
+
+Requirements: you need to have kubectl, helm, [aws cli](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) and [eksctl](https://docs.aws.amazon.com/eks/latest/userguide/install-kubectl.html#eksctl-install-update) installed beforehand.
+
+The scripts have been tested within the openremote account.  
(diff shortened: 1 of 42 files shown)


manager/Dockerfile
```diff
@@ -79,4 +172,4 @@ EXPOSE 8080
 EXPOSE 1883
 
 HEALTHCHECK --interval=5s --timeout=60s --start-period=5s --retries=120 CMD curl --fail --silent http://localhost:8080 || exit 1
-ENTRYPOINT java $OR_JAVA_OPTS -cp /opt/app/lib/*:/deployment/manager/extensions/* org.openremote.manager.Main
+ENTRYPOINT ["java", "-cp", "/opt/app/lib/*:/deployment/manager/extensions/*", "org.openremote.manager.Main"]
```
wborn: Why did `$OR_JAVA_OPTS` get removed here?

ebariaux: Because I blindly copied stuff from an AI without reading the full explanation!
I'll push a fix for that.

wborn: I saw some of the default password environment variables were also removed. Was that intended for security reasons?

Some tools allow for showing the default environment values which makes it easier for users to discover what values to override without having to read documentation.

ebariaux: Yes, it is for security reasons, having passwords baked into docker images is not a good practice.
Understandably those are "dummy defaults" OK for testing so we could:
- live with the warnings
- add specific ignore instructions for the warnings
Both options are not ideal

Default values are anyway provided by docker compose files

wborn (approved):
Thanks for the fixes!

PR #2020, opened 2025-07-24
Add new datapoint export options, including new CSV formats

### LINKED ISSUE #1396: Data export: improve performance and feedback of data export page
When selecting asset attributes, it takes a long time before they appear in the table. Is that because the start and end time are looked up? If so, should we remove those columns?

It is unclear for the user if the data export timed out, or they have to wait longer. We should provide feedback.

Can we already know how much datapoints are about to be downloaded? If its too much, we can already inform the user, or extend the timeout?

We can indicate they reached a limit, and only enable the button if the amount of datapoints is reduced?

### LINKED ISSUE #1574: Enhancement/alternative data export format ( postgres update req. )
Backlog issue for the alternative .csv data export format proposed in https://github.com/openremote/openremote/pull/1557#issuecomment-2576266895

To be revisited when a postgres image update is required.

Hackerberg43:
### Data export page:

- Added loading indicators that are activated while retrieving table data and while the export request is running.
Before, it was not visible whether the app was loading, becoming unresponsive or simply not doing anything. (closes https://github.com/openremote/openremote/issues/1396)
- Added a list input to select alternative export formats.

afbeelding

### asset/datapoint/export resource :
 - Added an extra csv export format that was previously suggested in https://github.com/openremote/openremote/pull/1557 (closes https://github.com/openremote/openremote/issues/1574)
  397464018-f3a13eee-4f2c-4d05-bbbb-738bb93ca2c6

 - Added 1 minute average aggregation option, potentially saves even more time post-processing the csv.

-  Fix resource leak using try-with-resources on FileInputStream (a leak was observed while we tried multiple times extracting data that resulted in database timeouts, not sure if this is the direct cause though)
 - Added migration script to add tablefunc extension to postgres, required for the new csv formatting.
## Checklist
- [x] 1. I've made "format" an optional query parameter in asset/datapoint/export to avoid an API breaking change. However, I'm not fully confident if this is the correct way. I did not find any usage of @DefaultValue in any of the existing api resources, while it seems a straightforward solution. I'm trying to create something similar to the optional parameters used in the datapoint query (gapfill, formula,..), but not sure if that is the best way either, because those did not get picked up by the automated API documentation.

- [x] 2. Tested on a local deployment only with a 2 million datapoint database.
- [x] 3. The format names are little lengthy, not sure how to improve.


Hackerberg43:
https://github.com/openremote/vdl-es/issues/161

Hackerberg43:
Intial comments/insights have been addressed, this should be ready for proper review.

manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointResourceImpl.java
```diff
@@ -174,7 +174,7 @@ public DatapointPeriod getDatapointPeriod(RequestParams requestParams, String as
     }
 
     @Override
-    public void getDatapointExport(AsyncResponse asyncResponse, String attributeRefsString, long fromTimestamp, long toTimestamp) {
+    public void getDatapointExport(AsyncResponse asyncResponse, String attributeRefsString, long fromTimestamp, long toTimestamp, int format) {
```
richturner: Switch to an enum for export format

richturner (changes requested):

richturner:
As discussed @Hackerberg43 a frontend test would be great to get this finished @MartinaeyNL can you review the frontend code changes please.

ui/app/manager/src/pages/page-export.ts
```diff
@@ -200,6 +204,14 @@ export class PageExport extends Page<AppStateKeyed> {
     @property({type: Number})
     private latestTimestamp: number = moment().valueOf();
 
+    @property()
+    protected _loading: boolean = false;
+
+    protected _exportFormats: {[key: number]: string} = {1: 'CSVDefault', 2: "CSVCrosstab", 3: "CSVCrosstab | 1min interval"};
```
MartinaeyNL: I'd try to use an enum for this.

As Rich proposed that for the backend as well, I would simply re-use that one here,
and update both the `_exportFormats` variable and `selectedFormat` property.

manager/src/main/resources/org/openremote/manager/setup/database/V20251201_01__addTableFunc.sql
MartinaeyNL, line 1: Update the date within the file name here

ebariaux:
@Hackerberg43 Any status update on this one ?

Hackerberg43:
Hey @ebariaux yes I do.
Review comments are processed and a frontend test is made, but I wanted to discuss the relevance here.
The core functionality is covered by the backend tests. The only thing I could think of for a frontend test is:
-whether the selected attribute is loaded into the table after selection.
-whether the download happens

MartinaeyNL:
Yeah we should've been more clear about what kind of frontend test we're talking about. 😅 
Sorry about that @Hackerberg43 

We tend to use frontend tests for end-to-end purposes, so it would also include an HTTP call and expected response.
So I think it makes sense to add a "general data export" test, as there isn't any at the moment.
So it'd be: Add asset attributes -> Select time start -> Select time end -> Select export format -> Press export -> Expect some form of UI response like a snackbar.

Nothing too extensive, but just a test to guarantee the export functionality keeps working as expected over time. 👍

Hackerberg43:
Added. The test fails on my PC because the backend returns a 500 on export because of file permissions in /tmp, I see that the the CI/CD test also fails on the export api call. But I can't see whats going on in the backend there.

Hackerberg43:
Ok so it really seems that the test environment has the same issue as the local: receiving a 500 from the backend:

```
 Usage:
        yarn playwright show-trace test-results/export-Export-datapoints-successfully-triggers-download-chromium-retry1/trace.zip
    ────────────────────────────────────────────────────────────────────────────────────────────────
    Retry #2 ───────────────────────────────────────────────────────────────────────────────────────
    Error: expect(received).toBe(expected) // Object.is equality
    Expected: 200
    Received: 500
      108 |
      109 |   const response = await exportResponsePromise;
    > 110 |   expect(response.status()).toBe(200);
```

Locally the manager logs:
```
2025-12-03 16:12:34.279  FINE    [WebService task-3             ] emote.container.web.WebServiceExceptions : Web service exception in 'RESTEasy Dispatch' for 'GET http://localhost:8080/api/asset/datapoint/export?attributeRefs=%5B%7B%22id%22%3A%2260AZOH0YyIPD4xXQdkRhpB%22%2C%22name%22%3A%22energyLevel%22%7D%5D&fromTimestamp=1764771120000&toTimestamp=1764771120000&format=CSV'⁠ type = WebApplicationException, message=HTTP 500 Internal Server Error

jakarta.ws.rs.WebApplicationException: HTTP 500 Internal Server Error at org.openremote.manager.datapoint.AssetDatapointResourceImpl.getDatapointExport(AssetDatapointResourceImpl.java:235)

2025-12-03 16:12:34.285  SEVERE  [WebService task-3             ] atapoint.AssetDatapointResourceImpl.DATA : Exception in ScheduledFuture: 

java.io.FileNotFoundException: tmp/datapoint/6sjp4jD8sGIdOd7eQ22mFw.csv (No such file or directory)
```

When fully building the stack and testing it manually it works just fine, so I guess something needs to be tweaked in the test environment regarding postgres usage of /tmp folder in dev-ui.yml to make this work.

richturner:
Looks like an issue with the `profiles/dev-ui.yml` not having the correct path mapping for the tmp folder and therefore tmp/datapoint dir doesn't exist, simple change needed in `dev-ui.yml` to align with `dev-testing.yml`
@Ekhorn still worth doing the work on using std out for exports

Hackerberg43:
Thanks for the improvements @Ekhorn !  I will continue to do some tests because found some new cases where the crosstab export does not work properly, possibly also causing the failure on your exports

manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointService.java
```diff
@@ -54,13 +57,10 @@ public class AssetDatapointService extends AbstractDatapointService<AssetDatapoi
 
     public static final String OR_DATA_POINTS_MAX_AGE_DAYS = "OR_DATA_POINTS_MAX_AGE_DAYS";
     public static final int OR_DATA_POINTS_MAX_AGE_DAYS_DEFAULT = 31;
-    public static final String OR_DATA_POINTS_EXPORT_LIMIT = "OR_DATA_POINTS_EXPORT_LIMIT";
-    public static final int OR_DATA_POINTS_EXPORT_LIMIT_DEFAULT = 1000000;
```
Ekhorn, line 62: @richturner mentioned the data export limit is needed to prevent the database potentially grinding to a halt.

So this [commit](https://github.com/Hackerberg43/openremote/pull/9/changes/57f349c43a2409a5f81164a1cb7f6f4f61699912) should be reverted (except for the "Purge old exports" and "Ensure export dir exists and is writable" code).

Ekhorn:
@Hackerberg43 let me know when you think it's ready for review. I can go do the same testing I did in https://github.com/Hackerberg43/openremote/pull/9.

Hackerberg43:
Hi Koen @Ekhorn , I have this commit running on our testrig right now with 200m datapoints. It is unable to export data, only gives timeouts. I'm currently narrowing down if this is caused by the DB bloat and not the export itself. Will keep you updated.

Ekhorn:
> I have this commit running on our testrig right now with 200m datapoints. It is unable to export data, only gives timeouts. I'm currently narrowing down if this is caused by the DB bloat and not the export itself. Will keep you updated.

@Hackerberg43 Just checking that you're aware of the fact that `@openremote/core` sets the request timeout to 20 seconds (on the frontend). Increasing [`rest.setTimeout(20000);`](https://github.com/openremote/openremote/blob/9695fd492201ca928fbe4e41e215b8fa701f2c0c/ui/component/core/src/index.ts#L588) avoids these timeouts and it might make sense to increase this timeout limit (for datapoint exports), because downloading hundreds of megabytes or gigabytes of data at this scale starts to get throttled by connection speeds.

Hackerberg43:
Ah good to know! I'm trying some of the queries directly into the testrig database right now, which take a couple of minutes to complete. I am currently trying to optimize/clean the database a bit to see if it can be brought down to a reasonable time. Otherwise it might need to be considered to move back to disk storage and decouple the exporting/downloading process.

Hackerberg43:
So after some DB tuning our testrig is just able to export 2m datapoints before the timeout. While still not sufficient for our use case, I think for most projects this is completely fine? We might diverge to a separate solution in our custom project that performs export jobs in the background and exposes files in an export folder as downloadable.

So this is ready to test @Ekhorn.

Ekhorn:
There seems to be an issue with the GitHub actions runner timing out on certain PRs. I'll check if I can get this resolved on my own first.

manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointService.java
```diff
@@ -57,10 +61,10 @@ public class AssetDatapointService extends AbstractDatapointService<AssetDatapoi
     public static final String OR_DATA_POINTS_EXPORT_LIMIT = "OR_DATA_POINTS_EXPORT_LIMIT";
     public static final int OR_DATA_POINTS_EXPORT_LIMIT_DEFAULT = 1000000;
     private static final Logger LOG = Logger.getLogger(AssetDatapointService.class.getName());
+    private static final Logger DATA_EXPORT_LOG = SyslogCategory.getLogger(DATA, AssetDatapointResourceImpl.class);
     protected static final String EXPORT_STORAGE_DIR_NAME = "datapoint";
```
Ekhorn: The `EXPORT_STORAGE_DIR_NAME` property is now unused and can be removed.

```suggestion
```

Ekhorn (commented):
Nice work @Hackerberg43!

I will approve this when the threads are resolved.

See https://github.com/openremote/openremote/pull/2020#discussion_r2518314966 and https://github.com/openremote/openremote/pull/2020#discussion_r2740445753

Hackerberg43:
@Ekhorn  Done.
The sql injection fix required some code updating but done now. db migration file updated for today.

Ekhorn (approved):
With these changes the normal CSV export runs a little faster again (after the recent SQL injection defensive measure) and prevents filesystem issues, also the crosstab formats are a nice addition.

Thanks @Hackerberg43!

richturner (approved):

PR #2022, opened 2025-07-24
UI fix: Tables appearing too wide on the screen

### LINKED ISSUE #1873: Gateway Tunnels: TCP connection does not have stop button
## User Story
As a user with an open TCP gateway tunnel, I want to be able to close the tunnel using the UI, so that I don't have to do it programmatically.

MartinaeyNL:
## Description
Fixes #1873 

This PR resolves an issue where buttons and text were "disappearing off the screen".
It was because of a hardcoded CSS `width` value, that Material wasn't happy with overflowing the data.
I've removed the hardcode, since there is no use for it outside of the Alarms table, that has an override for this.

## Changelog
- Removed hardcoded CSS `width` property, causing buttons to be hidden in the Material UI handling.
The function will stay to exist, to allow method overriding by classes like `or-alarms-table`.

## Checklist


MartinaeyNL:
I've tested the changes with all tables in the Manager UI, and they function as expected! 👍 
The PR is ready to review!

PR #2026, opened 2025-07-28
UI fixes for production builds after RSPack migration

MartinaeyNL:
## Description
This fix solves an issue with the Redux router callback not binding to `this`.
It was an issue that came to light after the RSPack migration in #1857, where production deployments were not functioning.
These changes will fix at least one of these issues.

I came across several other issues during the process, although I could not reproduce them consistently.
So, as I'm lost with what other "production bundling issues" are left, we'll just merge this first, and see what comes next.

## Changelog
- Adds `lib` directory deletion to clean tasks across all UI component build scripts, to improve consistent local builds.
- Fixes arrow function syntax in router hook configuration for binding `this` correctly in the function.

### DIFF
--- a/balena/manager.Dockerfile
+++ b/balena/manager.Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/openjdk-21-runtime
+FROM registry.access.redhat.com/ubi9/openjdk-21-runtime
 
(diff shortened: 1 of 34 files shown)


MartinaeyNL:
Added a description and changelog to the top comment. 👍

wborn (approved):
LGTM! :+1:

PR #2030, opened 2025-08-07
Fix inconsistent parallel rspack builds

Ekhorn:
### Summary

The Rspack builds are inconsistent resulting in varying sizes of the JavaScript bundle missing critical code for the manager UI to function. This is can happen when running production builds in parallel. If multiple processes need to transpile the TypeScript code this will result in issues. I've found through my testing that for some unknown reason the `manager` app and the `console_loader` app if built at the same time will result broken JavaScript bundles around 30-40% of the time (when running it through `gradle clean installDist`). Running those builds in a loop separately at the same increases the rate of incorrect builds and can even cause one of the processes to exit.

Why did this issue not occur before Rspack? It could be that the Webpack builds were too slow for this to occur.

### Results

I tested this issue on the [frontend testing branch](https://github.com/openremote/openremote/pull/1890) and master branch.

Using variations of the following script I checked the bundle JS bundle size after each build.

```sh
set -e

while :
do
    out="$(gradle clean installDist 2>&1)"
    size=$(du -s ui/app/manager/dist/bundle.*.js | cut -f1)
    echo "$size"
done
```

Letting it run in the following scenarios led to the following results:
image

I also tested this on the master branch:

image

As you can see the combination of building multiple apps in parallel is causing the bundle corruption issue. Disabling parallel task execution would be a possible solution, though not ideal.

I then tried to see if any particular app was conflicting with the manager:

image

The results show that the `console_loader` app is causing the bundle corruption issue. Although I never found out why it is only the `console_loader` app causing this issue, it is clear that this must have something to do with TypeScript transpilation.

I inspected this with Rsdoctor and found errors like these:

image

I eventually found the [`transpileOnly`](https://github.com/TypeStrong/ts-loader?tab=readme-ov-file#transpileonly) option in ts-loader which effectively disables type checking and this resulted in consistent builds again (while running them in parallel.)

Though we don't want to remove type-checking, so to reintroduce this we use the [`TsCheckerRspackPlugin`](https://github.com/rspack-contrib/ts-checker-rspack-plugin) to run the type checking a separate process. This plugin uses TypeScript's module resolution (instead of Rspack). This may be slower (see https://rspack.rs/guide/tech/typescript#type-checking) so I've also enabled [incremental mode](https://github.com/rspack-contrib/ts-checker-rspack-plugin?tab=readme-ov-file#enabling-incremental-mode).

## Checklist

### DIFF
--- a/ui/tsconfig.json
+++ b/ui/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "compileOnSave": false,
   "compilerOptions": {
+    "incremental": true,
     "target": "es6",
     "module": "esNext",
     "moduleResolution": "node",
--- a/ui/util/package.json
+++ b/ui/util/package.json
@@ -27,6 +27,7 @@
     "querystring-es3": "^0.2.1",
     "shx": "~0.4.0",
     "source-map-loader": "^5.0.0",
+    "ts-checker-rspack-plugin": "^1.1.5",
     "ts-loader": "~9.5.2",
     "typescript": "~5.8.3",
     "webpack": "^5.99.8"
--- a/ui/util/webpack.util.js
+++ b/ui/util/webpack.util.js
@@ -2,6 +2,7 @@ const fs = require("fs");
 const path = require("path");
 const { rspack } = require('@rspack/core');
 const HtmlWebpackPlugin = require("html-webpack-plugin");
+const { TsCheckerRspackPlugin } = require('ts-checker-rspack-plugin');
 
 function getStandardModuleRules() {
     return {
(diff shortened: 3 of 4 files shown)


Ekhorn:
I briefly tested build performance and it is essentially the same speed. There might be a slight speed up in development mode when type checking.

Ekhorn:
One unfortunate issue is that ts-loader tries to use the Webpack API that doesn't exist in Rspack resulting in annoying errors like these, but it doesn't hinder development.

image

MartinaeyNL:
While looking through the list of pull requests, I saw this one hanging 😅 

It's been resolved since by removing the `console_loader` app from the codebase, since it had no serve. (PR #2031)
Although it's not fully solved yet, since our bundling process using RSPack is not optimal yet.
We're hoping to overhaul the bundling later using RSBuild, whereof you can track the progress here: #2032
I'm closing this PR for now.

PR #2031, opened 2025-08-07
Remove console_loader UI app to prevent broken JavaScript bundles

wborn:
Rspack production builds are unreliable when multiple apps are built in parallel, particularly the manager and console_loader apps. When built together (e.g., via `gradle clean installDist`), the resulting JavaScript bundles for the manager UI are broken ~30–40% of the time, often missing critical code. This seems to stem from TypeScript transpilation conflicts when processes run concurrently.

Testing shows that removing the console_loader app resolves these issues and prevents random build failures and unexpected process exits. Until Rspack's concurrency issues are addressed, we remove console_loader to ensure reliable builds.

The console_loader app is no longer being used.

See: openremote/openremote#2030
Caused by:  #1857


MartinaeyNL:
I think we should also remove all backend references (like in `ManagerWebService.java` as well, right?

richturner:
@wborn Please create a tag for the console_loader so we can easily get back to it if needed

wborn:
Hi @richturner, I just created the [archive/console_loader](https://github.com/openremote/openremote/releases/tag/archive%2Fconsole_loader) tag for this.

Issue #1990, opened 2025-07-10
Improve Asset Tree UI performance using server-side pagination


MartinaeyNL:
## User Story
As a user of the Manager UI, I want to quickly retrieve my large dataset of assets (50.000+),
so that I can properly load and use the `/assets` page without too many hiccups or loading indicators.

## Description
Currently, the asset tree within the Manager UI has scalability issues with instances over 10.000 assets.
It either takes a long time to load because of network payloads, or hangs the browser due to
the amount of HTML elements it is rendering.

This was part of an initial research, done and described in spike #1950.

## Acceptance Criteria
- [ ] 1. A user with access to 100.000 assets, with 20 attributes each on the root level,
can load the `/assets` page within a second.
- [ ] 2. A user with access to 100.000 assets, with 20 attributes over 3 levels (50 parent assets, 2000 children each),
can load the `/assets` page within a second.
- [ ] 3. In scenarios 1 and 2, the asset tree can be sorted by name, creation date, and asset type, within a second.
- [ ] 4. In scenarios 1 and 2, a user navigates to `/assets/<id>`, view its details, see their parents, all within a second.
- [ ] 5. Navigating to a different page, and coming back to the `/assets` page, there is no fetching of data needed.

### Additional wishes
- [ ] 6. In scenarios 1 and 2, the asset tree can be filtered by asset type, within two seconds.
- [ ] 7. In scenarios 1 and 2, the asset tree can be filtered by attribute, within two seconds.
- [ ] 8. In scenarios 1 and 2, the asset tree can be filtered by attribute value, within two seconds.

## Status

Awaiting assignment on the internal sprint board.

ebariaux:
@MartinaeyNL all sub-tasks are done, I think we can close this epic, or do you think there's still something that needs to be covered ?

MartinaeyNL:
@ebariaux I think it's worth testing the latest release with the acceptance criteria.
As in, I didn't test whether 100k assets result in the correct behavior for example.
If that is complete, we can close this epic.

PR #2067, opened 2025-08-19
UI: Added ability to filter by ID in the asset tree

### LINKED ISSUE #1994: Asset tree: add UI support for filtering by ID
## User Story
As a user of the Manager UI, I want to search for my asset by asset ID,
so I can quickly find the asset I'm looking for, and view the asset on its own without clutter.

## Description
As part of #1990, we're improving the performance of the Asset Tree UI.

After #1993 has been implemented, you cannot navigate to `/assets/<id>` for viewing the asset.
This is because pagination has been added, and only asset 1-1000 is shown in the UI.
So if asset number 2500 is requested, it cannot be shown.

To solve this, we'd like to add ID support to the filter input field.
So if a 20 character string is filled in, we automatically check if it's an ID that exists in the database.
When found, we show a limited amount of assets, similar to the current filter feature.

## Acceptance Criteria
- [ ] 1. The user has the ability to fill in a 22-character ID into the "filter" input field, showing assets with the same ID. 
- [ ] 2. If the 22-character ID is also the name of a different asset, it should also be shown in the UI.
- [ ] 3. When filtering by ID, similar to the existing filter, the parents of this asset will also be shown. (grayed out)
- [ ] 4. If user navigates to `/assets/<id>`, it should automatically apply this new ID filter.

These acceptance criteria can be adjusted based on feedback, or technical limitations during implementation.

## Implementation Notes
- My recommendation is to add the ID request alongside the existing one, so both displayname and ID are covered.

MartinaeyNL:
## Description
Fixes #1994.

This PR adds the ability to filter the asset tree UI by asset ID.
It utilizes the same behavior and input field as filtering by asset name, or by attribute value.
Navigating to `/asset/.../<my asset ID>` will now automatically apply the filter for you.

> Be reminded, although this is part of #1990, there is no change in asset querying logic / pagination here.
> It is still using local asset cache, and all assets are fetched beforehand.

## Checklist

### DIFF
--- a/ui/component/or-asset-tree/src/index.ts
+++ b/ui/component/or-asset-tree/src/index.ts
@@ -1167,9 +1167,10 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
     }
 
     protected async getMatcherFromQuery(): Promise<((asset: Asset) => boolean)> {
-        let assetCond: StringPredicate[] | undefined = undefined;
-        let attributeCond: LogicGroup<AttributePredicate> | undefined = undefined;
-        let assetTypeCond: string[] | undefined = undefined;
+        let assetCond: StringPredicate[] | undefined;
+        let attributeCond: LogicGroup<AttributePredicate> | undefined;
+        let assetTypeCond: string[] | undefined;
+        const assetQueries: AssetQuery[] = [];
 
         if (this._filter.asset) {
             assetCond = [{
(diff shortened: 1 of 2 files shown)


MartinaeyNL:
Pushed a few bugfixes for issues I came across during testing. Should be OK now 👍 
I'll start trying to add frontend tests to cover the acceptance criteria.

ui/component/or-asset-tree/src/index.ts
```diff
@@ -908,15 +937,12 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
     }
 
     protected parseFromInputFilter(inputValue?: string): OrAssetTreeFilter {
```
Ekhorn:
```suggestion
    protected parseFromInputFilter(inputValue = this._filterInput?.value): OrAssetTreeFilter {
```

GitHub is being dumb cannot select this and the next line (pls fix locally) :wink:

Ekhorn (changes requested):
Still need to check some remaining changes and test it locally.

MartinaeyNL:
Put this back to draft, to take a look at a testing setup for `<or-asset-tree>`

ui/app/manager/test/assets.test.ts
```diff
@@ -47,10 +47,50 @@ assets.forEach(({ type, name, attributes }) => {
...
+   * @then The asset detail page is displayed
+   */
+  test(`Search by Asset ID and select the ${name} asset`, async ({ page, manager }) => {
+    await manager.setup("smartcity", { assets });
+    const id = manager.assets.find(asset => asset.name === name)?.id;
+    expect(id).toBeDefined();
+    await manager.goToRealmStartPage("smartcity");
+    await manager.navigateToTab("asset");
+    await page.fill('#filterInput input[type="text"]', id!);
+    await expect(page.locator(`#list-container > ol > li`)).toHaveCount(2); // It's 2 instead of 1, because of the "asset draggable area" on the bottom.
```
Ekhorn: I would opt to add an `assetTree` fixture and replace every occurrence here.

```page.locator(`or-asset-tree [node-asset-id]:not([node-asset-id=""])`)```

```suggestion
    await expect(assetTree.items()).toHaveCount(1);
```

I would do the same for the filter, like `locator('or-asset-tree input')` and `locator('[data-selected] [node-asset-id]:not([node-asset-id=""])')` for the selected items.

I recommend having a separate locator for the `or-asset-tree` and reusing it in the above mentioned ones.

MartinaeyNL: Done 👍 

I've still used somewhat extensive locators, to ensure accessibility standards. (such as checking for `ol li` elements)
But they're now in separate files at least 😄

MartinaeyNL:
I have addressed the code comments. Let me know what you think @Ekhorn !

Ekhorn (approved):
Looks good :smile:

PR #2070, opened 2025-08-20
EKS creation scripts now allows multiple clusters in same AWS account

ebariaux:


wborn:
Do the Copilot remarks make sense @ebariaux?

ebariaux:
This should never happen but being defensive does not hurt.
I'll add the check but want to test it before pushing to PR

ebariaux:
@wborn changes applied

PR #2079, opened 2025-08-26
Updates to the Manager UI tests to support predefined realms

MartinaeyNL:
## Description
Work in progress.

## Checklist


MartinaeyNL:
For consistency purposes, we should always use the "Empty setup" without any setup assets.
Initially my thoughts were to keep the frontend tests "setup independent", but it would create a lot of additional headaches.
Such as "the scenario where an asset was already predefined with the same name as in the test" for example.
So I'll close this issue.

PR #2080, opened 2025-08-27
Enhancement: allow more than 10 dashboard widgets. 

Hackerberg43:
## Description
In my custom app, I have extended the or-dashboard-builder to register more custom widgets.
However, the amount is limited to 10 because the grid positioning array was static:
`const coords: Array<[number, number]> = new Array<[number, number]>([0,0],[2,0], [0,2], [2,2], [0,4], [2,4], [0,6], [2,6], [0,8], [2,8]) // TODO: make this unlimited possibilities with a formula`

This pr makes the array dynamic based on the amount of registered widgetTypes and enables a scroll overflow for the widget browser.
I have tested it by temporarily adding copies of widgets:
afbeelding

## Checklist
I cannot think of any.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -707,7 +707,7 @@ jobs:
         timeout-minutes: 20
 
       - name: Run install playwright browser(s)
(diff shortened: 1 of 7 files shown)


MartinaeyNL:
Thanks for contributing @Hackerberg43 ❤️

PR #2094, opened 2025-09-02
Execute all tests when running gradle test

Ekhorn:
## Description

This PR changes the `gradle npmTest` task to `gradle test` for UI components that do not apply the `Java` plugin, so that the `gradle test` task also runs the UI component tests. This also moves the dependency on the TypeScript compiler build to the `gradle test` command itself.

Furthermore the UI app tests that do apply the `Java` plugin, are 

Finally, the `test-backend-command` step has been changed to include all `gradle test` tasks except for the UI directory, to be able to also run unit tests in other projects besides the test project.

## Checklist


ui/build.gradle
Ekhorn, line 1: Shouldn't we maybe rename the `npmTest` task to `testUI` and then `npmTestUI` to something like `serveTestUI` or better yet, make the `testUI` task accept arguments like `--ui` or `--headed` (to pass along to the Playwright CLI)?

Ekhorn: @richturner what do you think? If we do change this then I need to update the frontend testing documentation as well.

ui/build.gradle
```diff
@@ -18,3 +18,20 @@ afterEvaluate {
 task modelWatch {
     dependsOn resolveTask(":ui:component:model:build"), resolveTask(":ui:component:rest:build")
 }
+
+// Finalize the gradle test task with the npmTest task so that all tests can be run with gradle test.
+// Runs backend tests first and then frontend tests. UI tests are registered under npmTest to avoid
```
Ekhorn:
```suggestion
// Runs backend tests first and then frontend tests. UI tests are registered under npmTest to avoid
```

Not exactly. It runs the gradle test task for that subproject first and these tasks. Ideally the frontend tests run after all the backend tests, because it is really heavy on the system and things may conflict.

Ekhorn:
The [latest commit](https://github.com/openremote/openremote/pull/2094/commits/d57fb7082bf9a3c67f1eb4a1b137a7d6ab2f2a04) is causing these issues #2096 and #2095.

PR #2097, opened 2025-09-03
Fix: modbus serial

Hackerberg43:
## Description
The current PLC4X rtu implementation is resulting in a dead loop when trying to setup a connection.
This is a known bug. However, because the jserialcom library was allready used in this project, we could use that instead and do the modbus parsing ourselves. Hereby a PR for that.
I'm pretty sure I'm not able to test all possible hardware/register scenario's, but coming from a non working implementation, everything is an improvement I guess ;)

The current testing setup is OR -> Docker -> WSL -> usbipd link /dev/ttyUSB0 to COM port -> Windows -> USB-UART RS485 converter (chips CH341 and CP2104 tested) -> Eastron SMD630 energy meter.

## Checklist

- [ ] 1. I still need to test different value types since the modbus device I'm currently testing with only has 32 bit floats.
- [ ] 2. Test writing


agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
```diff
@@ -20,35 +20,61 @@
...
+        }
+        
+        public static ModbusClientParity fromValue(int value) {
+            for (ModbusClientParity parity : values()) {
+                if (parity.value == value) {
+                    return parity;
+                }
+            }
+            return EVEN; // Default for Modbus RTU
+        }
```
richturner: Ordinal value seems to be sufficient don't see how this is an improvement on the previous enum

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialProtocol.java
```diff
@@ -19,41 +19,412 @@
...
+    
+    private int readWithTimeout(byte[] buffer, long timeoutMs) throws InterruptedException {
+        long startTime = System.currentTimeMillis();
+        int totalBytesRead = 0;
+        
+        while (totalBytesRead < buffer.length && (System.currentTimeMillis() - startTime) < timeoutMs) {
+            int available = serialPort.bytesAvailable();
+            if (available > 0) {
+                int bytesRead = serialPort.readBytes(buffer, buffer.length - totalBytesRead, totalBytesRead);
+                totalBytesRead += bytesRead;
```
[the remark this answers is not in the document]
Hackerberg43: wont hurt, added.

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialProtocol.java
```diff
@@ -19,41 +19,412 @@
...
+                LOG.warning("Incomplete write: " + bytesWritten + " of " + request.length + " bytes");
+                return null;
+            }
+            
+            // Calculate expected response length
+            int expectedResponseLength;
+            if (functionCode == 0x01 || functionCode == 0x02) {
+                expectedResponseLength = 5 + ((quantity + 7) / 8); // Boolean values packed in bytes
+            } else {
+                expectedResponseLength = 5 + (quantity * 2); // Each register is 2 bytes
```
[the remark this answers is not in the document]
Hackerberg43: this makes no sense to me

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialProtocol.java
```diff
@@ -19,41 +19,412 @@
...
+        return null;
+    }
+    
+    private Object parseModbusResponse(byte[] response, byte functionCode, ModbusAgentLink.ModbusDataType dataType) {
+        int byteCount = response[2] & 0xFF;
+        
+        if (functionCode == 0x01 || functionCode == 0x02) {
+            // Read coils or discrete inputs - return boolean
+            return (response[3] & 0x01) != 0;
+        } else if (functionCode == 0x03 || functionCode == 0x04) {
```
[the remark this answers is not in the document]
Hackerberg43: multiple bits per read arent supported yet but will prepare for it, added

Hackerberg43:
closing this for now not to cause unecessary ci/cd requests

pankalog:
@Hackerberg43 I'm reopening this to help with reviewing the progress so far

Hackerberg43:
Hey Panos, hope you don't feel offended by the amount of work im putting into this, it's just all adhoc development because we have a hardware setup where the extra features are directly put to use.

richturner:
Hey @Hackerberg43 it's great that you want to put the effort into this and the field testing will help make for a robust protocol, what you've done so far seems very well put together and I see some tests as well which is great.

I've had some discussion on this PR with @pankalog and we're keen to minimise any re-inventing of the wheel when it comes to parsing the MODBUS packets at least and Panos came across another modbus lib that uses netty and jserialcomm and also seems fairly active:

https://github.com/digitalpetri/modbus/ 

Having quickly looked through this lib at may not be structured in a way that is helpful for reuse but something to consider.

I haven't looked into your PR in detail but I have read your comms with Panos and from our quick look at the code some questions which we can discuss in our next meeting:

- Why do we need `SerialPortManager` this seems like a logical multiplexing of the serial port and it's unclear what this provides vs having the `Unit ID` in the `ModbusAgentLink` and handling it on a per linked attribute basis
- Complications around batch requests
- What is write polling

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusAgent.java
```diff
@@ -19,31 +19,93 @@
...
+    public enum EndianFormat {
+        BIG_ENDIAN,              // ABCD - Big byte order, Big word order
+        LITTLE_ENDIAN,           // DCBA - Little byte order, Little word order
+        BIG_ENDIAN_BYTE_SWAP,    // BADC - Big byte order, Little word order
+        LITTLE_ENDIAN_BYTE_SWAP; // CDAB - Little byte order, Big word order
+
+        @JsonValue
+        public String getJsonValue() {
+            return toString();
+        }
```
richturner: This is default jackson serialisation for enums so this is unnecessary

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusAgent.java
```diff
@@ -19,31 +19,93 @@
...
+    public static class ModbusDeviceConfig implements java.io.Serializable {
+        private EndianFormat endianFormat;
+        private String illegalRegisters;
+        private Integer maxRegisterLength;
+
+        @JsonCreator
+        public ModbusDeviceConfig(
+                @JsonProperty("endianFormat") EndianFormat endianFormat,
+                @JsonProperty("illegalRegisters") String illegalRegisters,
+                @JsonProperty("maxRegisterLength") Integer maxRegisterLength) {
```
richturner: `@JsonProperty` annotations not needed as we compile code with method arg names included

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusAgent.java
```diff
@@ -19,31 +19,93 @@
...
+            this.endianFormat = endianFormat != null ? endianFormat : EndianFormat.BIG_ENDIAN;
+            this.illegalRegisters = illegalRegisters != null ? illegalRegisters : "";
```
richturner: Not sure what value using an empty string has here vs `null`

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusAgent.java
```diff
@@ -19,31 +19,93 @@
...
+            this.maxRegisterLength = maxRegisterLength != null ? maxRegisterLength : 1;
```
richturner, line 57: Not sure what value using a `1` here vs `null` has, if you always want a value here then change the field type to `int`

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusAgent.java
```diff
@@ -19,31 +19,93 @@
...
+        }
+
+        /**
+         * Default configuration factory method
+         */
+        public static ModbusDeviceConfig createDefault() {
+            return new ModbusDeviceConfig(EndianFormat.BIG_ENDIAN, "", 1);
```
richturner: Consider an additional single arg constructor

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusAgent.java
```diff
@@ -19,31 +19,93 @@
...
+
+        public String getIllegalRegisters() {
+            return illegalRegisters;
+        }
+
+        public void setIllegalRegisters(String illegalRegisters) {
+            this.illegalRegisters = illegalRegisters;
+        }
+
+        public Integer getMaxRegisterLength() {
```
richturner: Change type to `int` if this must always have a value

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusAgent.java
```diff
@@ -19,31 +19,93 @@
...
+            return maxRegisterLength;
+        }
+
+        public void setMaxRegisterLength(Integer maxRegisterLength) {
```
richturner: Change to `int`

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
```diff
@@ -20,35 +20,59 @@
...
+        
+        public int getValue() {
+            return value;
+        }
+        
+        @JsonValue
+        public String getJsonValue() {
+            return toString();
+        }
```
richturner: Not required

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
```diff
@@ -20,35 +20,59 @@
...
+        
+        private final int value;
+        
+        ModbusClientParity(int value) {
+            this.value = value;
+        }
+        
+        public int getValue() {
+            return value;
+        }
```
richturner: `Enum.ordinal()` will provide the same so a simple enum can be used

richturner: see previous comment below

ui/component/or-icon/build.gradle
```diff
@@ -3,7 +3,7 @@ buildDir = "dist"
 tasks.register('copyMdiFont', Exec) {
     dependsOn getYarnInstallTask()
     // Using yarn throws an error relating to node_module resolution
-    commandLine npmCommand("npm"), "run", "copyMdiFont"
+    commandLine 'bash', '-c', 'npm run copyMdiFont'
```
richturner: What does this change solve?

ui/component/or-attribute-input/src/index.ts
```diff
@@ -173,7 +173,7 @@ export const jsonFormsInputTemplateProvider: (fallback: ValueInputProvider) => V
...
...ails) into the string or alternatively write the string through to the protocol as is (static string)"},"writeValueConverter":{"type":"object","patternProperties":{".+":{"type":["null","number","integer","boolean","string","array","object"]}},"description":"Similar to valueConverter but will be applied to outgoing values allowing for the opposite conversion"},"type":{"const":"ZWaveAgentLink","default":"ZWaveAgentLink"}},"title":"Z Wave Agent Link","additionalProperties":true,"required":["type"]}},"title":"Agent Link","oneOf":[{"$ref":"#/definitions/ModbusAgentLink"},{"$ref":"#/definitions/MockAgentLink"},{"$ref":"#/definitions/HTTPAgentLink"},{"$ref":"#/definitions/SimulatorAgentLink"},{"$ref":"#/definitions/BluetoothMeshAgentLink"},{"$ref":"#/definitions/StorageSimulatorAgentLink"},{"$ref":"#/definitions/MailAgentLink"},{"$ref":"#/definitions/VelbusAgentLink"},{"$ref":"#/definitions/DefaultAgentLink"},{"$ref":"#/definitions/MQTTAgentLink"},{"$ref":"#/definitions/WebsocketAgentLink"},{"$ref":"#/definitions/KNXAgentLink"},{"$ref":"#/definitions/SNMPAgentLink"},{"$ref":"#/definitions/ZWaveAgentLink"}],"type":"object","properties":{"type":{"const":"AgentLink"}},"required":["type"]}');
```
richturner: Please merge master and this is no longer required

richturner:
@Hackerberg43 please see my review comments but also it would be nice if the Serial implementation made use of the `SerialIOClient` as it will take care of reconnects and is good code reuse. Can also look at `VelbusSerialProtocol` which shows how to define packet encoders/decoders.

Hackerberg43:
This is ready for proper review, I've processed all your recommendations.

See https://github.com/Hackerberg43/documentation/tree/new-modbus-docs for updated documentation.

Remarks:

- PLC4X deleted, now just uses existing dependencies and openremote classes.
- The serial test port mock is a little bit flaky, it is stable with retries however. Just wasn't able to figure out the root cause.
- It was assumed that the existing modbus implementation was not yet used professionaly, so no migration/backwards compatibility was considered. With the decision of moving unitID to the agentLink, it wouldn't be straight-forward either.

Runs stable on our hardware HIL setup with around 5 serial devices and 3 TCP modbus gateways, total of around 50 single read/write registers at 1 second interval. While properly tested, we do not have all combinations of hardware configurations register types, so some combinations remain untested other than the backend tests.

Device config as a map makes neat use of the new dynamic jsonschema:
afbeelding
afbeelding

richturner:
Thanks for all the effort @Hackerberg43 I will review asap

agent/src/main/java/org/openremote/agent/protocol/serial/SerialIOClient.java
```diff
@@ -54,6 +55,10 @@ public SerialIOClient(String port, Integer baudRate) {
 
     @Override
     protected Class<? extends Channel> getChannelClass() {
+        Class<? extends Channel> testClass = testChannelClass.get();
+        if (testClass != null) {
+            return testClass;
+        }
```
richturner: This can be achieved with a Spy on the class that instantiates the `SerialIOClient` from within the test; you can then return a `TestSerialIOClient` that overrides this method, no test leakage into actual code.

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
```diff
@@ -20,35 +20,49 @@
...
 
-    //TODO: Doesn't work, getting a frontend error TypeError: Cannot read properties of undefined (reading 'units') at getValueFormatConstraintOrUnits
-//    public static final AttributeDescriptor<ModbusClientParity> PARITY = new AttributeDescriptor<ModbusClientParity>("parity",
-//            new ValueDescriptor<ModbusClientParity>("modbusClientParity", ModbusClientParity.class)
-//    );
+    public static final ValueDescriptor<StopBits> VALUE_STOPBITS = new ValueDescriptor<>("StopBits", StopBits.class);
+    public static final AttributeDescriptor<StopBits> STOP_BITS = new AttributeDescriptor<>("stopBits", VALUE_STOPBITS);
 
+    public static final ValueDescriptor<Paritybit> VALUE_PARITY = new ValueDescriptor<>("Paritybit", Paritybit.class);
+    public static final AttributeDescriptor<Paritybit> PARITY = new AttributeDescriptor<>("parity", VALUE_PARITY);
```
richturner: We should look to add these to `AgentLink` as `SERIAL_STOPBITS`, `SERIAL_DATABITS`, `SERIAL_PARITYBIT` and use these in `SerialAgent` thus negating the need for this sub class.

agent/src/main/java/org/openremote/agent/protocol/modbus/AbstractModbusProtocol.java
```diff
@@ -19,159 +19,883 @@
...
-    PlcConnection client = null;
 
-    public AbstractModbusProtocol(T agent) {
-        super(agent);
+    protected interface ModbusResponse {
+        boolean isException();
+        byte[] getPdu();
+        int getUnitId();
+        byte getFunctionCode();
     }
```
richturner: Move to own file

agent/src/main/java/org/openremote/agent/protocol/modbus/AbstractModbusProtocol.java
```diff
@@ -19,159 +19,883 @@
...
-            setConnectionStatus(ConnectionStatus.CONNECTING);
+        RegisterRange(int start, int end) {
+            this.start = start;
+            this.end = end;
+        }
+
+        boolean contains(int register) {
+            return register >= start && register <= end;
+        }
+    }
```
richturner: Move to own file don't do connection logic within this data type

agent/src/main/java/org/openremote/agent/protocol/modbus/AbstractModbusProtocol.java
```diff
@@ -19,159 +19,883 @@
...
-            setConnectionStatus(client.isConnected() ? ConnectionStatus.CONNECTED : ConnectionStatus.ERROR);
-        } catch (Exception e) {
-            LOG.log(Level.SEVERE, "Failed to create PLC4X connection for protocol instance: " + agent, e);
-            setConnectionStatus(ConnectionStatus.ERROR);
-            throw e;
+        void addAttribute(AttributeRef ref, int offset) {
+            attributes.add(ref);
+            offsets.add(offset);
         }
     }
```
richturner: Move to own file don't do connection logic within this data type

agent/src/main/java/org/openremote/agent/protocol/modbus/AbstractModbusProtocol.java
richturner, line 1: Just wondering why we can't extend `AbstractIOClientProtocol` as Modbus TCP should be very similar to `TCPProtocol` and Modbus Serial should be very similar to `SerialProtocol` different will be in the encoders/decoders and the use of `ModbusTcpFrame` and `ModbusSerialFrame`

agent/src/main/java/org/openremote/agent/protocol/modbus/AbstractModbusProtocol.java
```diff
@@ -19,159 +19,883 @@
...
+                    ranges.add(new RegisterRange(register, register));
+                } catch (NumberFormatException e) {
+                    LOG.warning(getProtocolName() + " - Invalid register format: " + part);
+                }
+            }
+        }
+        return ranges;
+    }
+
+    protected byte[] convertValueToRegisterBytes(Object value, int registerCount,
```
richturner: Try and keep the data parsing in its' own code files rather than mixed in with the protocol implementation as this file is quite large

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusTcpIOClient.java
```diff
@@ -0,0 +1,113 @@
...
+                out.writeBytes(frame.getPdu());
+            }
+
+            LOG.finest(() -> String.format("Encoded Modbus TCP frame: TxID=%d, UnitID=%d, FC=0x%02X, PDU length=%d",
+                frame.getTransactionId(), frame.getUnitId(), frame.getFunctionCode(),
+                frame.getPdu() != null ? frame.getPdu().length : 0));
+        }
+    }
+
+    public static class ModbusTcpDecoder extends io.netty.handler.codec.ByteToMessageDecoder {
```
richturner: Move to own code file

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusTcpIOClient.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        );
+    }
+
+
+    public int getNextTransactionId() {
+        return transactionIdCounter.updateAndGet(current -> (current + 1) % 65536);
+    }
+
+    public static class ModbusTcpEncoder extends io.netty.handler.codec.MessageToByteEncoder<ModbusTcpFrame> {
```
richturner: Move to own code file

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusTcpProtocol.java
```diff
@@ -19,36 +19,123 @@
...
     }
 
     @Override
-    public String getProtocolInstanceUri() {
-        return connectionString;
+    public String getProtocolName() {
+        return "TCP";
+    }
+
+    private ModbusTcpFrame sendRequestAndWaitForResponse(int unitId, byte[] pdu, long timeoutMs) throws Exception {
```
richturner: Consider using protected accessors to allow easier subclassing of this class in future

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusTcpProtocol.java
```diff
@@ -19,36 +19,123 @@
...
+        }
+
+        // Wait for response outside synchronized block to avoid blocking other threads
+        try {
+            ModbusTcpFrame response = responseFuture.get(timeoutMs, TimeUnit.MILLISECONDS);
+            LOG.finest(() -> "Received Modbus TCP response - TxID: " + response.getTransactionId() + ", UnitID: " + response.getUnitId() + ", FC: 0x" + Integer.toHexString(response.getFunctionCode() & 0xFF));
+            return response;
+        } catch (TimeoutException e) {
+            LOG.warning("Modbus TCP request timeout - TxID: " + transactionId + ", UnitID: " + unitId);
+            timedOutRequests.put(transactionId, System.currentTimeMillis());
```
richturner: This `timedOutRequests` map seems un-necessary, the only use of it seems to be able to identify future data that relates to a timedout request and log that fact, you've already logged the fact the request timed out so this complication is un-necessary.

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusTcpProtocol.java
```diff
@@ -19,36 +19,123 @@
...
+            if (timeoutTime != null) {
+                long latency = System.currentTimeMillis() - timeoutTime;
+                LOG.warning("Received late response for timed-out transaction ID " + txId + " (arrived " + latency + "ms after timeout)");
+            } else {
+                LOG.warning("Received response for unknown transaction ID: " + txId);
+            }
+        }
+
+        // Clean up old timed-out requests (older than 30 seconds)
+        long now = System.currentTimeMillis();
```
richturner: We always use `TimerService` time in code as PSEUDO clock is used in tests but this should be removed anyway

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
Hackerberg43: Hi Richard, all your comments have been addressed. But this one does not make much sense to me. These variables are tied to an agent, they may never very per agentlink. I agree that the SerialAgent should maybe also have them, but the modbusserialagent is an extension of modbusagent where commons with the modbustcpagent is shared.

Hackerberg43:
@richturner Ready for final review. 
-tcp connectionstatus update added via netty ChannelHandler ReadTimeoutHandler
-moved classes to separate files

agent/src/main/java/org/openremote/agent/protocol/modbus/util/ModbusProtocolCallback.java
richturner, line 1: A more appropriate name would be `AbstractModbusProtocol` (see `AbstractVelbusProtocol`) and it should be an abstract class; any duplicate methods in Serial and TCP implementations should be moved to here for good code reuse

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialIOClient.java
```diff
@@ -0,0 +1,54 @@
...
+        this.stopBits = stopBits;
+        this.parity = parity;
+    }
+
+    @Override
+    protected void configureChannel() {
+        super.configureChannel();
+        bootstrap.option(DATA_BITS, dataBits);
+        bootstrap.option(STOP_BITS, stopBits);
+        bootstrap.option(PARITY_BIT, parity);
```
richturner: Add these as options on the `SerialIOClient` and then this class is not needed

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusTcpIOClient.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static final Logger LOG = SyslogCategory.getLogger(PROTOCOL, ModbusTcpIOClient.class);
+
+    private final AtomicInteger transactionIdCounter = new AtomicInteger(0);
+
+    public ModbusTcpIOClient(String host, int port) {
+        super(host, port);
+    }
+
+    public int getNextTransactionId() {
```
richturner: This should be moved to where it is used (i.e. at the protocol level)

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusTcpIOClient.java
richturner, line 1: This class is not needed

agent/src/main/java/org/openremote/agent/protocol/modbus/AbstractModbusProtocol.java
richturner, line 1: This looks like code that would belong in the proposed `AbstractModbusProtocol`

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
```diff
@@ -20,36 +20,30 @@
...
     public static final AttributeDescriptor<Integer> BAUD_RATE = Agent.SERIAL_BAUDRATE.withOptional(false);
     public static final AttributeDescriptor<Integer> DATA_BITS = new AttributeDescriptor<>("dataBits", ValueType.POSITIVE_INTEGER);
-    public static final AttributeDescriptor<Integer> STOP_BITS = new AttributeDescriptor<>("stopBits", ValueType.POSITIVE_INTEGER);
 
-    //TODO: Doesn't work, getting a frontend error TypeError: Cannot read properties of undefined (reading 'units') at getValueFormatConstraintOrUnits
-//    public static final AttributeDescriptor<ModbusClientParity> PARITY = new AttributeDescriptor<ModbusClientParity>("parity",
-//            new ValueDescriptor<ModbusClientParity>("modbusClientParity", ModbusClientParity.class)
-//    );
+    public static final ValueDescriptor<Stopbits> VALUE_STOPBITS = new ValueDescriptor<>("Stopbits", Stopbits.class);
+    public static final AttributeDescriptor<Stopbits> STOP_BITS = new AttributeDescriptor<>("stopbits", VALUE_STOPBITS);
```
richturner, line 43: Move to Agent as `SERIAL_STOP_BITS`

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
```diff
@@ -20,36 +20,30 @@
...
 
-
-    public enum ModbusClientParity {
-        NO_PARITY,
-        ODD_PARITY,
-        EVEN_PARITY,
-        MARK_PARITY,
-        SPACE_PARITY,
-    }
```
richturner, line 52: This not needed for new implementation?

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
Hackerberg43: ?

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
Hackerberg43: Nope, now uses the enums from the Jserialcommlibrary directly

agent/src/main/java/org/openremote/agent/protocol/modbus/ModbusSerialAgent.java
richturner: This would make a dependency on Jserialcomm in the model which we'll avoid so can ignore this

Hackerberg43:
Hey Richard, the patch is in, but I skipped the modbusserialagent change because that was incomplete.
Can you explain why you would like to remove these fields, or their dependency on jserialcom types? It fuly relies on SerialIOClient which uses the same types and doesn't create any unecessary remapping of enumerations?
These properties are quite necessary to be able to adjust, since modbus RTU devices are often very simple which do not always have all the options configurable themselves.

richturner:
> Hey Richard, the patch is in, but I skipped the modbusserialagent change because that was incomplete. Can you explain why you would like to remove these fields, or their dependency on jserialcom types? It fuly relies on SerialIOClient which uses the same types and doesn't create any unecessary remapping of enumerations? These properties are quite necessary to be able to adjust, since modbus RTU devices are often very simple which do not always have all the options configurable themselves.

Apologies I didn't mean to include the partial changes to ModbusSerialAgent, I was trying to move the AttributeDescriptors out to the Agent class for reuse across any serial protocol but that would introduce a dependency on JSerialComm in the model module which we don't want so would need to create our own versions of those enums but we'll skip for now

richturner (approved):

Issue #2076, opened 2025-08-26
Simular agent handles predicated data


ebariaux:

ebariaux:
Use Datapoint API directly in SimulatorProtocol (predictedDatapointService available in AbstractProtocol)

ebariaux:
PR is combined with issue #2084

PR #2118, opened 2025-09-12
Simulator agent scheduling config and inserting predicted datapoints

### LINKED ISSUE #2084: Simulator agent: support scheduling configuration
Currently, the replay (and generation of predicted data points if applicable) loops over a 24h period, repeating forever.
We want to provide more flexibility by allowing configuration of this behaviour, but keep the existing default behaviour if no configuration is provided.

The following configuration is added (as protocol link fields):
- startDate:
  - always a date, no time information, considered to be 00:00 on that day
  - if not provided, starts immediately
- duration:
  - uses [ISO 8601 duration format](https://docs.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm)
  - if not provided, 24h
  - defines the length of the replay loop (and of the filled-in predicted data points if applicable), if replayData contains data points after this duration, those values are ignored and never used
- recurrence:
  - recurrence rule, following [RFC 5545 RRULE format](https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.5.3)
  - if not provided, repeats indefinitely daily

See [openremote/model/src/main/java/org/openremote/model/calendar/CalendarEvent.java at master · openremote/openremote](https://github.com/openremote/openremote/blob/master/model/src/main/java/org/openremote/model/calendar/CalendarEvent.java), it should be possible to use this existing class to implement the scheduling logic even if the fields defined in the UI are slightly different.

A user changing this configuration can potentially impact the predicted data points that have already been generated for the linked attribute. In that case, the predicted data points are re-generated to be in accordance with the new scheduling rules.

This issue refines and replaces [Simulator agent: support arbitrary replayDuration period · Issue #2077 · openremote/openremote](https://github.com/openremote/openremote/issues/2077) and [Simulator agent: support definition of activity period · Issue #2078 · openremote/openremote](https://github.com/openremote/openremote/issues/2078).

Ekhorn:
## Description

Closes #2084 and #2076

### Changelog

- Simulator agent now inserts predicted datapoints when `HAS_PREDICTED_DATA_POINTS` is configured.
- A start- and end date for the Simulator Agent can now be set.
- The Simulator Agent can now follow a recurring schedule.

## Checklist

### DIFF
--- a/agent/build.gradle
+++ b/agent/build.gradle
@@ -45,8 +45,9 @@ dependencies {
 
     implementation "org.jdom:jdom2:$jdomVersion"
 
-    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.1'
-    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.8.1'
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.13.4'
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.13.4'
+    testRuntimeOnly "org.junit.platform:junit-platform-launcher"
 }
 
 base {


test/src/test/groovy/org/openremote/test/protocol/simulator/SimulatorProtocolTest.groovy
Ekhorn, line 1: Test fail when running multiple at once, because the database transactions are interfering.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn, line 1: Generate new JSON Schema for `or-attribute-input` and fix descriptions.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +151,86 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
+            return null;
+        }
+
+//        if ((timeSinceCycleStarted == 0 && !schedule.hasRecurRule()) || (schedule.hasRecurRule() && >)) {
+//        }
+
+        try {
+            updateLinkedAttributePredictedDataPoints(attributeRef,
+                    // TODO: also include all next occurrence datapoints
+                    Arrays.stream(simulatorReplayDatapoints)
```
Ekhorn: I think I can simply copy the array and apply the delay value a second time.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +151,86 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
+        long nextRun;
+        try {
+            nextRun = schedule.getDelay(nextDatapoint.timestamp);
+        } catch (Exception e) {
+            LOG.warning(e.getMessage() + attributeRef);
+            return null;
+        }
+
+//        if ((timeSinceCycleStarted == 0 && !schedule.hasRecurRule()) || (schedule.hasRecurRule() && >)) {
+//        }
```
Ekhorn: I tried to only call inserting predicted datapoints when the duration cycle has started if no recurrence is set or if it is set the start of the current occurrence.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -102,6 +116,9 @@ protected void doUnlinkAttribute(String assetId, Attribute<?> attribute, Simulat
         if (updateValueFuture != null) {
             updateValueFuture.cancel(true);
         }
+
+        // Purge previously configured predicted datapoints // TODO: does this always get triggered when changing agentLinks?
```
Ekhorn: Couldn't manually test this because the transactions interfere

test/src/test/groovy/org/openremote/test/protocol/simulator/SimulatorProtocolTest.groovy
```diff
@@ -0,0 +1,548 @@
...
+            then: "datapoint is present"
+            conditions.eventually {
+                getDatapointTimestamp(attribute) == weeks + 3600 * 98
+            }
+
+            // TODO: make this easier to follow
+            and: "the delay is 167 hour"
+            conditions.eventually {
+                delay == 3600 * 167
+            }
```
Ekhorn: Resolve to do

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -20,17 +20,49 @@
...
+    @JsonPropertyDescription("Defines the length of the replay loop (and if configured the predicted datapoints to add)." +
+        " Specify an ISO 8601 duration formatted string. If not provided defaults to 24 hours." +
+        " If replayData contains data points after this duration, those values are ignored and never used.")
+    @JsonSerialize(using = ToStringSerializer.class)
+    @JsonDeserialize(converter = TimeUtil.PeriodAndDurationConverter.class)
+    // TODO: consider @JsonSchemaFormat("duration") requires new or-mwc-input type
+    @JsonSchemaInject(merge = false, jsonSupplierViaLookup = JSONSchemaUtil.SCHEMA_SUPPLIER_NAME_STRING_TYPE)
+    protected TimeUtil.ExtendedPeriodAndDuration duration;
+
+    @JsonPropertyDescription("How the replay period is  recurrence rule, following RFC 5545 RRULE format;" +
```
ebariaux: Review wording of description, not sure what you meant with "How the replay period is recurrence rule"

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -84,7 +95,10 @@ protected void doLinkAttribute(String assetId, Attribute<?> attribute, Simulator
             .ifPresent(simulatorReplayDatapoints -> {
                 LOG.info("Simulator replay data found for linked attribute: " + attribute);
                 AttributeRef attributeRef = new AttributeRef(assetId, attribute.getName());
-                ScheduledFuture<?> updateValueFuture = scheduleReplay(attributeRef, simulatorReplayDatapoints);
+
+                // Seed date for recurrence rule for when startDate is not specified
+                linkedAtMap.put(attributeRef, LocalDateTime.ofInstant(timerService.getNow(), ZoneId.of("UTC")));
```
Ekhorn: Use local timezone

container/src/main/java/org/openremote/container/timer/TimerService.java
```diff
@@ -104,6 +109,12 @@ public void stop() {
                 }
             }
 
+            @Override
+            public void stopAtTime(long time) {
+                stopTime = time;
```
ebariaux, line 114: You're always setting the stopTime, irrelevant of the fact there was already one or not (i.e. if the clock was running or not).
This is different from what the other stop() method does and confusing with the name.
But I see how you're using it in the tests and I faced the same problem on m...
You can see in BatteryChargingRulesTest the helper functions that I implemented, but we should bring that to the pseudo clock to avoid replicating in everywhere we need it.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +151,86 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
+        } catch (Exception e) {
+            LOG.warning(e.getMessage() + attributeRef);
+            return null;
+        }
+
+//        if ((timeSinceCycleStarted == 0 && !schedule.hasRecurRule()) || (schedule.hasRecurRule() && >)) {
+//        }
+
+        try {
+            updateLinkedAttributePredictedDataPoints(attributeRef,
```
Ekhorn: Move the configuration item check here so we don't compute for predicted datapoints when they are never inserted.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +151,86 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
+                    // TODO: also include all next occurrence datapoints
+                    Arrays.stream(simulatorReplayDatapoints)
+                            .map(d -> {
+                                try {
+                                    return d.setTimestamp(schedule.getDelay(d).getTimestamp().get() + now);
+                                } catch (Exception e) {
+                                    throw new RuntimeException(e);
+                                }
```
Ekhorn: Somehow this code is changing the nextRun property value.

```suggestion
                                try {
                                    return d.setTimestamp(schedule.getDelay(d).getTimestamp().get() + now);
                                } catch (Exception e) {
                                    throw new RuntimeException(e);
                                }
```

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -47,4 +79,128 @@ public SimulatorAgentLink setReplayData(SimulatorReplayDatapoint[] replayData) {
...
+            this.agentLink = agentLink;
+            this.now = now;
+            this.duration = agentLink.getDurationSeconds();
+            this.linkedAt = linkedAt;
+            this.recurrenceDelay = getRecurrenceDelay();
+        }
+
+        public long getDelay(long point) throws Exception {
+            if (recurrenceDelay.isEmpty()) {
+                throw new Exception("Next recurrence not found so replay cancelled: ");
```
ebariaux: I'm not a big fan of using an exception for something that is not really exceptional but just another branch in the business logic.
And the exception message with the trailing colon seems weird.
You're catching the exception in SimulatorProtocol, completing the message for logging purposes.
But when calling it from getDelay(SimulatorReplayDatapoint), exception bubbles up, becomes a Runtime without message and ends up as SEVERE log.
I would probably return an Optional value instead.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -20,17 +20,49 @@
...
-    @JsonPropertyDescription("Used to store 24h dataset of values that should be replayed (i.e. written to the" +
-        " linked attribute) in a continuous loop.")
+    @JsonPropertyDescription("Used to store a dataset of values that should be replayed (i.e. written to the" +
+        " linked attribute) in a continuous loop based on a schedule (by default replays every 24h)." +
+        " Predicted datapoints can be added by configuring 'Store predicted datapoints' which will insert the datapoints" +
+        " immediately as determined by the schedule.")
     protected SimulatorReplayDatapoint[] replayData;
 
+    @JsonPropertyDescription("Set a start date, if not provided, starts immediately." +
+        " When replay datapoint timestamp is 0 it starts at 00:00.")
```
Ekhorn:
```suggestion
        " When replay datapoint timestamp is 0 it will insert it at 00:00.")
```

What about negative values?

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -20,17 +20,49 @@
...
+    @JsonSchemaFormat("date")
+    protected LocalDate startDate;
+
+    @JsonPropertyDescription("Defines the length of the replay loop (and if configured the predicted datapoints to add)." +
+        " Specify an ISO 8601 duration formatted string. If not provided defaults to 24 hours." +
+        " If replayData contains data points after this duration, those values are ignored and never used.")
```
Ekhorn:
```suggestion
        " If replayData contains data points after this duration, those values are ignored and never used.")
```

Untested

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -20,17 +20,49 @@
...
+    @JsonSerialize(using = ToStringSerializer.class)
+    @JsonDeserialize(converter = TimeUtil.PeriodAndDurationConverter.class)
+    // TODO: consider @JsonSchemaFormat("duration") requires new or-mwc-input type
+    @JsonSchemaInject(merge = false, jsonSupplierViaLookup = JSONSchemaUtil.SCHEMA_SUPPLIER_NAME_STRING_TYPE)
+    protected TimeUtil.ExtendedPeriodAndDuration duration;
+
+    @JsonPropertyDescription("How the replay period is  recurrence rule, following RFC 5545 RRULE format;" +
+        " if not provided, repeats indefinitely daily")
```
Ekhorn: fix

```suggestion
    @JsonPropertyDescription("How the replay period is  recurrence rule, following RFC 5545 RRULE format;" +
        " if not provided, repeats indefinitely daily")
```

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +151,85 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
 
+    public void updateLinkedAttributePredictedDataPoints(AttributeRef attributeRef, List<ValueDatapoint<?>> values) {
+        Attribute<?> attribute = linkedAttributes.get(attributeRef);
+
+        if (attribute == null) {
+            LOG.log(Level.WARNING, () -> "Update linked attribute predicted data points called for un-linked attribute: " + attributeRef);
+            return;
+        }
+
+        predictedDatapointService.updateValues(attributeRef.getId(), attribute.getName(), values);
```
Ekhorn, line 170: These values are inserted with epoch seconds timestamps and not millisecond, fix.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +151,85 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
-        long nextRunRelative = nextRun - now;
 
-        LOG.fine("Next update for asset " + attributeRef.getId() + " for attribute " + attributeRef.getName() + " in " + nextRunRelative + " second(s)");
+        try {
+            if (attribute.getMeta().get(HAS_PREDICTED_DATA_POINTS).flatMap(AbstractNameValueHolder::getValue).orElse(false)) {
+                List<ValueDatapoint<?>> current = new ArrayList<>();
+                List<ValueDatapoint<?>> next = new ArrayList<>();
+                long nextRecurrenceDelay = schedule.getNextRecurrenceDelay();
+                for (SimulatorReplayDatapoint d : simulatorReplayDatapoints) {
+                    long timestamp = schedule.getDelay(d).getTimestamp().get() + now;
```
Ekhorn: During manual testing I found that the timestamps of the predicted datapoints are slightly off.

agent/src/test/java/org/openremote/agent/protocol/simulator/SimulatorAgentLinkTest.java
```diff
@@ -1,60 +1,124 @@
...
+//        assertEquals(40, schedule.getDelay(100));
+//        schedule = link.getSchedule(LocalDateTime.of(2000, 1, 1, 0, 2).toEpochSecond(ZoneOffset.UTC), LocalDateTime.of(2000, 1, 1, 0, 0));
+//        assertEquals(86_380, schedule.getDelay(100));
+//    }
+
+
+    @Test
+    public void get() throws Exception {
+
     }
```
Ekhorn: Remove

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -19,18 +19,36 @@
...
     protected SimulatorReplayDatapoint[] replayData;
 
+    // TODO: consider implementing @JsonSchemaFormat("calendar-event") to reuse the `or-rule-validity` component.
+    // Current generator cannot handle injecting custom type and description at the same time
+    @JsonPropertyDescription("When defined overwrites the possible dataset length and when it is replayed." +
+        " This could be once when only a start- (and end) date are defined," +
+        " or a recurring event following the RFC 5545 RRULE format." +
+        " If not provided defaults to 24 hours. If the replay data contains datapoints scheduled after the" +
+        " default 24 hours or the recurrence rule the datapoints will be ignored.")
+    protected Schedule schedule;
```
ebariaux: I think this should just be of type CalendarEvent, see below

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -47,4 +65,126 @@ public SimulatorAgentLink setReplayData(SimulatorReplayDatapoint[] replayData) {
...
+     * @throws Exception If the one-time occurrence or recurring schedule has ended.
+     */
+    public long getTimeUntilNextOccurrence(long timeSinceOccurrenceStarted) throws Exception {
+        if (getSchedule().isPresent()) {
+            return getSchedule().get().getTimeUntilNextOccurrence(timeSinceOccurrenceStarted);
+        }
+        return 86400L - timeSinceOccurrenceStarted;
+    }
+
+    public static class Schedule extends CalendarEvent {
```
ebariaux: I don't think Schedule should be a subclass of CalendarEvent.
I think it's more keeping a state, that's using a CalendarEvent for its logic.
Inheritance induces a tight coupling that I don't think is required here, an association would be enough (so Schedule using a CalendarEvent).
Also, the fact that all attribute declared here are marked as @JsonIgnore is an indication for me that something does not feel right.

I also don't think that this Schedule class (not sure about the name, maybe more something like RunningState ?) belongs in the AgentLink, I think it's only relevant for the Protocol.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -47,4 +65,126 @@ public SimulatorAgentLink setReplayData(SimulatorReplayDatapoint[] replayData) {
...
+         * specifies {@code UNTIL} or {@code COUNT} the previous occurrence start time is used.
+         *
+         * @param epoch Seconds since the epoch
+         * @return Seconds since the occurrence started
+         */
+        public long getTimeSinceOccurrenceStarted(long epoch) {
+            Recur<LocalDateTime> recurrence = super.getRecurrence();
+
+            if (recurrence == null || epoch < startTime) {
+                currentOccurrence = startTime;
```
ebariaux: currentOccurrence (and count below) are modified as part of a call to a getter.
This is a side effect I would not expect and it's dangerous because e.g. currentOccurrence is used in other methods (e.g. getTimeUntilNextOccurrence) that I suppose would expect this method to have been appropriately called before.
I would rather have a method named with some action that indicates we're altering the state and have the getter not alter the state (similar to how a GET method in HTTP is supposed to be idempotent).

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -47,4 +65,126 @@ public SimulatorAgentLink setReplayData(SimulatorReplayDatapoint[] replayData) {
...
+     * @throws Exception If the one-time occurrence or recurring schedule has ended.
+     */
+    public long getDelay(long point, long timeSinceOccurrenceStarted) throws Exception {
+        if (point <= timeSinceOccurrenceStarted) {
+            return point + getTimeUntilNextOccurrence(timeSinceOccurrenceStarted);
+        }
+        return point - timeSinceOccurrenceStarted;
+    }
+
+    public long getOccurrenceDuration() {
```
ebariaux: This code is similar to code at the end of Schedule.getTimeUntilNextOccurrence except you don't have the test on next not being null.
I think that given the way this is called, this should be OK but it's complicated to reason about and opens the door to a potential NPE.

If code is similar in both places, can we have 1 function that does this logic called from both places?
In fact, if we want to compute the duration of an occurrence, without taking into account the fact that there's another occurrence or not (this is a separate concern), can't this be added to CalendarEvent ?

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +139,104 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
+            .orElse(now % defaultReplayLoopDuration); // Remainder since 00:00
+
+        // Find datapoint with timestamp after the current occurrence
         SimulatorReplayDatapoint nextDatapoint = Arrays.stream(simulatorReplayDatapoints)
-            .filter(replaySimulatorDatapoint -> replaySimulatorDatapoint.timestamp > now)
+            .filter(replaySimulatorDatapoint -> replaySimulatorDatapoint.timestamp > timeSinceOccurrenceStarted)
             .findFirst()
             .orElse(simulatorReplayDatapoints[0]);
 
+        boolean singleOccurrence = schedule
```
ebariaux: isSingleOccurrence could be a method on CalendarEvent
You would still need to map / orElse but a bit of logic is encapsulated where it belongs (IMHO)

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +139,104 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
+                .map(s -> s.getStart() != null && s.getRecurrence() == null)
+                .orElse(false);
+        boolean lastDatapoint = singleOccurrence && Arrays.stream(simulatorReplayDatapoints)
```
ebariaux: I was first confused by the name, expecting the variable to store a data point. I would probably renamed to isLastDatapoint (same of isSingleOccurrence above though less confusing).

This is using nextDatapoint in its computation, but you're testing for nextDatapoint being null below, would do this after.
In fact, you're only using lastDatapoint much further in the code, I would move declaration of this closer to its usage location, makes the code more readable.

The way this is computed seems confusion to me.
Aren't we just trying to determine if nextDatapoint is the last entry in the simulatorReplayDatapoints array ? If so can't we compare timestamps of the last element with the one of nextDataPoint ?

ui/component/or-attribute-input/src/index.ts
Ekhorn, line 1: On the test server I don't see the schedule parameter on the simulator agent link anymore.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +139,104 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
             LOG.warning("Next datapoint not found so replay cancelled: " + attributeRef);
             return null;
         }
+
         long nextRun = nextDatapoint.timestamp;
-        if (nextRun <= now) { //now is after so nextRun is next day
-            nextRun += 86400; //day in seconds
+
+        try {
+            nextRun = agentLink.getDelay(nextRun, timeSinceOccurrenceStarted);
```
ebariaux: Having nextRun first defined as nextDatapoint, then re-computed here was a bit strange to me, would directly use nextDatapoint.timestamp in the getDelay call.
Also using optional instead of exception if no nextRun (as pointed in another comment) would simplify the way to write this code.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +139,104 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
             try {
                 updateLinkedAttribute(attributeRef, nextDatapoint.value);
             } catch (Exception e) {
                 LOG.log(Level.SEVERE, "Exception thrown when updating value: %s", e);
-            } finally {
-                replayMap.put(attributeRef, scheduleReplay(attributeRef, simulatorReplayDatapoints));
             }
-        }, nextRunRelative, TimeUnit.SECONDS);
+
+            if (lastDatapoint) {
```
ebariaux: Why do we have this specific test to handle lastDatapoint in a specific way ?
If it's the lastDatapoint, wouldn't the call the scheduleReplay below not do anything but return null ? And the future being null would remove the entry from the map.

When in doubt, you could isolate some logic in a specific method and have a unit test cover the different possibilities.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +139,104 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
+        try {
+            nextRun = agentLink.getDelay(nextRun, timeSinceOccurrenceStarted);
+        } catch (Exception e) {
+            LOG.warning(e.getMessage() + attributeRef);
+            return null;
         }
-        long nextRunRelative = nextRun - now;
 
-        LOG.fine("Next update for asset " + attributeRef.getId() + " for attribute " + attributeRef.getName() + " in " + nextRunRelative + " second(s)");
+        try {
```
ebariaux: For clarity, I would probably put all the logic dealing with predicted data points in its own method.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +139,104 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
+            if (attribute.getMeta().get(HAS_PREDICTED_DATA_POINTS).flatMap(AbstractNameValueHolder::getValue).orElse(false)) {
+                List<ValueDatapoint<?>> current = new ArrayList<>();
```
ebariaux: current and next are singular names but contains a collection, would probably try to be more explicit with names

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -134,40 +139,104 @@ public void updateSensor(AttributeRef attributeRef, Object value, long timestamp
...
+                long occurrenceDuration = 0;
+                if (!singleOccurrence) {
+                    occurrenceDuration = agentLink.getOccurrenceDuration();
+                }
+                for (SimulatorReplayDatapoint d : simulatorReplayDatapoints) {
+                    long timestamp = agentLink.getDelay(d.timestamp, timeSinceOccurrenceStarted) + now;
+                    current.add(new SimulatorReplayDatapoint(timestamp*1000, d.value).toValueDatapoint());
+                    if (!singleOccurrence) {
+                        // TODO: until next startdate will cause this value to be way further into the future
+                        next.add(new SimulatorReplayDatapoint((timestamp+occurrenceDuration)*1000, d.value).toValueDatapoint());
```
ebariaux: The way this is done, I think the next data points are not really required, because you're filling a full cycle window for each scheduled update.
I remember we said insert 2 times the window to make things simpler and that would be useful if you would not perform this operation on each schedule.
Filling in the full window on each schedule is costly though, because each time you'll really be inserting just one new value and updating all the datapoints with the same timestamp and value (so no-op).

ebariaux (changes requested):

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: Has been addressed, see https://github.com/openremote/openremote/blob/0a3dc5c359148fb98e3b3fc4e03cd6914b40bf65/model/src/main/java/org/openremote/model/calendar/CalendarEvent.java#L99

container/src/main/java/org/openremote/container/timer/TimerService.java
Ekhorn: Totally agree with this, I see the `advanceToMidnight` and `setClockToMidnightAt` helper methods if you meant those. I was thinking about making the tests "dynamic" following the current time, so we don't accidentally rely on weird circumstances in the test.

I'll have a look.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: Has changed since. Though I do have some new locations that throw exceptions.

See https://github.com/openremote/openremote/blob/0a3dc5c359148fb98e3b3fc4e03cd6914b40bf65/agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java#L179

And https://github.com/openremote/openremote/blob/0a3dc5c359148fb98e3b3fc4e03cd6914b40bf65/agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java#L187

Agree this should be handled differently.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: I did this to have the same JSON Schema output as the CalendarEvent class, while keeping this Simulator Agent specific code separated.

> I think it's more keeping a state, that's using a CalendarEvent for its logic.

Agree, shouldn't really need to extend it.

> I also don't think that this Schedule class (not sure about the name, maybe more something like RunningState ?) belongs in the AgentLink, I think it's only relevant for the Protocol.

My perspective on this is that we have a schedule you can configure from the UI. That schedule is a property on the SimulatorAgentLink class which is part of the JSON Schema used in the JSON Forms. I felt that it made sense to try and bundle this logic with the schedule. 

I get that it should indeed be part of the protocol as the AgentLink classes are more meant to be describing the model while the protocol is the implementation.

If I would refactor this I would think it makes sense to copy over the CalendarEvent class properties separately to the Schedule class, then move it into the protocol class and make the agent link import the schedule class.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: Forgot to remove the comment.

> I remember we said insert 2 times the window to make things simpler and that would be useful if you would not perform this operation on each schedule.

Performing this on each schedule is not ideal, but I didn't want to add more weird conditions, because during which schedule should it insert? -> Probably the first schedule of the current window (which can be half way through). I will add a sliding window like you mentioned, I might as well. 

> Filling in the full window on each schedule is costly though, because each time you'll really be inserting just one new value and updating all the datapoints with the same timestamp and value (so no-op).

I can filter out duplicate timestamps before they're passed to `scheduleReplay` (and maybe sort them).

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: > Aren't we just trying to determine if nextDatapoint is the last entry in the simulatorReplayDatapoints array ? If so can't we compare timestamps of the last element with the one of nextDataPoint ?

Only if we sort the `simulatorReplayDatapoints` beforehand.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: > In fact, if we want to compute the duration of an occurrence, without taking into account the fact that there's another occurrence or not (this is a separate concern), can't this be added to CalendarEvent ?

Implementation has changed since and the duration property is no longer independent from the occurrence, thus we cannot calculate the occurrence length without taking into account the next occurrence.

Only for a single occurrence we consider the last datapoint to be the end of the occurrence.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: > I will add a sliding window like you mentioned, I might as well.

NVM, since the duration can be different per occurrence this becomes considerably more tricky. I will keep it as is, but try to reduce the amount of times we compute datapoints (so handling no-ops).

Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/2fb77ee5532ccfa8d0fb13f967d29e9d7c03def5

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/09046e0c5cb67f613551e96faad4ffa60c392a88

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/09046e0c5cb67f613551e96faad4ffa60c392a88

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/b1ad1c1cb5542a1b285725b53a76ae7e113493dc

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/d7e2671dab8788520f4c34981838eb32f32b709a

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/d7e2671dab8788520f4c34981838eb32f32b709a and https://github.com/openremote/openremote/pull/2118/commits/fc621ba7989fb2f86d3ddfe4d36f93ec569260d1

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/fc621ba7989fb2f86d3ddfe4d36f93ec569260d1

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: The Schedule has `getOccurrenceDuration` and the protocol itself, because schedule only handles custom scheduled occurrences and the protocol handles the default daily loop.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/fb802100b692f74e43a00082df24fd74f5433ae2, however the naming I'm not sure about (see my comments above).

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: Made it a separate class as per https://github.com/openremote/openremote/pull/2118#discussion_r2362317047

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/d7e2671dab8788520f4c34981838eb32f32b709a

container/src/main/java/org/openremote/container/timer/TimerService.java
Ekhorn: Maybe a new issue to address test robustness for this topic could be made. I don't want to break something further.

Was also facing issues with flaky tests, see https://github.com/openremote/openremote/actions/runs/17898723982/job/50888836166. Tried to address that with https://github.com/openremote/openremote/pull/2118/commits/64967543f7b7f25fa022ccc162712a1ac2bdd73d, not sure if it worked.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: UPDATE: Moved to the `SimulatorAgentProtocol`.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2118/commits/5d7580d4c7f2ecb9af9438e438cfff013dc78714 by simply renaming the method.

I would consider splitting the data mutating part and the "get time since occurrence started" part for clarity maybe.

container/src/main/java/org/openremote/container/timer/TimerService.java
ebariaux: Created #2145 to address this.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
ebariaux: > > Aren't we just trying to determine if nextDatapoint is the last entry in the simulatorReplayDatapoints array ? If so can't we compare timestamps of the last element with the one of nextDataPoint ?
> 
> Only if we sort the `simulatorReplayDatapoints` beforehand.

I was under the impression that this was a requirement that the `simulatorReplayDatapoints` were sorted by timestamp ascending.
But indeed might very well be the case that we never check/enforce that anywhere and that users can enter the data points in any order.
We might want to see if we can do something about this in UI (noted as something to address in the future).

ebariaux (dismissed):

PR #2119, opened 2025-09-12
Robustness improvements for MQTT Handlers

richturner:


test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
```diff
@@ -162,6 +162,10 @@ class UserAndAssetProvisioningTest extends Specification implements ManagerConta
             subscribeFailures.add(topic)
             LOG.info("device1Client failed to subscribe to topic: ${topic}")
         }
+        Consumer<String> deviceNSubscribeFailureCallback = {String topic ->
```
ebariaux:
```suggestion
        Consumer<String> deviceNSubscribeFailureCallback = { String topic ->
```

manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java
```diff
@@ -212,7 +212,7 @@ public boolean canSubscribe(RemotingConnection connection, KeycloakSecurityConte
         if (allowed) {
             // Only allow if there is no existing subscription for this topic
             RemotingConnection existingConnection = responseSubscribedConnections.get(topic.toString());
-            if (existingConnection != null) {
+            if (existingConnection != null && existingConnection.getTransportConnection().isOpen()) {
```
ebariaux, line 215: I don't understand this whole mechanism in enough details but is there a chance of a race condition and the transport connection being in the process of being opened, so isOpen() returns false, we allow and then the connection opens ?

richturner: `responseSubscribedConnections` is only populated once a subscription is in place on the response topic to prevent multiple subscribers to a device's reponse topic, so don't think the scenario you describe applies here.

The reason I added this check was that in the test scenario where the client reconnects quickly the existingConnection is detected and so the client is denied subscription on the new session so I wanted to guard against this.

PR #2122, opened 2025-09-15
Review services to expose metrics under k8s

ebariaux:
There is a breaking change in this PR.
The key for annotations of the main service of the proxy pod has changed from service.annotations to service.http.annotations.
Please update your values files accordingly.

@wborn Does the above text somehow make it to the release notes when creating a release ? Or how can I ensure this information makes it to the release notes ?


wborn:
> Does the above text somehow make it to the release notes when creating a release ? Or how can I ensure this information makes it to the release notes ?

After a release has been created the release notes can be edited and any breaking changes can be added. We could also create a new label for PRs having breaking changes.

PR #2144, opened 2025-09-19
Fix identity provider mapper update logic and improve KeycloakCleanSetup

richturner:


manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -1079,6 +1079,21 @@ public Map<String, String> getIdentityProviderImportConfig(String realm, Map<Str
         return getRealms(realmsResource -> realmsResource.realm(realm).identityProviders().importFrom(importData));
     }
 
+    public List<IdentityProviderRepresentation> getIdentityProviders(String realm) {
+        return getRealms(realmsResource -> {
```
ebariaux, line 1083: Sometimes, like in deleteClient, we validate the input, validate that we have a non-null realmResource ...
In other methods, just like here, we don't do any validation.
Can we decide on one way or another, align or document when to do it or not

test/src/test/groovy/org/openremote/test/security/IdentityProviderTest.groovy
```diff
@@ -88,21 +89,9 @@ class IdentityProviderTest extends Specification implements ManagerContainerTrai
     }
 
     def "Configure Keycloak chaining"() {
-        def realmName = "idprealm"
-        def realmDisplayName = "IdP Realm"
-        def realmEnabled = true
+        def realmName = Constants.MASTER_REALM
```
ebariaux, line 92: Why is the
        when: "a realm does not yet exist in the identity provider"
test being removed ? Is it not something that can also be tested ?
Maybe in a different test if it's confusing in this scenario ?

ebariaux (commented):
Couple of questions

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
richturner: I've tried to rationalise the usage of the JAX-RS client resources and indicated `ClientErrorException` in throws

test/src/test/groovy/org/openremote/test/security/IdentityProviderTest.groovy
richturner: I changed to use the existing master realm

PR #2151, opened 2025-09-25
Enhancement: remove sensitive info

ebariaux:


ebariaux:
Question is do we want to try to remove account ids from the git history and force push it ?
This is rather tricky / risky.

richturner:
Don't think we should rewrite the git history, it's at least a little less accessible but given none of the info is extremely sensitive I don't see a problem of leaving it in the history but we've established good practice going forward on not putting such info in the codebase

PR #2161, opened 2025-09-26
UI: Added pagination support to asset tree

### LINKED ISSUE #1993: Asset tree: add UI pagination with "Load more" button on each level
## User Story
As a user of the Manager UI, I want to quickly retrieve my large dataset of assets (50.000+),
so that I can properly load and use the /assets page without too many hiccups or loading indicators.

## Description
As part of #1990, we're improving the performance of the Asset Tree UI.
Blocked by #1991 and #1992.

We'd like to add pagination using a "Load more" button on each level of the Asset Tree.
Initially, when opening the asset tree, it will request 1.000 parent assets without any parent children.
These parent assets will be shown (without children) in the UI, and showing an expander when
the `isParent` field of the DTO object equals to `true`. On the bottom of the list, we'll show an
"Load more" button, that will load asset 1000 to 2000 using an `offset` field in the `AssetQuery` object.
When clicking an expander, trying to show their children, we make an HTTP call requesting the children
of an asset. While it's busy fetching, we show an loading indicator or similar. When loaded,
similar to the parent level, we'll show a "Load more" button, to load additional children.

Note: the number 1000 is an placeholder here, we could update this based on HTTP JSON payload size.
Based on simple calculations, I expect 1000 assets to be around 105kb, so we can adjust this if necessary.

## Acceptance Criteria
- [ ] 1. Initially opening the asset tree, will load a limited amount of assets, within a second.
- [ ] 2. All assets will show a clickable chevron/expander when `isParent` is set to `true`, and
load a limited amount of its children, within a second.
- [ ] 3. Every 'layer'/level has a "Load more" button, allowing the user to load more assets, within a second.
- [ ] 4. All JSON payloads are analyzed and optimized to load within a second. If not, adjust the amount of assets.

These acceptance criteria can be adjusted based on feedback, or technical limitations during implementation.

## Implementation Notes
- Take into count that `or-asset-tree` is used across multiple pages, and not only on `/assets`.
If optimizations are made there, make sure the other implementations do not break.

MartinaeyNL:
## Description
Work in progress.

Fixes #1993.

## Changelog
Work in progress.

## Checklist

### DIFF
--- a/ui/component/or-asset-tree/src/index.ts
+++ b/ui/component/or-asset-tree/src/index.ts
@@ -13,7 +13,6 @@ import {
     AssetQueryMatch,
     AssetsEvent,
     AssetTreeNode,
-    AssetTypeInfo,
     Attribute,
     AttributePredicate,
     ClientRole,
@@ -33,6 +32,7 @@ import {ListItem} from "@openremote/or-mwc-components/or-mwc-list";
 import "@openremote/or-mwc-components/or-mwc-list";
 import {i18next} from "@openremote/or-translate";
 import "@openremote/or-mwc-components/or-mwc-dialog";
+import {virtualize} from "@lit-labs/virtualizer/virtualize.js"; // TODO: Seems a bit unstable with the asset tree layout; might reconsider to remove this.
 
 import {
     OrMwcDialog,
@@ -272,13 +272,7 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
      * Allows arbitrary assets to be displayed using a tree
      */
     @property({type: Array, reflect: false})
-    public assets?: Asset[];
-
-    @property({type: Object})
-    public assetInfos?: AssetTypeInfo[];
-
-    @property({type: Array})
-    public _assetIdsOverride?: string[];
+    public readonly assets?: Asset[];
 
     @property({type: Array})
     public rootAssets?: Asset[];
@@ -324,7 +318,7 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
 
     protected config?: AssetTreeConfig;
 
-    @property({attribute: false})
+    @state()
     protected _nodes?: UiAssetTreeNode[];
 
     protected _loading: boolean = false;
@@ -595,26 +589,35 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
                     </div>
                 </div>
             `)}
-            
-            ${!this._nodes
-                ? html`
-                    <span id="loading"><or-translate value="loading"></or-translate></span>`
-                : ((this._nodes.length === 0 || !this.atLeastOneNodeToBeShown())
-                            ? html `<span id="noAssetsFound"><or-translate value="noAssetsFound"></or-translate></span>` 
-                            : html`
-                    <div id="list-container">
-                        <ol id="list">
-                            ${this._nodes.map((treeNode) => this._treeNodeTemplate(treeNode, 0)).filter(t => !!t)}
-                            <li class="asset-list-element">    
-                                <div class="end-element" node-asset-id="${''}" @dragleave=${(ev: DragEvent) => { this._onDragLeave(ev) }} @dragenter="${(ev: DragEvent) => this._onDragEnter(ev)}" @dragend="${(ev: DragEvent) => this._onDragEnd(ev)}" @dragover="${(ev: DragEvent) => this._onDragOver(ev)}"></div>
+
+            ${when(!this._nodes, () => html`
+                <span id="loading"><or-translate value="loading"></or-translate></span>
+            `, () => html`
+                ${when(this._nodes!.length === 0 || !this.atLeastOneNodeToBeShown(),
+                    () => html`<span id="noAssetsFound"><or-translate value="noAssetsFound"></or-translate></span>`,
+                    () => {
+                        const nodes = [
+                            ...this._nodes!.filter(n => !n.hidden).map(node => html`${this._treeNodeTemplate(node, 0)}`),
+                            html`<li class="asset-list-element">
+                                <div class="end-element" node-asset-id="${''}" @dragleave=${(ev: DragEvent) => { this._onDragLeave(ev) }}
+                                     @dragenter="${(ev: DragEvent) => this._onDragEnter(ev)}" @dragend="${(ev: DragEvent) => this._onDragEnd(ev)}"
+                                     @dragover="${(ev: DragEvent) => this._onDragOver(ev)}"
+                                ></div>
                             </li>
-                        </ol>
-                    </div>
-                `)
-            }
+                        `];
+                        return html`
+                            <div id="list-container">
+                                <ol id="list">
+                                    ${virtualize({items: nodes, renderItem: (n: TemplateResult) => html`${n}`, scroller: true})}
+                                </ol>
+                            </div>
+                        `;
+                    }
+                )}
+            `)}
 
             <div id="footer">
-            
+
             </div>
         `;
     }
@@ -632,7 +635,7 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
         }
 
         if (!this._nodes) {
-            this._loadAssets();
+            this._loadAssets().catch(e => console.warn(e));
             return true;
         }
 
@@ -722,16 +725,28 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
         nodes.forEach((node) => this._updateSort(node.children, sortFunction));
     }
 
-    protected _toggleExpander(expander: HTMLElement, node: UiAssetTreeNode | null, silent: boolean = false) {
+    protected async _toggleExpander(expander: HTMLElement, node: UiAssetTreeNode | null, silent: boolean = false) {
         if (node && node.expandable) {
             node.expanded = !node.expanded;
 
             if (node.expanded) {
                 this._expandedNodes.push(node);
+
+                // Load children (from cache or using WebSocket) of the now 'expanded' node
+                const hasCachedChildren = node.children?.length;
+                if(hasCachedChildren) {
+                    console.debug(`Reusing cache for loading children of asset ${node.asset?.id}...`);
+                } else {
+                    const cache: Asset[] = [];
+                    OrAssetTree._forEachNodeRecursive(this._nodes ?? [], n => n.asset && cache.push(n.asset));
+                    await this._loadAssets(node.asset?.id, cache).catch(e => console.warn(e));
+                }
+
             } else {
                 this._expandedNodes = this._expandedNodes.filter(n => n !== node);
             }
 
+            // Update HTML attributes of the now 'expanded' node
             const elem = expander.parentElement!.parentElement!.parentElement!;
             elem.toggleAttribute("data-expanded");
             if (!silent) {
@@ -1154,6 +1169,13 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
                     }
                 });
             }
+
+            // Scroll to top of list
+            // TODO: Should this code be here? Or is this from a previous commit?
+            /*let listElement = this.shadowRoot?.getElementById('list');
+            if(listElement) {
+                listElement.scrollTop = 0;
+            }*/
         }
     }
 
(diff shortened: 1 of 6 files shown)


MartinaeyNL:
Ready for a review by @Ekhorn 👍 

Feel free to comment on anything, as well as the tests I wrote.
I'm relatively new to Playwright, so if some of my practices are uncommon, let me know.

ui/util/webpack.util.js
Ekhorn, line 1: Had enough of the error screen ? :laughing:

ui/component/or-asset-tree/src/index.ts
```diff
@@ -1667,66 +1711,110 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
...
+        if (!this._connected) {
+            throw new Error("Not connected to the server; cannot load assets.");
+        }
 
-            if (this._loading) {
-                return;
-            }
+        if (this._loading) {
+            throw new Error("Already loading assets for asset tree; ignoring request.");
+        }
```
Ekhorn, line 1751: Early return here seems like the right approach, unless I'm missing something.

```suggestion
        if (this._loading) {
            console.debug("Already loading assets for asset tree; ignoring request.");
            return;
        }
```

ui/component/or-asset-tree/src/index.ts
```diff
@@ -1237,26 +1264,46 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
...
-           this._filter.assetType.forEach((assetT: string) => {
-               if (this._assetTypes.findIndex((assetD: AssetDescriptor) => assetD.name === assetT) === -1) {
-                   showSnackbar(undefined, "filter.assetTypeDoesNotExist", "dismiss");
-               }
-           });
-           foundAssetIds = [];
-       }
+        try {
+            console.debug(`Querying assets using filter '${this._filterInput.nativeValue}'...`);
+            const promises = assetQueries.map(q => manager.rest.api.AssetResource.queryAssets(q)); // TODO: Use assetTree for more efficient querying.
```
Ekhorn: Comeback to this

ui/component/or-asset-tree/src/index.ts
```diff
@@ -627,16 +617,17 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
...
-            || _changedProperties.has("rootAssetIds")) {
-            this._nodes = undefined;
+            || _changedProperties.has("rootAssetIds")
+            || _changedProperties.has("queryLimit")) {
+            this.refresh();
         }
 
         if (!this._nodes) {
-            this._loadAssets();
+            this._loadAssets().catch(e => console.warn(e));
```
Ekhorn: shorthand :wink: 

```suggestion
            this._loadAssets().catch(console.warn);
```

ui/component/or-asset-tree/src/index.ts
```diff
@@ -627,16 +617,17 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
         const result = super.shouldUpdate(_changedProperties);
         if (_changedProperties.has("assets")
             || _changedProperties.has("rootAssets")
-            || _changedProperties.has("rootAssetIds")) {
-            this._nodes = undefined;
+            || _changedProperties.has("rootAssetIds")
+            || _changedProperties.has("queryLimit")) {
+            this.refresh();
         }
```
Ekhorn: Looks like its part of the condition.

```suggestion
            || _changedProperties.has("queryLimit")
        ) {
            this.refresh();
        }
```

ui/app/manager/test/fixtures/manager.ts
```diff
@@ -217,7 +217,11 @@ export class Manager {
    * @param asset The asset to create
    * @param config The axios request config
    */
-  async createAsset(asset: Asset, config) {
+  async createAsset(asset: Asset, config?: any) {
```
Ekhorn:
```suggestion
  async createAsset(asset: Asset, config?: AxiosRequestConfig<any>) {
```

ui/app/manager/test/fixtures/manager.ts
```diff
@@ -228,6 +232,26 @@ export class Manager {
       });
   }
 
+  /**
+   * Updates an asset
+   * @param asset The asset to update
+   * @param config The axios request config
+   */
+  async updateAsset(asset: Asset, config?: any) {
```
Ekhorn:
```suggestion
  async updateAsset(asset: Asset, config?: AxiosRequestConfig<any>) {
```

ui/app/manager/test/fixtures/manager.ts
Ekhorn: The manager doesn't directly depend on axios, `@openremote/rest` does though.

import { AxiosRequestConfig } from "axios";

You might as well fix this for every method.

ui/app/manager/test/assettree.test.ts
```diff
@@ -0,0 +1,403 @@
...
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
```
Ekhorn, line 19: Don't remember anything under `ui` having a license header. Not sure what the policy is.

```suggestion
```

ui/app/manager/test/assettree.test.ts
```diff
@@ -0,0 +1,403 @@
...
+test(`Load more buttons are shown when there are a lot of assets`, async ({ page, manager, assetsPage, assetTree }) => {
+    const assets = [batteryAsset, batteryAsset, electricityAsset, electricityAsset];
+    await manager.setup("smartcity", { assets: assets });
+    await manager.goToRealmStartPage("smartcity");
+    await assetsPage.goto();
+    await expect(assetTree.getAssetNodes()).toHaveCount(5); // 2 battery assets + 2 electricity assets + 1 console group
+    await page.locator('or-asset-tree').evaluate(tree => {(tree as OrAssetTree).setAttribute('queryLimit', '2')});
+    await expect(assetTree.getAssetNodes()).toHaveCount(2);
+    await page.locator('.loadmore-element or-mwc-input').click();
```
Ekhorn: I would consider the following over selecting based on CSS class. CSS classes are an implementation detail and you may want to change it at some point (e.g. when introducing new styling).

```suggestion
    await page.getByRole('button', { name: 'Load More' }).click();
```

Please consider updating everywhere where applicable.

ui/app/manager/test/assettree.test.ts
```diff
@@ -0,0 +1,403 @@
...
+test(`Load more buttons are shown properly without any unexpected scroll behavior`, async ({page, manager, assetsPage, assetTree}) => {
+    const batteryAssets = createBatteryAssets(50);
+    await manager.setup("smartcity", { assets: batteryAssets });
+    await applyParentAssets(parentAssets, manager);
+    await manager.goToRealmStartPage("smartcity");
+    await assetsPage.goto();
+
+    // Test if expanding the node works correctly
+    let cityAsset1 = assetTree.getAssetNodes().filter({ hasText: parentAssets[0].name });
+    await cityAsset1.locator('.expander[data-expandable]').click();
```
Ekhorn: Consider removing the CSS class selection from the locator. A dedicated attribute to indicate an element is expandable seems enough. 

```suggestion
    await cityAsset1.locator('[data-expandable]').click();
```

Please consider updating everywhere where applicable.

ui/app/manager/test/assettree.test.ts
```diff
@@ -0,0 +1,403 @@
...
+ * @then The asset detail page is displayed
+ */
+test(`Search for and select the battery asset`, async ({ page, manager, assetTree, assetsPage }) => {
+    await manager.setup("smartcity", { assets: [batteryAsset, electricityAsset] });
+    await manager.goToRealmStartPage("smartcity");
+    await assetsPage.goto();
+    await assetTree.fillFilterInput(batteryAsset.name);
+    await expect(assetTree.getAssetNodes()).toHaveCount(1);
+    await page.click(`text=${batteryAsset.name}`);
+    await expect(page.locator(`#asset-header`, { hasText: batteryAsset.name })).toBeVisible();
```
Ekhorn: We might as well add a dedicated method at this point to get the header in the `AssetViewer` fixture. Might seem silly, but this way we clearly define what we expect people to use.

```suggestion
    await expect(assetViewer.headerLocator(batteryAsset.name)).toBeVisible();
```

Please consider updating everywhere where applicable.

ui/component/or-asset-tree/test/fixtures/index.ts
```diff
@@ -22,7 +23,15 @@ export class AssetTree {
      * Returns a locator of all nodes that contain assets
      */
     getAssetNodes() {
-        return this.page.locator(`or-asset-tree ol li [node-asset-id]:not([node-asset-id=""])`);
+        return this.page.locator(`or-asset-tree ol li:has(.node-container[node-asset-id]:not([node-asset-id=""]))`);
```
Ekhorn, line 18: Consider removing the CSS class selector from the locator. The attribute should be enough.

I would also avoid including generic elements, either leave them out or if needed use ids.

You can reuse the locator from `getChildNodes`.

```suggestion
        return this.getChildNodes(this.page.locator(`or-asset-tree`)));
```

ui/component/or-asset-tree/test/fixtures/index.ts
Ekhorn, line 1: Unrelated, however we should remove `fillFilterInput` TBH. Just getting the locator using `getFilterInput` and calling `.fill` is good enough.

ui/component/or-asset-tree/test/fixtures/index.ts
```diff
@@ -22,7 +23,15 @@ export class AssetTree {
...
+    }
+
+    /**
+     * Returns a locator of all child nodes of the supplied node
+     * @param node - Locator pointing to the parent node
+     */
+    getChildNodes(node: Locator) {
+        return node.locator(`ol li .node-container[node-asset-id]:not([node-asset-id=""])`);
```
Ekhorn, line 26: See previous comment

```suggestion
        return node.locator(`[node-asset-id]:not([node-asset-id=""])`);
```

ui/app/manager/test/assettree.test.ts
```diff
@@ -0,0 +1,403 @@
...
+
+    // Delete battery asset 5 of the first city, and check if state is the same
+    await assetsPage.deleteSelectedAsset(manager, batteryAssets[4].name!);
+    await expect(assetTree.getSelectedNodes()).toHaveCount(0);
+    await expect(assetTree.getChildNodes(cityAsset1)).toHaveCount(assets.length / 2 - 1);
+    await expect(assetTree.getChildNodes(cityAsset2)).toHaveCount(assets.length / 2);
+    await expect(assetTree.getAssetNodes()).toHaveCount(1 + 9 + 1 + 10 + 1);
+    await expect(page.locator('or-asset-tree li.asset-list-element .expander[data-expandable]')).toHaveCount(3);
+    await expect(page.locator('or-asset-tree li.asset-list-element[data-expanded] .expander[data-expandable]')).toHaveCount(2);
+    await expect(page.locator('or-asset-viewer or-translate')).toHaveAttribute('value', 'noAssetSelected');
```
Ekhorn: Seems a little odd to depend on the `or-translate` element for this. I would rather just search for the text instead.

```suggestion
    await expect(page.locator('or-asset-viewer', { hasText: "Please select an asset on the left"})).toBeVisible();
```

ui/app/manager/test/assettree.test.ts
```diff
@@ -0,0 +1,403 @@
...
+    await expect(assetTree.getAssetNodes()).toHaveCount(1 + 10 + 1 + 10 + 1);
+
+    // Delete battery asset 5 of the first city, and check if state is the same
+    await assetsPage.deleteSelectedAsset(manager, batteryAssets[4].name!);
+    await expect(assetTree.getSelectedNodes()).toHaveCount(0);
+    await expect(assetTree.getChildNodes(cityAsset1)).toHaveCount(assets.length / 2 - 1);
+    await expect(assetTree.getChildNodes(cityAsset2)).toHaveCount(assets.length / 2);
+    await expect(assetTree.getAssetNodes()).toHaveCount(1 + 9 + 1 + 10 + 1);
+    await expect(page.locator('or-asset-tree li.asset-list-element .expander[data-expandable]')).toHaveCount(3);
+    await expect(page.locator('or-asset-tree li.asset-list-element[data-expanded] .expander[data-expandable]')).toHaveCount(2);
```
Ekhorn:
```suggestion
    await expect(assetTree.getExpandable()).toHaveCount(3);
    await expect(assetTree.getExpanded())).toHaveCount(2);
```

Consider the following selectors

```
locator('or-asset-tree [data-expandable]')
locator('or-asset-tree [data-expanded]')
```

ui/component/or-asset-tree/src/index.ts
```diff
@@ -1741,7 +1829,7 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
 
     public onEventsConnect() {
         this._connected = true;
-        this._loadAssets();
+        this._loadAssets().catch(e => console.warn(e));
```
Ekhorn:
```suggestion
        this._loadAssets().catch(console.warn);
```

ui/util/webpack.util.js
MartinaeyNL: Yes! 😉 

It's common to see some low-level errors, for example the `ResizeObserver` ones, that can simply be ignored.
Since it's visible in the console anyway, I don't see a reason to do a full error screen. It's annoying mostly.

I propose to just disable it.

ui/component/or-asset-tree/src/index.ts
MartinaeyNL: It's like this explicitly, since the function is asynchronous.
This way, the consumer can use `.catch()` to handle the error, and use `.then()` if it succeeds.
I prefer this way instead of returning `false`/`true` what we do at several other places.
I believe this is the "JavaScript standard" as well.

ui/component/or-asset-tree/src/index.ts
MartinaeyNL: Yeah, I'll make an issue for this to keep track, since it's work that's not necessary to pick up.
The only benefit is the amount of data to be downloaded for the client.
And, since the request is limited to 100 assets, it'll be only a few megabytes at most.

I'll link the issue below in a second.

MartinaeyNL: See #2192

ui/app/manager/test/fixtures/manager.ts
MartinaeyNL: Done! 👍

ui/component/or-asset-tree/test/fixtures/index.ts
Ekhorn: This one: https://github.com/openremote/openremote/pull/2161#discussion_r2429571366

ui/app/manager/test/assettree.test.ts
MartinaeyNL: Done!

ui/app/manager/test/assettree.test.ts
MartinaeyNL: Done!

ui/component/or-asset-tree/test/fixtures/index.ts
MartinaeyNL: There's a slight difference in `getChildNodes()` and `getAssetNodes()`,
where they use `li:has(.node-container)` and `li .node-container`. These are different selectors.

ui/component/or-asset-tree/test/fixtures/index.ts
MartinaeyNL: Done!

ui/app/manager/test/assettree.test.ts
MartinaeyNL: Done!

ui/component/or-asset-tree/test/fixtures/index.ts
MartinaeyNL: I explicitly added the `ol li` locator here for maintaining the UI accessibility using semantics.
In some cases they're overdone, so I removed them there. Preferably I'd keep them here.

Ekhorn (approved):
Looking good 😎 !

Issue #2172, opened 2025-09-29
Line chart: asset page


pierrekil:
The line chart on both the asset and insights page shows the wrong labels (shows five days) once period 'month' is selected. The visualised data is correct, only the labels not.

<img width="452" height="596" alt="Image" src="https://github.com/user-attachments/assets/6fbb1cae-27f1-478f-a968-6d6f2b906b93" />

<img width="838" height="280" alt="Image" src="https://github.com/user-attachments/assets/ab78500d-119a-4c34-a58d-91dd2ff4a242" />

Hackerberg43:
I see what is happening here. 25-sep is actually 1st september '25, there is a typo in the format. Proposed format: 1-Sep-'25.

PR #2173, opened 2025-09-29
UI bugfix for the date formatting in Chart being incorrect

Hackerberg43:
Fixes https://github.com/openremote/openremote/issues/2172
Maybe good to take a quick look if all the formattings are to your liking.
Echarts triggers these formats when exactly at a round number, so uses year format when at 00:00 1st jan 2025, month when at 00:00 1st of feb, day when at 00:00 etc.

### DIFF
--- a/ui/component/or-attribute-barchart/src/index.ts
+++ b/ui/component/or-attribute-barchart/src/index.ts
@@ -1246,9 +1246,9 @@ export class OrAttributeBarChart extends LitElement {
                     interval: this._intervalConfig?.millis,
                     fontSize: 10,
                     formatter: {
-                        year: "{yyyy}",
-                        month: "{MMMM} '{yy}",
-                        day: "{MMM} {d}th",
+                        year: "1-{MMM}-{yyyy}",
+                        month: "1-{MMM}-'{yy}",
+                        day: "{d}-{MMM}",
                         hour: "{HH}:{mm}",
                         minute: "{HH}:{mm}",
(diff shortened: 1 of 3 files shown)


MartinaeyNL:
@Hackerberg43 Thanks for the PR!
I applied the same changes to `or-attribute-history` aka the assets page, since the issue occurred there as well.

I've also adjusted the Bar Chart widget to use the same format, is this OK with you?
Feel free to adjust it if necessary.

Hackerberg43:
Ah yes offcourse.

MartinaeyNL:
No problem! I'm happy to approve and merge this 👍 

I see there is an issue with the CI/CD failing, which probably is an issue on our end.
I'll check internally if we've enforced something that causes forks of OpenRemote to fail.

Hackerberg43:
I'm still a little bit in two minds about the formatting of days.
Like showing one day, the start of the day has no time. It makes sense but it is not 100% clear.
afbeelding
However if you add the time to it, when showing days in weeks/months it will also add 00:00 behind each tick, which is a bit messy I'd think.

Hackerberg43:
If you are fine with the above current format, ready to merge I'd say

MartinaeyNL:
Yeah sorry. I was hoping that internally someone could find time to solve the CI/CD bug.
Some recent changes causes the forks to fail, see #2176 

But I'm confident enough to merge it without a CI/CD check. 👍

PR #2174, opened 2025-09-29
Fix deployment decryption build config since gradle 9.1.0 migration

### LINKED ISSUE #2169: CI/CD: Demo setup JAR no longer included in the deployment image

richturner:
Closes #2169


richturner:
This needs to be done in all custom projects either before or as they are migrated to gradle 9.x

ebariaux:
Sorry about that issue, was not exercising that path locally.

In general, I believe having such logic in project.gradle is not really recommended.
Using custom gradle plugins might be a route to explore here.

Regarding custom projects, should I create GH issues to address each one of them ?

PR #2180, opened 2025-10-06
UI: Bugfix for agents not showing asset import and protocol discovery panels when both supported 

### LINKED ISSUE #2167: Agent asset import and asset autodiscovery GUI bug
When a user adds an agent that supports asset import (ProtocolAssetImport) and asset autodiscovery (ProtocolAssetDiscovery) only the UI elements that are related to the ProtocolAssetImport interface are shown in the GUI.

1. Add an agent to the asset tree that supports asset import and asset autodiscovery
2. Select the agent in the asset tree
3. Scroll down to the 'SETUP' group
4. The 'SETUP' group contains only the 'SELECT FILE' and file upload button.

**Expected behavior**
The 'SETUP' group should also contain the 'DISCOVER ASSETS' button.

Image

Image

richturner:
Fixes #2167

### DIFF
--- a/ui/component/or-asset-viewer/src/index.ts
+++ b/ui/component/or-asset-viewer/src/index.ts
@@ -531,22 +531,24 @@ function getPanelContent(id: string, assetInfo: AssetInfo, hostElement: LitEleme
                 </div>
             `;
         }
-        else if (descriptor.assetDiscovery) {
+        if (descriptor.assetDiscovery) {
             content = html`
-                <or-mwc-input outlined id="discover-btn" .type="${InputType.BUTTON}" label="discoverAssets" @or-mwc-input-changed="${() => discoverAssets()}"></or-mwc-input>
-                <or-mwc-input id="cancel-discover-btn" .type="${InputType.BUTTON}" label="cancel" @or-mwc-input-changed="${() => cancelDiscovery()}" hidden style="margin-left:20px"></or-mwc-input>
+                ${content}
+                <div id="discovery">
+                    <or-mwc-input outlined id="discover-btn" .type="${InputType.BUTTON}" label="discoverAssets" @or-mwc-input-changed="${() => discoverAssets()}"></or-mwc-input>
+                    <or-mwc-input id="cancel-discover-btn" .type="${InputType.BUTTON}" label="cancel" @or-mwc-input-changed="${() => cancelDiscovery()}" hidden style="margin-left:20px"></or-mwc-input>
+                </div>
             `;
-        } else {
+        }
+
+        if (!descriptor.assetImport && ! descriptor.assetDiscovery) {
             showSnackbar(undefined, "agent type doesn't support a known protocol to add assets", "dismiss");
         }
 
         return html`
-            <style>
-                [hidden] {
-                    display: none;
-                }
-            </style>
-            ${content}
+            <div id="setup-wrapper">
+                ${content}
+            </div>
         `;
 
     }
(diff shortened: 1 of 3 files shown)


ui/component/or-asset-viewer/src/index.ts
```diff
@@ -531,22 +531,24 @@ function getPanelContent(id: string, assetInfo: AssetInfo, hostElement: LitEleme
...
-            <style>
-                [hidden] {
-                    display: none;
-                }
-            </style>
-            ${content}
+            <div id="setup-wrapper">
+                ${content}
+            </div>
         `;
```
Ekhorn, line 542: I recommend replacing line 520-552 with the following. Using conditional directives (`import { when } from "lit/directives/when.js";`) from lit as mutating like this is a little confusing. Also the wrapper seems redundant to me (see my next comment).

```suggestion
        if (!descriptor.assetImport && !descriptor.assetDiscovery) {
            showSnackbar(undefined, "agent type doesn't support a known protocol to add assets", "dismiss");
        }

        return html`
            ${when(descriptor.assetImport, () => html`
                <div id="fileupload">
                    <or-mwc-input style="flex: 0 1 auto;" outlined label="selectFile" .type="${InputType.BUTTON}" @or-mwc-input-changed="${() => hostElement.shadowRoot!.getElementById('fileupload-elem')!.click()}">
                        <input id="fileupload-elem" name="configfile" type="file" accept=".*" @change="${() => updateFileName()}"/>
                    </or-mwc-input>
                    <or-mwc-input style="flex: 1 1 auto; margin: 0 4px 0 10px;" id="filename-elem" .label="${i18next.t("file")}" .type="${InputType.TEXT}" disabled></or-mwc-input>
                    <or-mwc-input style="flex: 0 1 auto;" id="fileupload-btn" icon="upload" .type="${InputType.BUTTON}" @or-mwc-input-changed="${() => doImport()}" disabled></or-mwc-input>
                    <progress id="progress-circular" class="hidden pure-material-progress-circular"></progress>
                </div>
            `)}
            ${when(descriptor.assetDiscovery, () => html`
                <div id="discovery">
                    <or-mwc-input outlined id="discover-btn" .type="${InputType.BUTTON}" label="discoverAssets" @or-mwc-input-changed="${() => discoverAssets()}"></or-mwc-input>
                    <or-mwc-input id="cancel-discover-btn" .type="${InputType.BUTTON}" label="cancel" @or-mwc-input-changed="${() => cancelDiscovery()}" hidden style="margin-left:20px"></or-mwc-input>
                </div>
            `)}
        `;
```

ui/component/or-asset-viewer/src/style.ts
```diff
@@ -256,13 +256,19 @@ export const style = css`
         display: none;
         cursor: pointer;
     }
-    
-    #fileupload {
+
+    #setup-wrapper {
+        width: 100%;
+        margin-top: -10px;
+    }
```
Ekhorn: Seems redundant

```suggestion
```

Ekhorn: We have the following code that handles margins across panels for the first and last element. Currently these are applied, but using margin-top -10px seems like bad practice. However some panel elements aren't applying these styles (I think only the attributes panel applies these styles properly). You could add a `!important` declaration, however small inconsistencies like these feel like something to let @remyberden go over at some point.

```
.panel-content > :first-child {
    margin-top: 0;
}
.panel-content > :last-child {
    margin-bottom: 0;
}
```

Ekhorn (commented):
Works fine. Just some styling inconsistencies.

ui/component/or-asset-viewer/src/style.ts
richturner: I just wanted to get a quick fix for this issue but indeed this needs another look as part of bigger issue

PR #2190, opened 2025-10-13
Update dependencies except critical ones

ebariaux:


ebariaux:
As I've seen some new high vulnerabilities that should be fixed with dependencies updates, I did a round of all available updates.

PR #2194, opened 2025-10-16
Dynamic schema generation with translation support

### LINKED ISSUE #1766: Dynamic JSON Forms: Add the ability to translate the JSON Schema descriptions

Ekhorn:
## Description

Adds an endpoint to generate JSON Schemas by providing a `valueDescriptor` name. We proactively generate common JSON Schemas including a map of hashes linking to each type. This makes cache keys accessible through the `/api/master/info` endpoint, in preparation for https://github.com/openremote/openremote/issues/1768 https://github.com/openremote/openremote/issues/1767 https://github.com/openremote/openremote/issues/1704.

image

> Note: these hashes are hex values (base64 encoding would reduce their size), we expect the JSON Schema generator to currently output a schema the same way every time. So we assume the order of properties doesn't change. However changing them in a class currently might result in a different schema and thus in a different hash value. If we want to prevent this we need to update the generator to sort all properties and list items recursively.

JSON Schemas are intended to be fetched like so: 

```sh
curl "http://localhost:8080/api/master/model/getValueDescriptorSchema?name=valueConstraint[]&hash=E04EBD3AE7D2B3472B84C387C7784279
```

This PR also adds translation support to configuration items. It adds `i18n` keys to the JSON Schema of classes that apply the `@JsonSchemaTitle` and `@JsonSchemaDescription` annotations that refer to the class by their FQCN. When titles or descriptions are added inside the model they can be added(/updated) in the translation files using the `processTranslations` task. The pipeline will fail if this task finds that translations haven't been updated.

Closes #1766

## Changelog

- Add dynamic JSON Forms for configuration items
- Add `getValueDescriptorSchema` endpoint
- Add `metaItem` cache key map to `getInfo` endpoint
- Add translation support for configuration items

## Checklist

### DIFF
--- a/agent/src/main/java/org/openremote/agent/protocol/openweathermap/OpenWeatherMapAgent.java
+++ b/agent/src/main/java/org/openremote/agent/protocol/openweathermap/OpenWeatherMapAgent.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2025, OpenRemote Inc.
+ *
+ * See the CONTRIBUTORS.txt file in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.openremote.agent.protocol.openweathermap;
+
+import java.util.Optional;
+
+import jakarta.persistence.Entity;
+import org.openremote.model.asset.agent.Agent;
+import org.openremote.model.asset.agent.AgentDescriptor;
+import org.openremote.model.attribute.MetaItem;
+import org.openremote.model.value.AttributeDescriptor;
+import org.openremote.model.value.MetaItemType;
+import org.openremote.model.value.ValueType;
+
+@Entity
+public class OpenWeatherMapAgent extends Agent<OpenWeatherMapAgent, OpenWeatherMapProtocol, OpenWeatherMapAgentLink> {
+
+    public static final AgentDescriptor<OpenWeatherMapAgent, OpenWeatherMapProtocol, OpenWeatherMapAgentLink> DESCRIPTOR = new AgentDescriptor<>(
+            OpenWeatherMapAgent.class, OpenWeatherMapProtocol.class, OpenWeatherMapAgentLink.class);
+
+    public static final AttributeDescriptor<String> API_KEY = new AttributeDescriptor<>("ApiKey", ValueType.TEXT);
+
+    // Read only attribution, required by OpenWeatherMap API
+    // Set during protocol start
+    public static final AttributeDescriptor<String> ATTRIBUTION = new AttributeDescriptor<>("Attribution", ValueType.TEXT,
+            new MetaItem<>(MetaItemType.READ_ONLY));
+
+    // Button to provision a weather asset with the appropriate agent links
+    public static final AttributeDescriptor<Boolean> PROVISION_WEATHER_ASSET = new AttributeDescriptor<>("ProvisionWeatherAsset", ValueType.BOOLEAN);
+
+    protected OpenWeatherMapAgent() {
+    }
+
+    public OpenWeatherMapAgent(String name) {
+        super(name);
+    }
+
+    public Optional<String> getAPIKey() {
+        return getAttributes().getValue(API_KEY);
+    }
+
+    public OpenWeatherMapAgent setAPIKey(String value) {
+        getAttributes().getOrCreate(API_KEY).setValue(value);
+        return this;
+    }
+
+    public Optional<Boolean> getProvisionWeatherAsset() {
+        return getAttributes().getValue(PROVISION_WEATHER_ASSET);
+    }
+
+    public OpenWeatherMapAgent setProvisionWeatherAsset(Boolean value) {
+        getAttributes().getOrCreate(PROVISION_WEATHER_ASSET).setValue(value);
+        return this;
+    }
+
+    @Override
+    public OpenWeatherMapProtocol getProtocolInstance() {
+        return new OpenWeatherMapProtocol(this);
+    }
+
+}
--- a/agent/src/main/java/org/openremote/agent/protocol/openweathermap/OpenWeatherMapAgentLink.java
+++ b/agent/src/main/java/org/openremote/agent/protocol/openweathermap/OpenWeatherMapAgentLink.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2025, OpenRemote Inc.
+ *
+ * See the CONTRIBUTORS.txt file in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.openremote.agent.protocol.openweathermap;
+
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+import jakarta.validation.constraints.NotNull;
+import org.openremote.model.asset.agent.AgentLink;
+
+public class OpenWeatherMapAgentLink extends AgentLink<OpenWeatherMapAgentLink> {
+
+    @NotNull
+    @JsonPropertyDescription("Select which weather property (e.g. temperature, humidity) to use as the data source")
+    private OpenWeatherMapProperty weatherProperty;
+
+    // For Hydrators
+    protected OpenWeatherMapAgentLink() {
+    }
+
+    public OpenWeatherMapAgentLink(String id) {
+        super(id);
+    }
+
+    public OpenWeatherMapProperty getWeatherProperty() {
+        return weatherProperty;
+    }
+
+    public OpenWeatherMapAgentLink setWeatherProperty(OpenWeatherMapProperty weatherProperty) {
+        this.weatherProperty = weatherProperty;
+        return this;
+    }
+
+}
--- a/agent/src/main/java/org/openremote/agent/protocol/openweathermap/OpenWeatherMapProperty.java
+++ b/agent/src/main/java/org/openremote/agent/protocol/openweathermap/OpenWeatherMapProperty.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2025, OpenRemote Inc.
+ *
+ * See the CONTRIBUTORS.txt file in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.openremote.agent.protocol.openweathermap;
+
+/*
+ * These are the core weather properties that are returned by the OpenWeatherMap API
+ * It is used by the {@link OpenWeatherMapAgentLink} to determine which property to use as the data source for
+ * updating the attribute.
+ */
+public enum OpenWeatherMapProperty {
+    TEMPERATURE,                            // Temperature
+    ATMOSPHERIC_PRESSURE,                   // Atmospheric pressure
+    HUMIDITY,                               // Humidity percentage
+    CLOUD_COVERAGE,                         // Cloud coverage percentage
+    WIND_SPEED,                             // Wind speed (m/s)
+    WIND_DIRECTION,                         // Wind direction (degrees)
+    WIND_GUST_SPEED,                        // Wind gust (m/s)
+    PROBABILITY_OF_PRECIPITATION,           // Probability of precipitation
+    RAIN_AMOUNT,                            // Rain amount (mm)
+    ULTRAVIOLET_INDEX                       // Ultraviolet index
+
+}
--- a/agent/src/main/java/org/openremote/agent/protocol/openweathermap/OpenWeatherMapProtocol.java
+++ b/agent/src/main/java/org/openremote/agent/protocol/openweathermap/OpenWeatherMapProtocol.java
@@ -1,0 +1,483 @@
+/*
+ * Copyright 2025, OpenRemote Inc.
+ *
+ * See the CONTRIBUTORS.txt file in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.openremote.agent.protocol.openweathermap;
+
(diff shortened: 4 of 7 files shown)


manager/src/main/java/org/openremote/manager/asset/AssetModelService.java
```diff
@@ -240,6 +247,8 @@ public void init(Container container) throws Exception {
         );
 
         container.getService(MessageBrokerService.class).getContext().addRoutes(this);
+
+        dynamicJsonSchemas.putAll(ValueUtil.getJsonSchemas());
```
Ekhorn: Should not have duplicate copies of these schemas.

ui/app/shared/locales/ar/or.json
```diff
@@ -212,8 +212,8 @@
   "buttonTextConfirm": "نص لزر التفعيل",
   "buttonTextDecline": "نص لزر عدم التفعيل",
   "custom": "إدخال يدوي",
-  "realm": "نطاق",
-  "realm_plural": "النطاقات",
+  "realm": "عالم دنيا",
+  "realm_plural": "عالم دنيا",
```
Ekhorn, line 216: @richturner I found duplicate translations. Pierre added `realm` and `realm_plural` 4 years ago, and you added the same but translated differently 3 years ago. I removed the newest (yours). Let me know if we should keep yours instead.

ui/app/shared/locales/cn/or.json
```diff
@@ -225,8 +225,8 @@
   "buttonTextConfirm": "确认按键文本",
   "buttonTextDecline": "拒绝按键文本",
   "custom": "自定义参数",
-  "realm": "租户",
-  "realm_plural": "租户",
+  "realm": "场所",
+  "realm_plural": "场所",
```
Ekhorn, line 229: See https://github.com/openremote/openremote/pull/2194#discussion_r2435889978

model/src/main/java/org/openremote/model/asset/AssetModelResource.java
```diff
@@ -99,4 +101,15 @@ public interface AssetModelResource {
...
+    /**
+     * Retrieve the JSON Schema for a {@link ValueDescriptor} available in this system. A value descriptor schema is only meant to be retrieved
+     * once per client. Either when a new {@code version}, {@code descriptorType} or {@code arrayDimensions} are requested. The HTTP client should cache the response based on the
+     * {@code Cache-Control} header. The {@code version} for actively cached descriptors can be found by requesting the {@link StatusResource#getInfo()} endpoint.
+     */
+    @GET
+    @Path("getValueDescriptorSchema")
+    @Produces(APPLICATION_JSON)
+    @Operation(operationId = "getValueDescriptorSchema", summary = "Retrieve the valueDescriptor JSON Schema.")
+    Response getValueDescriptorSchema(@BeanParam RequestParams requestParams, @QueryParam("version") String version, @QueryParam("descriptorType") String descriptorType, @QueryParam("arrayDimensions") Integer arrayDimensions);
```
Ekhorn: Might want to consider renaming the following for clarity:

```suggestion
    Response getValueDescriptorSchema(@BeanParam RequestParams requestParams, @QueryParam("hash") String hash, @QueryParam("descriptorType") String descriptorType, @QueryParam("arrayDimensions") Integer arrayDimensions);
```

.github/workflows/ci_cd.yml
```diff
@@ -546,6 +546,12 @@ jobs:
           MANAGER_DOCKER_CMD: ${{ steps.manager-docker-command.outputs.value }}
           DEPLOYMENT_DOCKER_CMD: ${{ steps.deployment-docker-command.outputs.value }}
 
+      - name: Define translation processor command
+        id: translation-processor-command
+        shell: bash
+        run: |
+            echo "value=./gradlew -p ui/app/shared processTranslations --no-parallel" >> $GITHUB_OUTPUT
```
richturner: This shouldn't be a CI/CD build step but should be part of regular build, maybe some explanation why this should be here?

manager/src/main/java/org/openremote/manager/system/StatusResourceImpl.java
```diff
@@ -66,7 +67,8 @@ public StatusResourceImpl(Container container, List<HealthStatusProvider> health
 
         serverInfo = Map.of(
             "version", version,
-            "authServerUrl", authServerUrl
+            "authServerUrl", authServerUrl,
+            "metaItems", ValueUtil.getJsonSchemaCacheKeys()
```
richturner: Why do we need meta items here?

ui/component/or-attribute-input/src/index.ts
```diff
@@ -173,7 +173,7 @@ export const jsonFormsInputTemplateProvider: (fallback: ValueInputProvider) => V
...
..."}]},"i18n":"org.openremote.model.asset.agent.AgentLink.valueFilters"},"writeValue":{"type":"string","format":"or-multiline","i18n":"org.openremote.model.asset.agent.AgentLink.writeValue"},"writeValueConverter":{"type":"object","patternProperties":{".+":{"type":["null","number","integer","boolean","string","array","object"]}},"i18n":"org.openremote.model.asset.agent.AgentLink.writeValueConverter"},"type":{"const":"ZWaveAgentLink","default":"ZWaveAgentLink"}},"title":"Z Wave Agent Link","additionalProperties":true,"required":["type"]}},"title":"Agent Link","oneOf":[{"$ref":"#/definitions/WebsocketAgentLink"},{"$ref":"#/definitions/DefaultAgentLink"},{"$ref":"#/definitions/KNXAgentLink"},{"$ref":"#/definitions/MailAgentLink"},{"$ref":"#/definitions/BluetoothMeshAgentLink"},{"$ref":"#/definitions/SimulatorAgentLink"},{"$ref":"#/definitions/MQTTAgentLink"},{"$ref":"#/definitions/HTTPAgentLink"},{"$ref":"#/definitions/VelbusAgentLink"},{"$ref":"#/definitions/ZWaveAgentLink"},{"$ref":"#/definitions/SNMPAgentLink"},{"$ref":"#/definitions/StorageSimulatorAgentLink"},{"$ref":"#/definitions/ModbusAgentLink"}],"type":"object","properties":{"type":{"const":"AgentLink"}},"required":["type"]}');
```
richturner: Thought we'd be getting rid of the baked in schema

Ekhorn: I didn't think that would be introduced in this PR. Thought it would be part of UX improvements PR.

Added it see https://github.com/openremote/openremote/pull/2194/commits/03f5332b72f5da71fde58618e0cea31fdd444413

manager/src/main/java/org/openremote/manager/system/StatusResourceImpl.java
Ekhorn: contains hashes for caching JSON Schemas of meta items which are commonly used for attribute configuration (for configuration items)

See related changes https://github.com/openremote/openremote/pull/2194#discussion_r2443805754

.github/workflows/ci_cd.yml
Ekhorn: This is not a build step, but a check. The name of the Gradle task might seem to imply that, but how we use it (especially in CI) is different. We use it to check no changes occur when it is run, so we can ensure translations are always up-to-date.

Ekhorn: I can rename the following task to "Check for outdated translations".

```
      - name: Run translations processor
        if: ${{ steps.translation-processor-command.outputs.value != '' }}
        run: |
          set -e

          # Run the tests
          ${{ steps.translation-processor-command.outputs.value }}

          out=$(git status --porcelain | grep "ui/app/shared/locales" || true)
          if [[ -z "$out" ]]; then
              echo "✅ No pipeline changes detected."
          else
              echo "❌ Pipeline files have changed. Please run '${{ steps.translation-processor-command.outputs.value }}'"
              echo "$out"
              exit 1
          fi
        timeout-minutes: 20
```

richturner:
@Ekhorn The asset model doesn't use FQCN's it expects simple class name to be unique and the asset model backend does all the concrete type mapping. Simple class names must be unique for this reason and the schema should just use this and always use the ValueUtil helper functions for descriptor lookup of simple class names

.github/workflows/ci_cd.yml
richturner: Yes please rename and I don't think the two separate steps are necessary and should be combined into a single one

manager/src/main/java/org/openremote/manager/asset/AssetModelResourceImpl.java
```diff
@@ -29,10 +29,17 @@
...
+
 import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import jakarta.ws.rs.core.Response;
 
 public class AssetModelResourceImpl extends ManagerWebResource implements AssetModelResource {
 
+    private static final Logger LOG = Logger.getLogger(AssetModelResourceImpl.class.getName());
```
richturner:
```suggestion
    private static final Logger LOG = SyslogCategory.getLogger(MODEL_AND_VALUES, AssetModelResourceImpl.class);
```

manager/src/main/java/org/openremote/manager/asset/AssetModelResourceImpl.java
```diff
@@ -64,4 +71,15 @@ public Map<String, ValueDescriptor<?>> getValueDescriptors(RequestParams request
...
+
+    @Override
+    public Response getValueDescriptorSchema(RequestParams requestParams, String hash, String name) {
+        JsonNode schema = assetModelService.getValueDescriptorSchema(name);
+        if (schema == null) {
+            LOG.log(Level.INFO, "Could not find value descriptor: '" + name + "'");
+            return Response.status(404).build();
+        }
+        // A 1-year immutable cache, as the responses are versioned, making invalidation largely automatic
+        return Response.ok(schema).header("Cache-Control", "public,max-age=" + 31536000 + ",immutable").build();
```
richturner: Caching can be configured with Resteasy @Cache annotation:

https://docs.redhat.com/en/documentation/jboss_enterprise_application_platform_common_criteria_certification/5/html/resteasy_reference_guide/cache_nocache_cachecontrol

manager/src/main/java/org/openremote/manager/system/StatusResourceImpl.java
richturner: The link provided comes back to here so still not clear what problem this solves and also not clear why we do this for meta items but not value descriptors, why should they be treated separately?

model/src/main/java/org/openremote/model/util/ValueUtil.java
```diff
@@ -158,8 +164,20 @@ public int compare(Class<?> c1, Class<?> c2) {
     protected static Map<String, Class<? extends AgentLink<?>>> agentTypeMap = new HashMap<>();
     protected static Map<String, MetaItemDescriptor<?>> metaItemDescriptors = new HashMap<>();
     protected static Map<String, ValueDescriptor<?>> valueDescriptors = new HashMap<>();
+    protected static Map<String, ObjectNode> jsonSchemas = new HashMap<>();
+    protected static Map<String, String> jsonSchemaCacheKeys = new HashMap<>();
```
richturner: Use a better name for these: `valueDescriptorSchemas` and `valueDescriptorSchemaHashes`

model/src/main/java/org/openremote/model/util/ValueUtil.java
```diff
@@ -640,6 +658,14 @@ public static Optional<MetaItemDescriptor<?>> getMetaItemDescriptor(String name)
         return Optional.ofNullable(metaItemDescriptors.get(name));
     }
 
+    public static Map<String, ObjectNode> getJsonSchemas() {
```
richturner: Use better method name, what schemas does this get?

model/src/main/java/org/openremote/model/util/ValueUtil.java
```diff
@@ -919,12 +945,37 @@ static void doInitialise() throws IllegalStateException {
...
     protected static void doSchemaInit() {
         generator = new SchemaGenerator(JSONSchemaUtil.getJsonSchemaConfig(JSON));
     }
 
+    public static JsonNode getValueDescriptorSchema(String name) {
+        if (ValueUtil.getValueDescriptor(name).isEmpty()) {
+            return null;
+        }
+        ValueDescriptor<?> vd = ValueUtil.getValueDescriptor(name).get();
+        return jsonSchemas.computeIfAbsent(vd.getType().getSimpleName(), key -> (ObjectNode)ValueUtil.getSchema(vd.getType()));
```
richturner: schemaHash map isn't updated as well so they are out of sync

Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2194/commits/fb8660787e4fcb7b92c4b9adb53d32078cd453ed

model/src/main/java/org/openremote/model/util/ValueUtil.java
Ekhorn: Addressed in https://github.com/openremote/openremote/commit/fb8660787e4fcb7b92c4b9adb53d32078cd453ed (removed unused method)

model/src/main/java/org/openremote/model/util/ValueUtil.java
Ekhorn: Addressed in https://github.com/openremote/openremote/commit/fb8660787e4fcb7b92c4b9adb53d32078cd453ed and https://github.com/openremote/openremote/pull/2194/commits/2de6fe73e28fc14c9000589203d5e9794b15d07f (and https://github.com/openremote/openremote/pull/2194/commits/a99fa09f5bf5e11fc721dff07d3358796359e0ae)

manager/src/main/java/org/openremote/manager/asset/AssetModelResourceImpl.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2194/commits/da2f542940f60967e0a399cd1d5f5000d9cbf53b

Ekhorn: `immutable` cannot be specified through this annotation, though we don't need it allowing https://github.com/openremote/openremote/pull/2194/commits/093133ab4fedd3987cd939f9053c82000f94ae90

manager/src/main/java/org/openremote/manager/asset/AssetModelResourceImpl.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2194/commits/6a861e2748a896ac4b40189cc521fd9f6b900945

.github/workflows/ci_cd.yml
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2194/commits/94ba07659afe41283e3825bab9c4d51be6fd465f

manager/src/main/java/org/openremote/manager/system/StatusResourceImpl.java
Ekhorn: Addressed in https://github.com/openremote/openremote/pull/2194/commits/fb8660787e4fcb7b92c4b9adb53d32078cd453ed

ui/app/shared/build.gradle
```diff
@@ -84,6 +87,38 @@ tasks.register('installDist', Copy) {
...
+            "-Ashared=${project.getProjectDir()}".toString()
+        )
+
+        // Run javac with processor
+        def task = compiler.getTask(null, fileManager, null, options, null, units)
+        def success = task.call()
+        println "Processor success: $success"
+    }
+}
+
```
richturner: This doesn't feel like the `gradle` declarative way of achieving this, could the sourceSet outputs be used see this for example and other source set outputs:

```
sourceSets.main.allJava
```

manager/src/main/java/org/openremote/manager/asset/AssetModelResourceImpl.java
```diff
@@ -73,13 +75,12 @@ public Map<String, MetaItemDescriptor<?>> getMetaItemDescriptors(RequestParams r
     }
 
     @Override
-    public Response getValueDescriptorSchema(RequestParams requestParams, String hash, String name) {
+    public JsonNode getValueDescriptorSchema(RequestParams requestParams, String name, String hash) {
         JsonNode schema = assetModelService.getValueDescriptorSchema(name);
         if (schema == null) {
             LOG.log(Level.INFO, "Could not find value descriptor: '" + name + "'");
```
richturner: Maybe drop the level of this as I don't know what value it adds cluttering up the manager logs page

Ekhorn: Just removed it, indeed no real purpose just got carried over from previous code. Also the name of the descriptor should be logged as part of the incoming request path.

Addressed in https://github.com/openremote/openremote/pull/2194/commits/994d53b23a160df38149952bc02ceefdf4750869

ui/app/shared/build.gradle
Ekhorn: I tried to use a more declarative way to define the compiler options through Gradle now in https://github.com/openremote/openremote/pull/2194/commits/2d215b734f67c9bd8c1c5852bdb761667ec14220.

Though I can imagine the weird unused `destinationDirectory` and the forced cache invalidation, i.e. not integrating with the Gradle inputs/outputs caching mechanism might be something to improve (not sure how though).

PR #2198, opened 2025-10-18
Gateways: Move gateway tunnel auto-close timer handling to Edge Gateway, Tunneling Integration Test

### LINKED ISSUE #1809: Gateways: Edge gateway should handle automatic closure of tunnel after timer expiration
## User Story
Currently the central manager handles tunnel expiration timers and the closure of the tunnels but this should be handled by the edge gateway for robustness.

pankalog:
## Description

This PR modifies the current Gateway Tunnelling logic and moves the responsibility of auto-closing tunnels after the relevant timer expires to the Edge Gateway, for better robustness.

Closes #1809 .

Also, this PR includes a comprehensive test of gateway tunnelling, but it is `@Ignore`d, because of the requirement of a properly setup central instance to exist. The relevant documentation to create the testing environment for edge gateway tunneling development is added to the `Gateway Tunneling Setup` documentation page, including a Git diff that shows the needed changes to the docker compose profiles that need to be ran, [here's the PR for it](https://github.com/openremote/documentation/pull/110).
 
## Checklist

### DIFF
--- a/manager/src/main/java/org/openremote/manager/gateway/GatewayClientConnector.java
+++ b/manager/src/main/java/org/openremote/manager/gateway/GatewayClientConnector.java
@@ -140,7 +140,6 @@ public class GatewayClientConnector implements AutoCloseable {
                     this.activeTunnelSessions.forEach(GatewayTunnelSession::disconnect);
                     this.activeTunnelSessions.clear();
                 }
-
             } catch (Exception e) {
                 LOG.log(Level.WARNING, "An exception occurred whilst trying to disconnect the gateway IO client", e);
             } finally {
@@ -240,7 +239,7 @@ public class GatewayClientConnector implements AutoCloseable {
         SharedEvent event = messageFromString(message, SharedEvent.MESSAGE_PREFIX, SharedEvent.class);
 
         if (event == null) {
-            LOG.finer("Received empty message from central manager: realm=" + connection.getLocalRealm());
+            LOG.warning("Received empty message from central manager: realm=" + connection);
             return;
         }
 
@@ -277,7 +276,7 @@ public class GatewayClientConnector implements AutoCloseable {
             // The central manager sends N number of these to synchronise gateway descendant assets on the central manager
             case ReadAssetsEvent readAssets -> {
                 if (gatewayAPIVersion == null) {
-                    LOG.finer("Pre version 1.0.0 central manager read assets request so closing all active tunnels: " + connection);
+                    LOG.info("Pre version 1.0.0 central manager read assets request so closing all active tunnels: " + connection);
                     stopAllGatewayTunnels();
                 }
 
@@ -297,7 +296,7 @@ public class GatewayClientConnector implements AutoCloseable {
                 if (gatewayAPIVersion == null) {
                     // This is a legacy version of the openremote on the central instance so we won't get a GatewayInitialisedEvent
                     // so we mark this client as connected now
-                    LOG.fine("Central manager running an older version so assuming connection is initialised");
+                    LOG.info("Central manager running an older version so assuming connection is initialised");
                     initFuture.complete(null);
                 }
 
@@ -373,13 +372,7 @@ public class GatewayClientConnector implements AutoCloseable {
 
         Container.EXECUTOR.submit(() -> {
 
-            if (activeTunnels == null || activeTunnels.length == 0) {
-                LOG.fine("No gateway tunnel sessions on central instance so stopping any active sessions: " + connection);
-                stopAllGatewayTunnels();
-                return;
-            }
-
-            LOG.fine("Synchronising gateway tunnel sessions with central instance: " + connection);
+            LOG.info("Synchronising gateway tunnel sessions with central instance: " + connection);
 
             // Synchronise the SSH sessions with what the central instance thinks are open
             // Filter out any tunnels expiring in the next 5 seconds
@@ -391,6 +384,12 @@ public class GatewayClientConnector implements AutoCloseable {
                 return true;
             }).toList();
 
+            if (tunnels.isEmpty()) {
+                LOG.finer("No gateway tunnel sessions on central instance so stopping any active sessions: " + connection);
+                stopAllGatewayTunnels();
+                return;
+            }
+
             // Create any new tunnels currently not established
             tunnels.forEach(activeTunnel -> {
                 GatewayTunnelSession activeSession = this.activeTunnelSessions.stream()
@@ -401,7 +400,7 @@ public class GatewayClientConnector implements AutoCloseable {
                     LOG.finer("Active tunnel session found for tunnel: " + activeTunnel);
                 } else {
                     // No existing session for this active tunnel
-                    LOG.fine("Active tunnel session not found for tunnel so starting: " + activeTunnel);
+                    LOG.finer("Active tunnel session not found for tunnel so starting: " + activeTunnel);
                     startGatewayTunnel(activeTunnel);
                 }
             });
@@ -410,7 +409,7 @@ public class GatewayClientConnector implements AutoCloseable {
             this.activeTunnelSessions.removeIf(activeTunnelSession -> {
                 boolean obsolete = tunnels.stream().noneMatch(tunnel -> tunnel.equals(activeTunnelSession.getTunnelInfo()));
                 if (obsolete) {
-                    LOG.fine("Removing obsolete tunnel session: " + activeTunnelSession);
+                    LOG.finer("Removing obsolete tunnel session: " + activeTunnelSession);
                     activeTunnelSession.disconnect();
                 }
                 return obsolete;
@@ -444,10 +443,11 @@ public class GatewayClientConnector implements AutoCloseable {
 
     protected void sendCentralManagerMessage(SharedEvent event) {
         if (client != null) {
-            if (LOG.isLoggable(Level.FINEST)) {
-                LOG.finest("Sending message to central manager: realm=" + connection.getLocalRealm() + ", " + event);
-            } else {
-                LOG.finer(() -> "Sending message to central manager: realm=" + connection.getLocalRealm() + ", " + event.getClass().getSimpleName());
+            boolean isAttributeEvent = event instanceof AttributeEvent;
+            if (isAttributeEvent && LOG.isLoggable(Level.FINEST)) {
+                LOG.finest("Sending attribute event to central manager: realm=" + connection.getLocalRealm() + ", " + event);
+            } else if (!isAttributeEvent) {
+                LOG.fine(() -> "Sending message to central manager: realm=" + connection.getLocalRealm() + ", " + event);
             }
             client.sendMessage(messageToString(SharedEvent.MESSAGE_PREFIX, event));
         }
@@ -704,7 +704,7 @@ public class GatewayClientConnector implements AutoCloseable {
             if (allowEvent && connection.getAssetSyncRules() != null) {
                 GatewayAssetSyncRule syncRule = connection.getAssetSyncRules().getOrDefault(ev.getAssetType(), connection.getAssetSyncRules().get("*"));
                 if (syncRule != null && syncRule.excludeAttributes != null && syncRule.excludeAttributes.contains(ev.getName())) {
-                    LOG.finer(() -> "Attribute event excluded due to sync rule: " + ev);
+                    LOG.finest(() -> "Attribute event excluded due to sync rule: " + ev);
                     allowEvent = false;
                 }
             }


model/src/main/java/org/openremote/model/gateway/GatewayCapabilitiesResponseEvent.java
```diff
@@ -9,19 +9,42 @@
...
+    protected final boolean tunnelTimeoutManagementSupported;
+    protected final String version;
 
     @JsonCreator
-    public GatewayCapabilitiesResponseEvent(@JsonProperty("timestamp") Date timestamp, @JsonProperty("tunnelingSupported") boolean tunnelingSupported) {
+    public GatewayCapabilitiesResponseEvent(
+        @JsonProperty("timestamp") Date timestamp,
+        @JsonProperty("tunnelingSupported") boolean tunnelingSupported,
+        @JsonProperty("version") String version,
+        @JsonProperty("tunnelTimeoutManagementSupported") boolean tunnelTimeoutManagementSupported) {
```
richturner: @JsonProperty annotations we compile with method arg names so jackson can hydrate without these annotations

model/src/main/java/org/openremote/model/gateway/GatewayCapabilitiesResponseEvent.java
```diff
@@ -9,19 +9,42 @@
 public class GatewayCapabilitiesResponseEvent extends SharedEvent {
 
     public static final String TYPE = "gateway-capabilities-response";
+    public static final String CURRENT_VERSION = "1.0.0";
```
richturner: Maybe we should just use the manager version string rather than introduce another version number

test/src/test/groovy/org/openremote/test/gateway/GatewayTest.groovy
```diff
@@ -1080,33 +1096,60 @@ class GatewayTest extends Specification implements ManagerContainerTrait {
     @Ignore
     def "Verify gateway tunnel factory"() {
         given: "an ssh private key and the URL of a manager instance with tunnelling configured"
-        def keyPath = Paths.get(System.getProperty("user.home"), ".ssh", "test_key")
-        def tunnelSSHHost = "test.openremote.app"
+        def keyPath = Paths.get("/Users/panos/IdeaProjects/openremote/deployment/sish/client/client")
+        def tunnelSSHHost = "localhost"
```
richturner: This needs fixing, we can include test keys in the test resources (like the provisioning keys that are already inlcuded for provisioning tests).

test/src/test/groovy/org/openremote/test/gateway/GatewayTest.groovy
```diff
@@ -1080,33 +1096,60 @@ class GatewayTest extends Specification implements ManagerContainerTrait {
...
         def tunnelSSHPort = 2222
 
         and: "the container environment is started"
-        def conditions = new PollingConditions(timeout: 15, delay: 0.2)
-        def container = startContainer(defaultConfig() << [(GatewayService.OR_GATEWAY_TUNNEL_SSH_KEY_FILE): keyPath.toAbsolutePath().toString()], defaultServices())
+        def conditions = new PollingConditions(timeout: 3, delay: 0.5)
+        def config = defaultConfig();
+        def container = startContainer(config << [(GatewayService.OR_GATEWAY_TUNNEL_SSH_KEY_FILE): keyPath.toAbsolutePath().toString()], defaultServices())
```
richturner: Is this change necessary

test/src/main/groovy/org/openremote/test/ContainerTrait.groovy
```diff
@@ -480,6 +488,45 @@ trait ContainerTrait {
...
     }
+    UriBuilder serverUri(boolean secure, String serverHost, int serverPort) {
+        UriBuilder.fromUri("")
+                .scheme(secure ? "https" : "http").host(serverHost).port(serverPort)
+    }
+
+    /**
+     * Makes a call to a remote OpenRemote manager to retrieve an access token for a given user
+     * */
+    AccessTokenResponse authenticate(boolean secure,
```
richturner: Move this to where existing `authenticate` methods are

manager/src/main/java/org/openremote/manager/gateway/GatewayClientService.java
```diff
@@ -638,4 +651,65 @@ protected void applySyncRuleToMeta(String attributeName, MetaMap meta, GatewayAs
...
+            return;
+        }
+
+        LOG.info("Automatically closing tunnel due to timeout: " + tunnelId);
+
+        try {
+            if (gatewayTunnelFactory != null) {
+                gatewayTunnelFactory.stopTunnel(tunnelInfo);
+            }
+            cancelAutoCloseTunnel(tunnelId);
```
richturner: Put this in finally so cleanup is done

model/src/main/java/org/openremote/model/gateway/GatewayCapabilitiesResponseEvent.java
pankalog: I agree, but to do this, I introduced some static methods in `StatusResourceImpl` to allow other parts of the manager to retrieve the version number of the currently running manager.

manager/src/main/java/org/openremote/manager/system/StatusResourceImpl.java
```diff
@@ -90,4 +78,22 @@ public Map<String, Object> getHealthStatus() {
...
+                Properties versionProps = new Properties();
+                versionProps.load(resourceStream);
+                return versionProps.getProperty("version");
+            }else {
+                throw new RuntimeException("Could not load version.properties");
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
richturner: Lets move this into its' own static class called `VersionInfo` and then `StatusResourceImpl()` can look it up there and GatewayClientService can use this new class also

manager/src/main/java/org/openremote/manager/gateway/GatewayConnector.java
```diff
@@ -258,11 +259,12 @@ protected void getCapabilities() {
                     }
                 }
                 tunnellingSupported = response != null && response.isTunnelingSupported();
-                this.gatewayVersion = response.getVersion();
+                this.gatewayVersion = Objects.requireNonNull(response).getVersion();
```
richturner: We're now throwing an exception here which is different behaviour, just check if `response != null` which is needed for older gateways that don't support the capabilities request

manager/src/main/java/org/openremote/manager/gateway/GatewayConnector.java
```diff
@@ -258,11 +259,12 @@ protected void getCapabilities() {
...
                 this.tunnelTimeoutManagementSupported = response.isTunnelTimeoutManagementSupported();
 
-                if(!Objects.equals(getGatewayVersion(), GatewayCapabilitiesResponseEvent.CURRENT_VERSION)){
-                    LOG.warning("Gateway Version is not up to date, latest version is "+GatewayCapabilitiesResponseEvent.CURRENT_VERSION + ", Gateway version is " + getGatewayVersion() + ". GatewayId: " + gatewayId);
+                String currentManagerVersion = StatusResourceImpl.getManagerVersion();
+                if(!Objects.equals(getGatewayVersion(), currentManagerVersion)) {
```
richturner: This logic is not correct we might increase the manager version but have no changes to the Gateway API so we instead need to keep a latest compatible version field to track this.

model/src/main/java/org/openremote/model/gateway/GatewayCapabilitiesResponseEvent.java
```diff
@@ -10,18 +9,32 @@ public class GatewayCapabilitiesResponseEvent extends SharedEvent {
...
         super(timestamp != null ? timestamp.getTime() : new Date().getTime());
         this.tunnelingSupported = tunnelingSupported;
+        this.version = version != null ? version : "0.0.1";
+        this.tunnelTimeoutManagementSupported = tunnelTimeoutManagementSupported;
     }
 
-    public GatewayCapabilitiesResponseEvent(final boolean tunnelingSupported) {
+    public GatewayCapabilitiesResponseEvent(final boolean tunnelingSupported, final String version) {
         this.tunnelingSupported = tunnelingSupported;
+        this.tunnelTimeoutManagementSupported = version != null;
```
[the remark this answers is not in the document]
pankalog: This is fine, because versioning the tunnel is implemented in this PR, the same PR that implements tunnel self-managed timeout

model/src/main/java/org/openremote/model/gateway/GatewayCapabilitiesResponseEvent.java
```diff
@@ -10,18 +9,32 @@ public class GatewayCapabilitiesResponseEvent extends SharedEvent {
...
     protected final boolean tunnelingSupported;
+    protected final boolean tunnelTimeoutManagementSupported;
+    protected final String version;
 
     @JsonCreator
-    public GatewayCapabilitiesResponseEvent(@JsonProperty("timestamp") Date timestamp, @JsonProperty("tunnelingSupported") boolean tunnelingSupported) {
+    public GatewayCapabilitiesResponseEvent(Date timestamp, boolean tunnelingSupported, String version, boolean tunnelTimeoutManagementSupported) {
         super(timestamp != null ? timestamp.getTime() : new Date().getTime());
         this.tunnelingSupported = tunnelingSupported;
+        this.version = version != null ? version : "0.0.1";
```
richturner: Don't like this, why fallback to some arbitrary version `0.0.1` leave as null if the gateway doesn't know its' version (i.e. is on a version pre introduction of version in the response)

manager/src/main/java/org/openremote/manager/gateway/GatewayConnector.java
```diff
@@ -257,6 +259,13 @@ protected void getCapabilities() {
                     }
                 }
                 tunnellingSupported = response != null && response.isTunnelingSupported();
+                this.gatewayVersion = response != null ? response.getVersion() : "";
```
richturner: Use null not empty strings

richturner (commented):
Hey @pankalog I'm sorry to do this but after thinking about this a bit more I don't think using the manager version is the best approach as we'll need to store a compatibility matrix which quickly gets difficult to manage.

I think we go back to the approach you had initially with the dual versioning i.e. we version the gateway API, it is useful to use an adapter approach where incoming data is adapted to the latest version and then the logic only has to deal with the latest version, we mostly have this pattern so far but useful for future versions.

pankalog:
Hey @richturner, as you said I reverted back to the logic of having a gateway API versioning system instead of using the manager versions. I also made sure that all comments are either addressed or replied to. Thanks for taking a look!

model/src/main/java/org/openremote/model/gateway/GatewayCapabilitiesResponseEvent.java
```diff
@@ -16,14 +17,18 @@ public class GatewayCapabilitiesResponseEvent extends SharedEvent {
...
+    public GatewayCapabilitiesResponseEvent(final boolean tunnelingSupported) {
+        this(tunnelingSupported, false);
+    }
+
+    public GatewayCapabilitiesResponseEvent(final boolean tunnelingSupported, final boolean tunnelTimeoutManagementSupported) {
         this.tunnelingSupported = tunnelingSupported;
-        this.tunnelTimeoutManagementSupported = version != null;
-        this.version = version;
+        this.tunnelTimeoutManagementSupported = tunnelTimeoutManagementSupported;
+        this.version = tunnelTimeoutManagementSupported ? CURRENT_VERSION : "0.0.1";
```
richturner: Leave version as null if it is not set in JSON payload, `0.0.1` means as much as null but occupies memory and you'll then be string matching it

manager/src/main/java/org/openremote/manager/gateway/GatewayConnector.java
```diff
@@ -259,12 +259,12 @@ protected void getCapabilities() {
                     }
                 }
                 tunnellingSupported = response != null && response.isTunnelingSupported();
-                this.gatewayVersion = response != null ? response.getVersion() : "";
+                this.gatewayVersion = response != null ? response.getVersion() : null;
                 this.tunnelTimeoutManagementSupported = response != null ? response.isTunnelTimeoutManagementSupported() : false;
 
                 String currentManagerVersion = VersionInfo.getManagerVersion();
```
richturner: We aren't comparing manager versions we need to compare Gateway API versions so use VersionInfo.GatewayAPIVersion

model/src/main/java/org/openremote/model/gateway/GatewayCapabilitiesResponseEvent.java
```diff
@@ -8,6 +8,7 @@
 public class GatewayCapabilitiesResponseEvent extends SharedEvent {
 
     public static final String TYPE = "gateway-capabilities-response";
+    public static final String CURRENT_VERSION = "1.0.0";
```
richturner: Move to VersionInfo as GatewayAPIVersion

manager/src/main/java/org/openremote/manager/gateway/GatewayConnector.java
```diff
@@ -262,9 +262,9 @@ protected void getCapabilities() {
                 this.gatewayVersion = response != null ? response.getVersion() : null;
                 this.tunnelTimeoutManagementSupported = response != null ? response.isTunnelTimeoutManagementSupported() : false;
 
-                String currentManagerVersion = VersionInfo.getManagerVersion();
-                if(!Objects.equals(getGatewayVersion(), currentManagerVersion)) {
-                    LOG.warning("Gateway Version is not up to date, latest version is "+currentManagerVersion + ", Gateway version is " + getGatewayVersion() + ". GatewayId: " + gatewayId);
+                String currentGatewayApiVersion = VersionInfo.getGatewayApiVersion();
+                if(this.getGatewayVersion() != null && VersionInfo.isVersionLess(this.getGatewayVersion(), currentGatewayApiVersion)) {
```
richturner: What if gateway version is newer than the central instance?

pankalog: I can add another `LOG.warning` if the edge gateway's version is newer than the central instance, but at least for the tunnel time-out management feature, this means that the timeout is going to be managed by the edge gateway as normal. 

If this code is being ran, it means that the central instance has the gateway API versioning and tunnel auto-close support, and the gateway also has that support, since the version is greater than the central instance.

pankalog: @richturner just added this logging in the new commit

richturner: My comment is a more general point about future version changes it isn't related to the one feature that has been added at the same time versioning has been introduced.

I don't see why we need multiple if statements a single if versions using `not equal` and a single log statement like:

"Gateway API version does not match central manager: gatewayVersion=X"

richturner (changes requested):
Hey Panos, Happy new year!
Almost there but a small request.
Can you add the manager version to the `GatewayCapabilitiesRequestEvent` and then in `GatewayClientService` also log when there is a version mismatch so the owner of the gateway knows there is a version difference

MartinaeyNL:
As other fixes related to the gateway tunnels have been merged, it makes sense to revisit this PR.
Requested a new review from the OpenRemote developers group.

PR #2204, opened 2025-10-22
Improvements to simulator agent

### LINKED ISSUE #2147: SimulatorAgentLink: Schedule without start date fails to save
Adding the schedule property to a SimulatorAgentLink, if no start date is set, saving fails with an NPE.
If should be possible to define a schedule without a start date, which means start now.
Reference is today 00:00 or today beginning of current hour for an hourly recurrence.

### LINKED ISSUE #2164: SimulatorAgentLink: End date
When adding parameters to schedule field of a simulator agent link, there is the possibility to select "End" in the list, but with a message saying "Not implemented, within the recurrence rule you can specify an end date."

I understand the idea to guide the user, but I can still add that parameter and a field will be display, where I can enter a value. This is confusing.

### LINKED ISSUE #2165: SimulatorAgent: start date inconsistencies
When setting the start on a schedule on a SimulatorAgentLink, there is a time component in it.
This was excluded from the spec, was this changed after discussing with @pierrekil ?

But if it was, the description still says "Set a start date, if not provided, starts immediately. When the replay datapoint timestamp is 0 it will insert it at 00:00.", which is not correct, timestamp 0 corresponds to the time in the start property.

Timestamp is only an offset from 00:00 if there is no schedule property at all. In that case, it's 00:00 UTC. But if there is a schedule and a date, it's offset from the displayed time, in the local time zone, not UTC.

### LINKED ISSUE #2171: SimulatorAgent: issues when using UNTIL in Schedule recurrence
Having the following replay data
```
[
  {
    "timestamp": 60,
    "value": 60
  },
  {
    "timestamp": 120,
    "value": 20
  },
  {
    "timestamp": 240,
    "value": 40
  }
]
```

I setup a scheduling starting at 9:21, using FREQ=HOURLY;UNTIL=20250929T083000Z as recurrence
This is 8:30 UTC, should be 10:30 CEST
But shows predicted data points until 11:25:09
After first period passed, it showed data points until 12:25:09
-> It seems to apply the timezone difference in the wrong direction, calculating 12:30 instead of 8:30

I did setup an line graph to explore what was going on and it shows very strange things:
- still showing the predicted points that are in the past (might be an insights bug, not sure)
- showing frequent value changes with the same attribute value

Image
Image
Image

After that, the UI behaves in very strange way. I expanded the replay data point but when I click on the JSON button, it does not show JSON but closes the data points displayed. Then the UI is frozen.

https://github.com/user-attachments/assets/b1e17f74-42d0-44bc-adab-a3236c3e72ff

### LINKED ISSUE #2248: SimulatorAgent: inconsistencies setting schedule on agent link
When I add a schedule property, it starts by using an "empty" period at the current time for the single occurrence plan.
Image

Seems a bit weird to me but OK
I then click on the button to edit -> OK
Image

I switch to all day, the date is correct and the range also (although I'm not sure if things would are 100% OK if there's a data point in [23:59:00-24:00:00[ range.
Image

I then un-check all day, and instead of going back to the values that were selected, it now defaults to a one hour window at the current time.
1. I would prefer keeping my old values
2. on creation the default is an empty period, here it's a one hour period -> discrepancy is confusing
Image

As this is not what I wanted, I click on cancel
Image
but the change has already been saved anyway

### LINKED ISSUE #2249: SimulatorAgent: disappearing schedule property
I added a schedule property to an agent link.
By default, it's a "Plan an occurrence" type.
If I select "Every 24hrs" for the type, I can see that information displayed "behind" the modal dialog on the agent link.
Image

But as soon as I click on apply, that information disappears and the "schedule" parameter is gone.
Image

As this is the default behaviour if no schedule is set, I would guess it works OK and I can understand the technical implementation behind (haven't looked at the code though).

This is confusing for an end user. If we're using the absence of a schedule property on the agent link to mean default, then maybe display the message "Every 24hrs at 00:00 UTC" always on the agent link would be more consistent.

@pierrekil @remyberden was this discussed ? any input ?

### LINKED ISSUE #2251: SimulatorAgent: missing predictions
I have an attribute with the "has predicted data points" meta set.
In the agent link, I use the "Plan an occurence" default type and set a one hour period (tried both with and without overlap with current time).
I used replay data such as
```
[
  {
    "timestamp": 60,
    "value": 60
  },
  {
    "timestamp": 120,
    "value": 20
  },
  {
    "timestamp": 240,
    "value": 40
  }
]
```
I created an Insight dashboard for that property with a chart. I could see the previous values as time passed by but I never saw the predicted data points.
I had some predicated data points with other schedule configurations (e.g. every 24h or whole day repeat).

Ekhorn:
## Description

- Closes #2147 is addressed by setting the current date as value by default, see https://github.com/openremote/openremote/blob/55de498d64219873749557a5e21adbe89ebc7424/agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java#L389
- Closes #2248 Time values are no longer reset when switching options
- Closes #2249 The default schedule now stays around by applying a recurrence rule matching the default replay loop 
- Closes #2251 ~~Resolved~~ and wrote an integration test https://github.com/openremote/openremote/blob/55de498d64219873749557a5e21adbe89ebc7424/test/src/test/groovy/org/openremote/test/protocol/simulator/SimulatorProtocolTest.groovy#L559
- Closes #2164 End date is now implemented for both single occurrence and recurring events
- Closes #2165 We discussed that we'll have a time component for the start date and it will thus be relative to that.
- Closes #2171 `UNTIL` rule part ends as specified and the `COUNT` rule part is now consistent (only issue left is if you set `COUNT` 1 and it already started it exits early).

In preparation for #2250 this PR also makes it easier for any user to configure a Simulator schedule in the UI. This also updates the rule validity scheduling component and adds additional frequency options.

## Changelog

- Start date now defaults to start of day at UTC
- Implement end date property for single occurrence and recurring events
- Resolve error when current time is between the start date and the first occurrence when using BYXXX recurrence rule parts
- Predicted datapoints now stop at the recurrence rule cutoff time
- Predicted datapoint values previously added once surpassed are now removed
- The schedule property for the Simulator has an improved UI

## Checklist

### DIFF
--- a/ui/app/manager/test/fixtures/data/rules.ts
+++ b/ui/app/manager/test/fixtures/data/rules.ts
@@ -1,8 +1,8 @@
 export const energyRule = {
   name: "Energy rule",
-  attribute_type: "Thing asset",
+  asset_type: "Thing asset",
   asset: "Battery",
   attribute_when: "Energy level",
   attribute_then: "Energy capacity",
-  value: 50,
+  value: 50
 };
--- a/ui/app/manager/test/rules.test.ts
+++ b/ui/app/manager/test/rules.test.ts
@@ -1,63 +1,127 @@
-import { expect } from "@openremote/test";
-import { test, userStatePath } from "./fixtures/manager.js";
-import { preparedAssetsForRules as assets } from "./fixtures/data/assets.js";
-import { energyRule } from "./fixtures/data/rules.js";
+import {expect} from "@openremote/test";
+import {test, userStatePath} from "./fixtures/manager.js";
+import {preparedAssetsForRules as assets} from "./fixtures/data/assets.js";
+import {energyRule} from "./fixtures/data/rules.js";
+import {Asset} from "@openremote/model";
+
+test.use({storageState: userStatePath});
 
 /**
- * @given The realm "smartcity" with user "smartcity" and assets are set up
- * @and Logged into OpenRemote "smartcity" realm as "smartcity"
- * @and Navigated to the "Rules" tab
- * @and Click add new rule
+ * Simple function that generates assets based on {@link assets}
+ * @param multiplier - Amount of assets to generate per asset type
  */
-test.beforeEach(async ({ page, manager }) => {
-  await manager.setup("smartcity", { assets });
-  await manager.goToRealmStartPage("smartcity");
-  await manager.navigateToTab("Rules");
-  await page.click(".mdi-plus >> nth=0");
-});
-
-test.use({ storageState: userStatePath });
+function generateALotOfAssets(multiplier = 5): Asset[] {
+    return Array.from({length: multiplier}, (_, i) =>
+        assets.map((a: Asset) => ({...a, name: `${a.name} ${i}`}))
+    ).flat();
+}
 
 /**
- * @when Selecting When-Then rule
+ * @when Creating a When-Then rule
  * @and Naming the rule
  * @and Configuring a When condition on the asset
  * @and Configuring a Then action on the same asset
  * @and Saving the rule
  * @then The When-Then rule should appear in the rule list
  */
-test("Create a When-Then rule for an asset with a trigger and action", async ({ page, manager, shared }) => {
-  await page.locator("li").filter({ hasText: "When-Then" }).click();
-  await page.fill("text=Rule name", energyRule.name);
+test("Create a When-Then rule for an asset with a trigger and action", async ({page, manager, shared}) => {
+    await manager.setup("smartcity", {assets});
+    await manager.goToRealmStartPage("smartcity");
+    await manager.navigateToTab("Rules");
+    await page.click(".mdi-plus >> nth=0");
+    await page.locator("li").filter({hasText: "When-Then"}).click();
+    await page.fill("text=Rule name", energyRule.name);
 
-  await page.click("or-rule-when #component span");
-  await page.locator("or-rule-when li").filter({ hasText: energyRule.attribute_type }).click();
-  await page.click("text=Asset Any of this type");
-  await page.click(`#idSelect >> text=${energyRule.asset}`);
-  await page.getByRole("button", { name: "Attribute", exact: true }).click();
-  await page.click(`li[role="option"]:has-text("${energyRule.attribute_when}")`);
-  await page.getByRole("button", { name: "Operator", exact: true }).click();
-  await page.click("text=Less than or equal to");
-  await page.getByRole("spinbutton", { name: "Energy level" }).fill(energyRule.value.toString());
+    await page.click("or-rule-when #component span");
+    await page.locator("or-rule-when li").filter({hasText: energyRule.asset_type}).click();
+    await page.click("text=Asset Any of this type");
+    await page.click(`#idSelect >> text=${energyRule.asset}`);
+    await page.getByRole("button", {name: "Attribute", exact: true}).click();
+    await page.click(`li[role="option"]:has-text("${energyRule.attribute_when}")`);
+    await page.getByRole("button", {name: "Operator", exact: true}).click();
+    await page.click("text=Less than or equal to");
+    await page.getByRole("spinbutton", {name: "Energy level"}).fill(energyRule.value.toString());
 
-  await page.getByRole("button", { name: "Add action" }).click();
-  await page.click(`or-rule-then-otherwise li[role="menuitem"]:has-text("${energyRule.attribute_type}")`);
-  await page.click("text=Asset Matched");
-  await page.click(`#matchSelect li[role="option"]:has-text("${energyRule.asset}")`);
-  await page.getByRole("button", { name: "Attribute", exact: true }).click();
-  await page.click(`li[role="option"]:has-text("${energyRule.attribute_then}")`);
-  await page.getByRole("spinbutton", { name: "Value" }).fill(energyRule.value.toString());
+    await page.getByRole("button", {name: "Add action"}).click();
+    await page.click(`or-rule-then-otherwise li[role="menuitem"]:has-text("${energyRule.asset_type}")`);
+    await page.click("text=Asset Matched");
+    await page.click(`#matchSelect li[role="option"]:has-text("${energyRule.asset}")`);
+    await page.getByRole("button", {name: "Attribute", exact: true}).click();
+    await page.click(`li[role="option"]:has-text("${energyRule.attribute_then}")`);
+    await page.getByRole("spinbutton", {name: "Value"}).fill(energyRule.value.toString());
 
-  await shared.interceptResponse<number>("**/rules/realm", (rule) => {
-    if (rule) manager.rules.push(rule);
-  });
+    await shared.interceptResponse<number>("**/rules/realm", (rule) => {
+        if (rule) manager.rules.push(rule);
+    });
 
-  await page.getByRole("button", { name: "Save" }).click();
-  await expect(page.locator(`text=${energyRule.name}`)).toHaveCount(1);
+    await page.getByRole("button", {name: "Save"}).click();
+    await expect(page.locator(`text=${energyRule.name}`)).toHaveCount(1);
 });
 
 /**
- * @when Selecting Flow rule
+ * @when Creating a When-Then rule
+ * @and Naming the rule
+ * @and Configuring a When condition by searching for a specific asset
+ * @and Configuring a Then action by searching for a different asset
+ * @and Saving the rule
+ * @then The When-Then rule should appear in the rule list
+ */
+test("Create a When-Then rule by searching for an asset", async ({page, manager, shared}) => {
+    const multiplier = 100;
+    const aLotOfAssets = generateALotOfAssets(multiplier);
+    await manager.setup("smartcity", {assets: aLotOfAssets});
+    await manager.goToRealmStartPage("smartcity");
+    await manager.navigateToTab("Rules");
+
+    // Make sure the correct amount assets are set up, and the variables for this rule are adjusted
+    expect(manager.assets.length).toBe(multiplier * assets.length);
+    const firstAssetName = [...manager.assets].find(a => a.name?.includes(energyRule.asset))?.name ?? "";
+    const lastAssetName = [...manager.assets].reverse().find(a => a?.name?.includes(energyRule.asset))?.name ?? "";
+    expect(firstAssetName).toContain(energyRule.asset);
+    expect(lastAssetName).toContain(energyRule.asset);
+
+    await page.click(".mdi-plus >> nth=0");
+    await page.locator("li").filter({hasText: "When-Then"}).click();
+    await page.fill("text=Rule name", energyRule.name);
+
+    // Select asset type of the When clause, search for the last asset in the list, and select the attribute
+    await page.click("or-rule-when #component span");
+    await page.locator("or-rule-when li").filter({hasText: energyRule.asset_type}).click();
+    await page.click(`text=Asset Any of this type`);
+    await expect(page.getByRole('textbox', { name: 'Search'})).toBeVisible();
+    await page.getByRole('textbox', {name: 'Search'}).click();
+    await page.keyboard.type(lastAssetName);
+    await expect(page.locator("or-rule-asset-query or-mwc-input li[role=option]").getByText(lastAssetName, { exact: true })).toBeVisible();
+    await page.locator("or-rule-asset-query or-mwc-input li[role=option]").filter({hasText: lastAssetName}).click();
+    await page.getByRole("button", {name: "Attribute", exact: true}).click();
+    await page.click(`or-rule-asset-query or-mwc-input li[role="option"]:has-text("${energyRule.attribute_when}")`);
+    await page.getByRole("button", {name: "Operator", exact: true}).click();
+    await page.click("text=Less than or equal to");
+    await page.getByRole("spinbutton", {name: "Energy level"}).fill(energyRule.value.toString());
+
+    // Configure Then clause
+    await page.getByRole("button", {name: "Add action"}).click();
+    await page.click(`or-rule-then-otherwise li[role="menuitem"]:has-text("${energyRule.asset_type}")`);
+    await page.click("text=Matched");
+    await expect(page.getByRole('textbox', { name: 'Search'})).toBeVisible();
+    await page.getByRole('textbox', {name: 'Search'}).click();
+    await page.keyboard.type(firstAssetName);
+    await expect(page.locator("or-rule-action-attribute or-mwc-input li[role=option]").getByText(firstAssetName, { exact: true })).toBeVisible();
+    await page.locator("or-rule-action-attribute or-mwc-input li[role=option]").filter({hasText: firstAssetName}).click();
+    await page.getByRole("button", {name: "Attribute", exact: true}).click();
+    await page.click(`or-rule-action-attribute or-mwc-input li[role=option]:has-text("${energyRule.attribute_then}")`);
+    await page.getByRole("spinbutton", {name: "Value"}).fill(energyRule.value.toString());
+
+    await shared.interceptResponse<number>("**/rules/realm", (rule) => {
+        if (rule) manager.rules.push(rule);
+    });
+
+    await page.getByRole("button", {name: "Save"}).click();
+    await expect(page.locator(`text=${energyRule.name}`)).toHaveCount(1);
+});
+
+/**
+ * @when Creating a Flow rule
  * @and Naming the rule
  * @and Dragging elements onto the canvas
  * @and Assigning attributes and values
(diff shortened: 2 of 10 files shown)


ui/component/or-scheduler/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## License
+[GNU AGPL](https://www.gnu.org/licenses/agpl-3.0.en.html)
+
+[npm-image]: https://img.shields.io/npm/v/live-xxx.svg
+[npm-url]: https://npmjs.org/package/@openremote/or-calendar-event
+[travis-image]: https://img.shields.io/travis/live-js/live-xxx/master.svg
+[travis-url]: https://travis-ci.org/live-js/live-xxx
+[coveralls-image]: https://img.shields.io/coveralls/live-js/live-xxx/master.svg
+[coveralls-url]: https://coveralls.io/r/live-js/live-xxx?branch=master
```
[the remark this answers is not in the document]
Ekhorn: Wrong url

ui/component/or-calendar-event/src/index.ts
```diff
@@ -0,0 +1,518 @@
...
+import {customElement, property, query} from "lit/decorators.js";
+import {CalendarEvent} from "@openremote/model";
+import {InputType, OrInputChangedEvent} from "@openremote/or-mwc-components/or-mwc-input";
+import {translate,i18next} from "@openremote/or-translate";
+
+import {OrMwcDialog, showDialog} from "@openremote/or-mwc-components/or-mwc-dialog";
+import {ByWeekday, Frequency, Options, RRule, Weekday, WeekdayStr} from 'rrule'
+import moment from "moment";
+import { Days } from "rrule/dist/esm/rrule";
+import { getMonthDays, isLeapYear, MONTH_DAYS } from "rrule/dist/esm/dateutil";
```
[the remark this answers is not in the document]
Ekhorn: Addressed locally

ui/component/or-scheduler/rspack.config.js
```diff
@@ -0,0 +1,16 @@
...
+      "index": {
+          vendor: {
+              "moment": "moment"
+          },
+          excludeOr: true
+      },
+      "index.bundle": {
+          excludeOr: true,
+      },
+    "index.bundle": undefined
```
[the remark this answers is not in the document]
Ekhorn: standard

ui/component/or-attribute-input/src/scheduler-renderer.ts
```diff
@@ -0,0 +1,66 @@
...
+        ...mapStateToControlProps({jsonforms: {...state}}, props),
+        ...mapDispatchToControlProps(state.dispatch)
+    };
+
+    const onCalendarEventChanged = (event: OrCalendarEventChangedEvent | undefined) => {
+        console.log(event, props.data)
+        props.handleChange("schedule", event?.detail.value);
+    };
+
+    return getTemplateWrapper(html`
```
Ekhorn, line 94: Add delete button for calendar renderer

ui/component/or-scheduler/src/index.ts
```diff
@@ -0,0 +1,505 @@
...
+            </div>`;
+    }
+
+    protected getByRulePart<T>(part: RulePartKey, type: InputType, options: T[]): TemplateResult | undefined {
+        if (!this._parts?.includes(part)) {
+          return undefined
+        }
+
+        const value = part === "byweekday"
+            ? this._rrule?.options?.byweekday?.map(d => new Weekday(d).toString())
```
Ekhorn: The current week day is somehow resolved and automatically applied while it shouldn't get applied by default.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn, line 1: Move some checks in `calculatePredictedDatapoints` to dedicated method for clarity

ui/component/or-rules/src/or-rule-viewer.ts
Ekhorn, line 1: Also consider whether certain frequencies should be disabled. Like `SECONDLY` which shows 

Image

ui/component/or-scheduler/src/index.ts
Ekhorn: Also on the rules pages at least the weekdays don't show only after switching frequency.

ui/component/or-rules/src/or-rule-viewer.ts
Ekhorn: You can't control at what seconds it starts without bysecond anyways.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
```diff
@@ -19,26 +19,29 @@
...
-        " This could be once when only a start- (and end) date are defined," +
-        " or a recurring event following the RFC 5545 RRULE format." +
+    @JsonSchemaTitle("Schedule")
+    @JsonSchemaDescription("Overwrites the possible dataset length and when it is replayed." +
+        " The dataset can be scheduled to stop based on the UNTIL rule part" +
+        " or set to recur by a certain amount using the COUNT rule part. The schedule also allows adjusting the frequency" +
+        " at which the dataset is replayed and at what times following the RFC 5545 RRULE format." +
         " If not provided defaults to 24 hours. If the replay data contains datapoints scheduled after the" +
         " default 24 hours or the recurrence rule the datapoints will be ignored.")
+    @JsonSchemaFormat("or-scheduler")
```
MartinaeyNL: I think we should not have any frontend references in the backend here.
We need to think of something else here.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -251,56 +286,106 @@ public List<ValueDatapoint<?>> calculatePredictedDatapoints(
...
-
-        for (SimulatorReplayDatapoint d : simulatorReplayDatapoints) {
-            OptionalLong delay = getDelay(d.timestamp, timeSinceOccurrenceStarted, schedule.orElse(null));
-            if (delay.isEmpty()) return predictedDatapoints;
-            long timestamp = delay.getAsLong() + now;
-
-            if (status.equals(PredictedDatapointWindow.BOTH)) {
-                predictedDatapoints.add(new SimulatorReplayDatapoint(timestamp*1000, d.value).toValueDatapoint());
+        if (status.equals(PredictedDatapointWindow.BOTH)) {
+            if (schedule.map(s -> s.getCurrent() == null).orElse(false)) return predictedDatapoints;
```
MartinaeyNL: Couldn't you make this more readable using something like this?
(I'm not sure if it works, so please take a look at this yourself haha)
```
if (schedule.map(Schedule::getCurrent).isEmpty()) {
```

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -251,56 +286,106 @@ public List<ValueDatapoint<?>> calculatePredictedDatapoints(
...
-        @JsonSchemaFormat("date-time")
-        @JsonSchemaTypeRemap(type = String.class)
-        protected Date start;
+    public static class Schedule implements Serializable {
 
-        @JsonPropertyDescription("Not implemented, within the recurrence rule you can specify an end date.")
-        @JsonSchemaFormat("date-time")
-        @JsonSchemaTypeRemap(type = String.class)
-        protected Date end;
+        private static class EpochLocalDateTimeDeserializer extends JsonDeserializer<LocalDateTime> {
```
MartinaeyNL, line 335: Shouldn't all these top-level classes live in their own files?
I notice it was already the case, but it looks quite messy.

agent/src/main/java/org/openremote/agent/protocol/AbstractProtocol.java
```diff
@@ -117,6 +120,7 @@ final public void linkAttribute(String assetId, Attribute<?> attribute) throws E
         try {
             doLinkAttribute(assetId, attribute, agent.getAgentLink(attribute));
         } catch (Exception e) {
+            log.error("e: ", e);
```
MartinaeyNL: Remove this before merging 😉

agent/build.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
 
     testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
+    testImplementation 'org.mockito:mockito-core:5.20.0'
```
MartinaeyNL: As this dependency is also used in the model package,
I would use the variable approach and specify its version in the `gradle.properties` file.

ui/component/or-attribute-input/src/scheduler-renderer.ts
MartinaeyNL, line 1: Add license header

ui/component/or-attribute-input/src/scheduler-renderer.ts
```diff
@@ -0,0 +1,101 @@
+import {
+    RankedTester,
+    rankWith,
+    ControlProps,
+    mapStateToControlProps,
+    mapDispatchToControlProps,
+    and,
+    uiTypeIs,
+    formatIs
+} from "@jsonforms/core";
```
MartinaeyNL: I would think, if it's necessary to configure some jsonforms specific logic at the "consumer level", then our own `@openremote/or-jsonforms` needs to expose these variables, right?

As in, you don't want `or-attribute-input` to have a dependency on `@jsonforms/core`.

Or, if the `scheduler-render` class is something **every** consumer needs,
then I'd export a `SchedulerRenderer` base class that everyone can extend upon.

ui/component/or-mwc-components/test/fixtures/index.ts
MartinaeyNL, line 1: Add license header 😉

ui/component/or-mwc-components/test/fixtures/index.ts
```diff
@@ -0,0 +1,26 @@
...
+    constructor(private readonly page: Page) {}
+
+    /**
+     * Returns the locator for the specified {@link InputType.SELECT|select input} option
+     */
+    getSelectInputOption(option: string, locator?: Locator): Locator {
+        return (locator ?? this.page).locator("or-mwc-input li[role=option]").getByText(option, { exact: true });
+    }
+
+    // getSelect
```
MartinaeyNL: Should this comment be here?

ui/component/or-rules/src/or-rule-viewer.ts
```diff
@@ -246,9 +259,25 @@ export class OrRuleViewer extends translate(i18next)(LitElement) {
                             <or-translate value="enabled"></or-translate>
                             <or-mwc-input .type="${InputType.CHECKBOX}" .value="${this.ruleset && this.ruleset.enabled}" ?disabled="${!this.ruleset.id}" @or-mwc-input-changed="${this._toggleEnabled}"></or-mwc-input>
                         </span>
-                        <or-rule-validity id="rule-header-validity" .ruleset="${this.ruleset}"></or-rule-validity>
+                        <or-scheduler
+                            id="rule-header-validity"
+                            header="${i18next.t("scheduleRuleActivity")}"
```
MartinaeyNL: Maybe we could auto-translate this inside `<or-scheduler>` ?

ui/component/or-rules/src/or-rule-viewer.ts
```diff
@@ -246,9 +259,25 @@ export class OrRuleViewer extends translate(i18next)(LitElement) {
...
+                                recurrence: i18next.t("planRecurrence"),
+                            } as LabeledEventTypes}"
+                            disabledRRuleParts="${DISABLED_RRULE_PARTS}"
+                            @or-scheduler-changed="${(event: OrSchedulerChangedEvent | undefined) => {
+                                if (this.ruleset) {
+                                    this.ruleset.meta ??= {};
+                                    this.ruleset.meta.validity = event?.detail.value;
+                                    this.requestUpdate("ruleset");
+                                }
+                            }}"
```
MartinaeyNL: I would move this to a separate callback function to allow function overrides.

ui/component/or-scheduler/src/util.ts
MartinaeyNL, line 1: For consistency purposes across components, I'd rename the file to either `model.ts` or `util.ts`

MartinaeyNL, line 1: Add the license header 😉

ui/component/or-scheduler/src/index.ts
```diff
@@ -0,0 +1,555 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+import {html, LitElement, PropertyValues, TemplateResult} from "lit";
+import {customElement, property, query, state} from "lit/decorators.js";
+import {CalendarEvent} from "@openremote/model";
+import {InputType, OrInputChangedEvent} from "@openremote/or-mwc-components/or-mwc-input";
+import {translate,i18next} from "@openremote/or-translate";
+
```
MartinaeyNL: You can remove the whitespace here.

ui/component/or-scheduler/src/index.ts
```diff
@@ -0,0 +1,555 @@
...
+import {OrMwcDialog, showDialog} from "@openremote/or-mwc-components/or-mwc-dialog";
+import {Frequency as FrequencyValue, RRule, Weekday, WeekdayStr} from 'rrule'
+import moment from "moment";
+import { Days } from "rrule/dist/esm/rrule";
```
MartinaeyNL, line 28: Is this the correct import path? Because it looks really weird 😂

ui/component/or-scheduler/src/index.ts
```diff
@@ -0,0 +1,555 @@
...
+}
+
+@customElement("or-scheduler")
+export class OrScheduler extends translate(i18next)(LitElement) {
+
+    @property()
+    public calendarEvent?: CalendarEvent = this.default;
+
+    @property()
+    public default?: CalendarEvent;
```
MartinaeyNL: Why is this default here? Maybe these properties need a description before I can properly understand them.

ui/component/or-scheduler/src/index.ts
```diff
@@ -0,0 +1,555 @@
...
+    public eventType: EventTypes = EventTypes.default;
+
+    @property()
+    public eventTypes: LabeledEventTypes = EventTypes;
+
+    @property()
+    public header?: string;
+
+    @property()
+    public isAllDay = true;
```
MartinaeyNL: Preferably these properties include a `@property({type: XXXX})`. It's very convenient in practice,
because `type: boolean` replaces `<or-scheduler isAllDay="true">` with `<or-scheduler isAllDay>`
It's also useful for documentation purposes.

ui/component/or-scheduler/package.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "@openremote/or-scheduler",
+    "version": "1.8.0-snapshot",
```
MartinaeyNL: You can update this version 😉

ui/component/or-mwc-components/test/fixtures/index.ts
Ekhorn: nope

ui/component/or-scheduler/src/index.ts
Ekhorn: Well it's not directly exported under their root namespace.

The `rrule` package doesn't define any `exports` in their `package.json` and thus it exports all source files.

However this object is not registered on their root module. It's only accessible under a subpath.

There is something to say that it might mean the intention of the developer is to not expose this, but they should've defined the `package.json` exports correctly. The package hasn't updated in 2 years and the only reason we use it is to have the time label format and data like the `Days` object.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: Well the de/serializer are meant to be used by the Schedule specifically for the Simulator. If there is some other location these usually go, but I see we don't have many custom de/serializers and the once we do have are nested like this.

I guess the schedule class doesn't need to be directly part of the protocol, but I would like some more input what to do here.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
Ekhorn: Haven't pushed this yet :wink:

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
Ekhorn: Well no matter what, I need a way to distinguish this so the frontend knows to use the custom renderer. Using the component name makes sense to me, but it could also be just `scheduler` or more specific `simulator-schedule` (for which we then can display the related JSON forms renderer).

Something else I tried was extending the `SimulatorProtocol.Schedule` from the `CalendarEvent` class so it generated a dedicated type definition in the JSON Schema that we can check for instead of this format, but these have diverged too much in types. TMK the `Date` type and the `LocalDateTime` type don't overlap directly.

Ekhorn: Also we can always change this down the line as this is an internal mapping.

ui/component/or-scheduler/src/index.ts
Ekhorn: Because 1 of the event types is `default` and can be set by the user.

ui/component/or-attribute-input/src/scheduler-renderer.ts
Ekhorn: I'll expose them from the JSON Forms.

The scheduler-renderer could be part of the JSON Forms, but that needs further standardization. I'll write an issue for that.

ui/component/or-scheduler/src/index.ts
MartinaeyNL: All good. Let's stick to this then 👍

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorAgentLink.java
MartinaeyNL: In my opinion, any name / identifier other than `or-scheduler` is OK.

I just want to prevent any references to the "frontend name", causing confusion in the backend.
As you mentioned, an identifier focused on the object, like `simulator-schedule`, seems more logical to me.

ui/test/fixtures/component.ts
```diff
@@ -1,23 +1,39 @@
 import type { Locator } from "@playwright/experimental-ct-core";
+import { Shared } from "./shared";
 
 type ComponentProps<Component extends HTMLElement> = Partial<Component>;
 type ComponentSlot = number | string | ComponentSlot[];
 type ComponentSlots = Record<string, ComponentSlot> & { default?: ComponentSlot };
 
-type ComponentEvents = Record<string, Function>;
+type ComponentEvents = Record<string, (detail: any) => void>;
```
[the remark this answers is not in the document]
Ekhorn: Types will be incompatible and I don't know how I would grab a parameter type from a callback through generics. Also the param is the `detail` of `CustomEvent` which is inherently `any` without applying generics.

ui/component/or-scheduler/src/index.ts
```diff
@@ -0,0 +1,582 @@
...
+
+function range(start: number, end: number): number[] {
+    return Array.from({ length: end - start + 1 }, (_, i) => start + i )
+}
+
+// TODO: use es2023
+function toSpliced<T>(arr: T[], index: number, count: number): T[] {
+    arr.splice(index, count);
+    return arr;
+}
```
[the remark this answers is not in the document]
Ekhorn: Made it shallow copy instead.

MartinaeyNL (dismissed):
Great work!

MartinaeyNL (approved):

PR #2205, opened 2025-10-22
Fixed migrations to use the supplied schema for the indexes for asset_datapoint_id and asset_predicted_datapoint_id.

Miggets7:
Migration history is preserved


MartinaeyNL (approved):
Thanks!

PR #2222, opened 2025-10-31
Add postgres tmpfs test container

Ekhorn:
## Description

This uses the Docker [tmpfs mount](https://docs.docker.com/engine/storage/tmpfs/) feature that mounts to [tmpfs](https://en.wikipedia.org/wiki/Tmpfs) in the Linux kernel. Writes to memory instead of disk which should improve write speeds. Though if low on memory might switch to swap.

Benefits
- Faster writes for tests
- Separate storage

~~Issues~~
~~- Additional memory used (spinning up additional test containers) (about 1gb)~~

## Checklist

### DIFF
--- a/manager/src/main/java/org/openremote/manager/setup/database/V20250916_01__AddServiceRoles.java
+++ b/manager/src/main/java/org/openremote/manager/setup/database/V20250916_01__AddServiceRoles.java
@@ -50,13 +50,21 @@ public class V20250916_01__AddServiceRoles extends BaseJavaMigration {
     @Override
     public void migrate(Context context) throws Exception {
 
+        // Get the Keycloak port value from system properties for tests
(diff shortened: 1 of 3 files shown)


profile/dev-testing.yml
Ekhorn, line 1: Consider `dev-ui` as well, so we can run all tests with this.

MartinaeyNL:
I don't know the context of this in full, but wouldn't it be better to use separate profiles for this?
As in, `dev-testing` is regularly used for local development, where these two additional containers would create overhead.

Ekhorn:
Added a dedicated profile for our tests.

profile/dev-testing.yml
Ekhorn: done!

Issue #1627, opened 2025-01-24
Investigate approach and effort to implement mTLS


DonWillems:

pankalog:
@ebariaux @richturner, thanks for the meeting on Monday!

Please find a potential connection/autoprovisioning sequence diagram for mTLS below. It's quite crowded because it contains multiple alternative flows, depending on if we want to auto-provision a user that authenticated via mTLS during connection, or after they subscribe to a given topic.

I've purposefully ignored the precise way we get/create a client/service user, as it is implementation-specific and not very helpful at this stage as discussed in the meeting.

Looking forward to your feedback!

```mermaid
sequenceDiagram
    autonumber
    participant D as MQTT Device<br/>(x509 cert)
    participant B as MQTT Broker<br/>(ActiveMQ Artemis)
    participant M as OpenRemote Manager
    participant K as Keycloak

    Note over D,B: Device connects using x509 certs (mTLS)
    D->>B: TLS ClientHello + client cert
    B-->>D: TLS ServerHello + server cert
    D-->>B: CertificateVerify / Finished
    B-->>D: Finished
    Note over D,B: mTLS established

    B->>M: Hand off connection/session
    M->>M: Extract clientId & realm from DN of last cert in chain

    alt Autoprovisioning DURING connect (alternative flow)
        M->>K: Get Client's Service Account
        K-->>M: 404 Not Found
        M->>M: Select autoprovisioning policy
        M->>K: Create confidential client
        K-->>M: 201 Created + credentials/metadata
        M->>M: Create assets, autoprovisioning flow
        M-->>B: Authorize CONNECT (new subject)
        B-->>D: CONNACK (success)
        Note over M,D: Manager could PUBLISH provisioning summary<br/>(Keycloak + OR asset details) right after connect<br/>to some provisioning topic
    end

    alt Autoprovisioning AFTER connect (main flow)
        M->>K: Get Client's Service Account
        alt Service account found
            K-->>M: 200 OK + subject info
            M->>M: Evaluate allow/deny policy
            alt Allowed
                M-->>B: Authorize CONNECT
                B-->>D: CONNACK (success)
            else Denied
                M-->>B: Reject CONNECT
                B-->>D: CONNACK (failure)
            end
        else Not found (autoprovisioning scenario)
            K-->>M: 404 Not Found
            M->>M: Mark "needs autoprovisioning"
            M-->>B: Authorize CONNECT as Subject<br/> with no permissions
            B-->>D: CONNACK (success)
[...]

PR #2226, opened 2025-11-02
MQTT Broker: Allow connection, authentication, and auto-provisioning of a client to the Broker using X509 Client Certificates (mTLS)

### LINKED ISSUE #1626: Auto-provisioning: X.509 mTLS implementation
X.509 mTLS

- [x] #1627

### LINKED ISSUE #2246: MQTT Broker: Allow authentication and autoprovisioning of client devices via X.509 Client Certificates
## User Story
As an OpenRemote user, I want to connect to the MQTT broker using the Client Certificates of an OpenRemote Service User, so that I can connect to the OpenRemote broker, without hard-coding username and password credentials in my connections.

Along with that, I want to auto-provision a device that connects using client certificates, with the information that is provided in the certificate and the corresponding `ProvisioningConfig`.

## Description
A lot of third-party devices that are able to communicate with OpenRemote, Like [Teltonika Telematics](https://www.teltonika-gps.com/) devices, are only able to do so using a client certificate, instead of username/password.

It is also safer to use Client Certificates instead of Username/Password, because the private key never leaves the client device and cannot be easily reused elsewhere, providing strong cryptographic proof of identity, rather than relying on a shared secret. 

Thus, we require the MQTT broker to allow users to authenticate to OpenRemote using client certificates.

Also, we need to allow the connected client to auto-provision itself using the provided client certificate, in the same provisioning flow as with current functionality. The client should publish a message to the auto-provisioning topic, and their certificate is checked. 

Enabling mTLS requires setting the keystore and truststore, and their passwords, so this is not something managed through `KeystoreService`.

The format of the certificate Distinguished Name is what is used for determining the identity of the authenticated subject. The `CN` field should be the service user's username, and the `OU` field should be the realm it is referring to. If OpenRemote code is reached, we can assume that the mTLS handshake has completed and we can assume that the user is "authorized" to be authenticated.

## Acceptance Criteria
- [x] 1. Client is able to connect to OpenRemote using Client Certificates
- [x] 2. If the user exists, authenticate them using Keycloak
- [x] 3. If the user doesn't exist, allow the connection as anonymous and store the certificate as a private credential
- [x] 4. If the client asks for auto-provisioning, use the added certificate private credential to create the service user/asset.

## Implementation Notes
The fundamental reasons why the implementation was somewhat short, is that:
- `ManagerIdentityProvider#getUserByUsername` returns the service user's secret, which means that we can then add that secret as a private credential, which the other LoginModule in the chain is able to use to perform a `client_credentials` request to Keycloak.
- The autoprovisioning MQTT handler was easily modifiable to allow checking for private credentials, which stores the client certificate. If it is present, then we can use what is defined in the CN and OU of the DN to create the service user etc.

pankalog:
## Description
This PR enables connecting, authenticating, and auto-provisioning a client to OpenRemote's MQTT broker using X509 mTLS Client Certificates.

To do this, it enables X.509 mTLS authentication by:

1. Adding a new Artemis acceptor with the needed keystores setup and `needClientAuth=true`
2. Modifying the `ActiveMQORSecurityManager` to: 
  a. Move the `GuestLoginModule` to the bottom of the `AppConfigurationEntry` array, so that it runs last, since we now have to check for the existence of an X509 certificate as well, not just a password
  b. Use the `ManagerIdentityProvider#getUserByUsername` method to retrieve the Confidential Client's secret when calling `getAuthenticatedSubject`, and then setting that as the password used for authentication (if that user exists)
3. Modifying the `MultiTenantClientCredentialsGrantsLoginModule` to check for the existence of a certificate, and if it exists, insert a `X509CertificatesCredential` into the private credentials, used for identifying and retrieving certificates during the autoprovisioning process.

As a result, when a user tries to connect to the broker with client certificates, first, the certificates are authenticated by Artemis for the mTLS handshake which assures that the client is who they claim they are, then:

- If there is a service user's username with the certificate subject's `CN` field in the realm specified by `OU`, they assume that identity
- If there isn't a service user, they are assigned `anonymous` `Principal`s, but they are also assigned with a private credential that contains their leaf client certificate
- If no password or no certificates are provided, only the anonymous `Principal`s are added, and the user is allowed to connect as a guest.

This PR also enables auto-provisioning functionality for connections established by means of a client certificate, by:

- Modifying `UserAssetProvisioningMQTTHandler#processX509ProvisioningMessage` to allow a new `MTLSProvisioningMessage`, which a connected client can send to the autoprovisioning request topic
- Depending on the type of `ProvisioningMessage`, the aforementioned `X509CertificatesCredential` is used to create the new service user.
- Modifying `UserAssetProvisioningMQTTHandler#getCreateClientServiceUser` to allow a custom client prefix, since the service user's username should be the same as the X500 Subject's CN. 

## Why this works

There are two main reasons why this feature could somewhat easily be added to OpenRemote:

1. I was not aware that `ManagerIdentityProvider#getUserByUsername` could return the service user's secret. This makes it a lot easier to keep the main core of the `ActiveMQORSecurityManager` the same as it was. 

**Something to note:** eventually, when we do reduce the tightness of the coupling of OpenRemote with Keycloak, this will have to also change. The way that this should be changed in the future, is by implementing a new `LoginModule` that uses [Standard Token Exchange](https://www.keycloak.org/securing-apps/token-exchange#_standard-token-exchange) to exchange a confidential client's JWT token for another confidential client's JWT token.

2. The unexpected ease of modifying `UserAssetProvisioningMQTTHandler` to add support for reading that `X509CertificateCredential`, so that the provisioning flow can then continue with no other effects. This made the implementation effort minimal.

In totality, I was expecting to write completely new classes for a `LoginModule` and a new Autoprovisioning MQTT handler, but I'm happy to report that wasn't the case.

## How to make this work

While I intend to write more extensive documentation when this is close to merging (even though I do have something in the works), here's how someone could configure OpenRemote to get this to work:

1. Create the server keystore and truststore, create the CA, and the client certificates, with the correct EKUs for each, and the correct format for the clients' subject DN: `CN=<service user name>,OU=<realm>`
2. Set all `OR_MQTT_MTLS` environment variables 
4. Use the root CA certificate as a server certificate, the client cert's public key, and the client cert's private key, to connect to the broker at the broker's mTLS port (default is `8884`). 
5. If you want to auto-provision, subscribe to the autoprovisioning response topic, and send a message to the request topic, with a message like `{"type":"mtls"}`.
6. You should then receive the autoprovisioning response data on the response topic. 

## How to verify it works

The `UserAndAssetProvisioningTest#mTLS MQTT Autoprovisioning integration test` is a full integration test of everything discussed above:

- It generates the needed certificates (CA, client certs) within the test, instead of using stored tests, which can also help documentation efforts, and also double-checks that compatibility isn't broken. A simple LLM query of the test can provide a fully documented guide for setup.
- It tests the mTLS functionality using `MQTT_IOClient` (indirectly also testing the MQTT Agent's client mTLS functionality, which could then replace the flaky mTLS test there)
- It creates 3 different situations: connecting with the client cert of an already created service user, connecting with an invalid client cert that hasn't been signed by the CA, and connecting with a certificate that currently does not have a service user
- All connections are checked for their subject and Principals after authentication, to assert that everything is populated correctly and no Principals are inadvertently added
- It tests the autoprovisioning functionality by using the above unprovisioned client to perform the auto-provisioning flow and assert that everything has been created correctly.

## Further Work

- This PR does not contain any changes to the `KeystoreService`, as we can assume that someone that would like to use client certificates is proficient enough at setting up keystores and environment variables. Down the road, if we want tighter integration with our UI and `KeystoreService`, we could potentially create a new `SSLContextFactory`, and plug that in to the Acceptor, so that it uses that instead of its own `DefaultSSLContextFactory`. 
- This PR does not contain changes to the proxy, which I need to experiment more with and ensure that everything works correctly when passing mTLS traffic from the proxy to the manager.
- This PR also does not change anything UI-wise. This is purely a backend change, and no UI elements need to be changed, as a user can create the provisioning config in the same way as they would for the standard autoprovisioning config.

Closes #1626 #1627 

## Checklist


pankalog:
I've marked this PR as ready for review. There are still some lines of code that are there for debugging reasons like print statements, but they will be removed as we get close to merging, so you can comment on them if you would like but I will go through the PR again before merging to make sure all logging statements are correct.

Also, more than half of this PR is about setting up and performing the test. To make the cognitive load lighter, I created a helper class in the Groovy file that handles everything related to keystores and certificate generation. It was entirely vibe-coded but thoroughly tested, so you can skip that entire class since it is only there as testing infrastructure.

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +104,58 @@ protected Subject getAuthenticatedSubject(String user,
...
                 user = realmAndUsername[1];
             }
+        }else if (certs != null && certs.length > 0) {
+
+            X509Certificate leaf = certs[0];
+            String dn = leaf.getSubjectX500Principal().getName();
+            try {
+                // Check for Client Authentication EKU (This denotes that this is the actual client certificate to enforce)
+                if(!leaf.getExtendedKeyUsage().contains("1.3.6.1.5.5.7.3.2")) {
+                    // TODO: Not sure about what extent to which we would like to enforce this.
```
richturner: LOG and fail as it suggests misconfiguration and we shouldn't make exceptions to things not following best practices

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +104,58 @@ protected Subject getAuthenticatedSubject(String user,
...
+                    // auth is the correct one to inspect.I will have to perform some more research about it, but I think
+                    // that the client certificate is always going to be the leaf. For now, log a warning about it.
+                    LOG.log(Level.WARNING, "Presented certificate DOES NOT have Client Authentication Extended Key Usage. " +
+                            "Attempting to use, but please fix this for subsequent requests.");
+                }
+                LdapName ldapName = new LdapName(dn);
+                for (Rdn rdn : ldapName.getRdns()) {
+                    String type = rdn.getType();
+                    String value = rdn.getValue().toString();
+                    if ("OU".equalsIgnoreCase(type) && realm == null) {
```
richturner: put `realm == null` check first it is much quicker than the string comparison check (don't think the compiler is smart enough to reorder such things for you)

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +104,58 @@ protected Subject getAuthenticatedSubject(String user,
...
+                        realm = value;
+                    } else if ("CN".equalsIgnoreCase(type) && (user == null || user.isEmpty())) {
```
richturner: Use `TextUtil.isNullOrEmpty()` and put as first logical check as explained above

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +104,58 @@ protected Subject getAuthenticatedSubject(String user,
...
+                        user = value;
+                    }
+                }
+            } catch (InvalidNameException e) {
+                LOG.log(Level.FINE, "Failed to parse subject DN from client certificate: " + dn, e);
+            } catch (Exception e) {
+                LOG.log(Level.SEVERE, "Failed to process given client certificate");
+            }
+
+            if(!(user == null || user.isEmpty())) {
```
richturner: Use `TextUtil.isNullOrEmpty()`

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +104,58 @@ protected Subject getAuthenticatedSubject(String user,
...
+            X509Certificate leaf = certs[0];
+            String dn = leaf.getSubjectX500Principal().getName();
+            try {
+                // Check for Client Authentication EKU (This denotes that this is the actual client certificate to enforce)
+                if(!leaf.getExtendedKeyUsage().contains("1.3.6.1.5.5.7.3.2")) {
+                    // TODO: Not sure about what extent to which we would like to enforce this.
+                    // I have seen codebases use this EKU key to ensure that the certificate they are examining for client
+                    // auth is the correct one to inspect.I will have to perform some more research about it, but I think
+                    // that the client certificate is always going to be the leaf. For now, log a warning about it.
+                    LOG.log(Level.WARNING, "Presented certificate DOES NOT have Client Authentication Extended Key Usage. " +
```
richturner: Don't use first person language in code comments

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
```diff
@@ -107,6 +105,13 @@ public class MQTTBrokerService extends RouteBuilder implements ContainerService,
     public static final int PRIORITY = MED_PRIORITY;
     public static final String MQTT_SERVER_LISTEN_HOST = "MQTT_SERVER_LISTEN_HOST";
     public static final String MQTT_SERVER_LISTEN_PORT = "MQTT_SERVER_LISTEN_PORT";
+    public static final String OR_MQTT_MTLS_SERVER_LISTEN_HOST = "OR_MQTT_MTLS_SERVER_LISTEN_HOST";
+    public static final String OR_MQTT_MTLS_SERVER_LISTEN_PORT = "OR_MQTT_MTLS_SERVER_LISTEN_PORT";
+    public static final String OR_MQTT_MTLS_KEYSTORE_PATH = "OR_MQTT_MTLS_KEYSTORE_PATH";
+    public static final String OR_MQTT_MTLS_KEYSTORE_PASSWORD = "OR_MQTT_MTLS_KEYSTORE_PASSWORD";
+    public static final String OR_MQTT_MTLS_TRUSTSTORE_PATH = "OR_MQTT_MTLS_TRUSTSTORE_PATH";
+    public static final String OR_MQTT_MTLS_TRUSTSTORE_PASSWORD = "OR_MQTT_MTLS_TRUSTSTORE_PASSWORD";
+//    public static final String OR_MQTT_MTLS_ENABLED = "MQTT_MTLS_ENABLED";
```
richturner: Should be used but changed to `OR_MQTT_MTLS_DISABLED` [default: `false`]

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
```diff
@@ -262,16 +273,16 @@ public void start(Container container) throws Exception {
...
                         MultiTenantClientCredentialsGrantsLoginModule.INCLUDE_REALM_ROLES_OPTION, "true",
                         AbstractKeycloakLoginModule.ROLE_PRINCIPAL_CLASS_OPTION, RolePrincipal.class.getName()
                     )),
-                    new AppConfigurationEntry(PrincipalConversionLoginModule.class.getName(), AppConfigurationEntry.LoginModuleControlFlag.REQUISITE, Map.of(PrincipalConversionLoginModule.PRINCIPAL_CLASS_LIST, KeycloakPrincipal.class.getName()))
+                    new AppConfigurationEntry(PrincipalConversionLoginModule.class.getName(), AppConfigurationEntry.LoginModuleControlFlag.REQUISITE, Map.of(PrincipalConversionLoginModule.PRINCIPAL_CLASS_LIST, KeycloakPrincipal.class.getName())),
+                    new AppConfigurationEntry(GuestLoginModule.class.getName(), AppConfigurationEntry.LoginModuleControlFlag.SUFFICIENT, Map.of("debug", "true", "credentialsInvalidate", "true", "org.apache.activemq.jaas.guest.user", ANONYMOUS_USERNAME, "org.apache.activemq.jaas.guest.role", ANONYMOUS_USERNAME)),
```
richturner: Why did you need to change the order of login modules?

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
```diff
@@ -639,4 +650,52 @@ public void authenticateConnection(RemotingConnection connection, String realm,
...
+     * Add mTLS acceptor configuration to the server
+     */
+    protected void addMTLSAcceptor(Container container) throws Exception {
+        String mtlsHost = getString(container.getConfig(), OR_MQTT_MTLS_SERVER_LISTEN_HOST, "0.0.0.0");
+        int mtlsPort = getInteger(container.getConfig(), OR_MQTT_MTLS_SERVER_LISTEN_PORT, 8884);
+        final Path keystoreDirPath = Paths.get("keystores");
+        String keystorePath = getString(container.getConfig(), OR_MQTT_MTLS_KEYSTORE_PATH, persistenceService.resolvePath(keystoreDirPath.resolve("server_keystore.p12")).toString());
+        String keystorePassword = getString(container.getConfig(), OR_MQTT_MTLS_KEYSTORE_PASSWORD, "secret");
+        String truststorePath = getString(container.getConfig(), OR_MQTT_MTLS_TRUSTSTORE_PATH, persistenceService.resolvePath(keystoreDirPath.resolve("server_truststore.p12")).toString());
+        String truststorePassword = getString(container.getConfig(), OR_MQTT_MTLS_TRUSTSTORE_PASSWORD, "secret");
```
richturner: Move defaults to constants like we do everywhere else

model/src/main/java/org/openremote/model/provisioning/MTLSProvisioningMessage.java
richturner, line 1: Doesn't seem like this message type is needed, mTLS provisioning request doesn't require a payload unless there is something valuable to include then just the type descriminitor is sufficient

manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java
```diff
@@ -329,23 +329,43 @@ protected String getResponseTopic(Topic topic) {
...
+                .map(X509CertificateCredential::getCertificate)
+                .orElse(null);
+            if(clientCertificate == null){
+                LOG.info("No client certificate found in MTLS provisioning message: " + MQTTBrokerService.connectionToString(connection));
+                publishMessage(getResponseTopic(topic), new ErrorResponseMessage(ErrorResponseMessage.Error.CERTIFICATE_INVALID), MqttQoS.AT_MOST_ONCE);
+                return;
+            }
+
+            certUniqueId = ProvisioningUtil.getSubjectCN(clientCertificate.getSubjectX500Principal());
+            uniqueId = ProvisioningUtil.getSubjectCN(clientCertificate.getSubjectX500Principal());
```
richturner: Need to still verify the topic index 1 string matches the unique ID if we are still using the same handler

manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java
```diff
@@ -382,12 +398,52 @@ protected void processX509ProvisioningMessage(RemotingConnection connection, Top
...
+                // Why this is a problem: The device will subsequently disconnect from the broker and try to reconnect.
+                // The service user will be created in a different realm than the one that the device is trying to connect to.
+                // Which means that when the device reconnects it will fail to authenticate as the service user won't exist in that realm.
+                // Instead of making a weird decision about which realm to use, we just reject the connection outright.
+                LOG.info(() -> "mTLS certificate subject realm doesn't match provisioning config realm: " + MQTTBrokerService.connectionToString(connection));
+                publishMessage(getResponseTopic(topic), new ErrorResponseMessage(ErrorResponseMessage.Error.UNAUTHORIZED), MqttQoS.AT_MOST_ONCE);
+                return;
+            }
+        }
+
```
richturner: This seems to be repeating checks done in the authentication step but it raises an interesting point about using the OU to determine the realm, as the provisioning config currently determines which realm to provision the user and asset into now we have this potential mismatch and no way to allocate devices to realms on the manager just at cert generation time.

manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java
```diff
@@ -475,8 +531,9 @@ protected X509ProvisioningConfig getMatchingX509ProvisioningConfig(RemotingConne
             .orElse(null);
     }
 
-    public static User getCreateClientServiceUser(String realm, ManagerKeycloakIdentityProvider identityProvider, String uniqueId, ProvisioningConfig<?, ?> provisioningConfig) throws RuntimeException {
-        String username = (PROVISIONING_USER_PREFIX + uniqueId);
+    public static User  getCreateClientServiceUser(String realm, ManagerKeycloakIdentityProvider identityProvider, String uniqueId, String userPrefix, ProvisioningConfig<?, ?> provisioningConfig) throws RuntimeException {
```
richturner: Why introduce this configurable prefix, it doesn't seem to be serving any purpose

model/src/main/java/org/openremote/model/provisioning/ProvisioningUtil.java
```diff
@@ -61,6 +63,18 @@ public static X509Certificate getX509Certificate(String pem) throws CertificateE
...
+        sw.write("-----BEGIN CERTIFICATE-----\n");
+
+        // Encode the binary certificate data to Base64, with line wrapping at 64 chars
+        String base64 = Base64.getMimeEncoder(64, new byte[]{'\n'})
+                .encodeToString(certificate.getEncoded());
+        sw.write(base64);
+        sw.write("\n-----END CERTIFICATE-----\n");
+
+        return sw.toString();
+    }
```
richturner: Don't create methods that are only used in tests

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
```diff
@@ -639,4 +650,52 @@ public void authenticateConnection(RemotingConnection connection, String realm,
...
+        try {
+            String mtlsServerURI = new URIBuilder()
+                .setScheme("tcp")
+                .setHost(mtlsHost)
+                .setPort(mtlsPort)
+                .setParameter("protocols", "MQTT")
+                .setParameter("allowLinkStealing", "false")
+                .setParameter("defaultMqttSessionExpiryInterval", Integer.toString(DEFAULT_SESSION_EXPIRY_MILLIS))
+                // SSL/TLS configuration for mTLS
+                .setParameter("sslEnabled", "true")
```
richturner, line 704: How do server certs get into Artemis from the proxy?

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
```diff
@@ -833,4 +864,618 @@ class UserAndAssetProvisioningTest extends Specification implements ManagerConta
...
+            keystoreService.getKeyManagerFactory(unprovisionedKeyAlias),
+            keystoreService.getTrustManagerFactory()
+        )
+
+        unprovisionedClient.connect()
+
+        then: "the connection should be created, but there should be no roles, but there is a ProvisioningPrincipal, that contains the correct certificate"
+        conditions.eventually {
+            assert unprovisionedClient.getConnectionStatus() == ConnectionStatus.CONNECTED
+        }
```
richturner: The fact they can authenticate without a service user existing feels strange and no intuitive

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
```diff
@@ -833,4 +864,618 @@ class UserAndAssetProvisioningTest extends Specification implements ManagerConta
...
+            assert unprovisionedJaasSubject.getPrincipals(UserPrincipal)[0].getName() == "anonymous"
+            assert unprovisionedJaasSubject.getPrincipals(KeycloakPrincipal.class).isEmpty()
+            // Should have a ProvisioningPrincipal with the certificate
+            assert unprovisionedJaasSubject.getPrivateCredentials(X509CertificateCredential.class).size() == 1
+            def provisioningPrincipal = unprovisionedJaasSubject.getPrivateCredentials(X509CertificateCredential.class)[0]
+            assert provisioningPrincipal.getCertificate() != null
+            assert provisioningPrincipal.getCertificate().getSubjectX500Principal().getName().contains("CN=$unprovisionedUsername")
+        }
+
+        when: "the provisioning config is created to allow autoprovisioning using mTLS"
```
richturner, line 980: Move to separate test

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
```diff
@@ -833,4 +864,618 @@ class UserAndAssetProvisioningTest extends Specification implements ManagerConta
             clientEventService.removeSubscription {internalConsumer}
         }
     }
+
+    @SuppressWarnings("GroovyAccessibility")
+    def "mTLS MQTT Autoprovisioning integration test"() {
```
richturner, line 876: The whole first part of this test doesn't seem to be testing auto provisioning just mTLS authentication

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
```diff
@@ -833,4 +864,618 @@ class UserAndAssetProvisioningTest extends Specification implements ManagerConta
...
+
+        device1Client.connect()
+
+        then: "the client should connect successfully using mTLS"
+        conditions.eventually {
+            assert device1Client.getConnectionStatus() == ConnectionStatus.CONNECTED
+        }
+
+        and: "the authenticated client should have the correct subject"
+        conditions.eventually {
```
richturner: Integration test should test more behaviour, so they are authenticated and now they should be able to publish attribute events, etc.

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
```diff
@@ -833,4 +864,618 @@ class UserAndAssetProvisioningTest extends Specification implements ManagerConta
...
+            assert autoProvisioningResponses.size() == 1
+            assert autoProvisioningResponses.get(0) instanceof SuccessResponseMessage
+        }
+
+        when: "the device disconnects"
+        unprovisionedClient.disconnect()
+
+        and: "reconnects again"
+        autoProvisioningResponses.clear()
+        unprovisionedClient.connect()
```
richturner: rename autoprovisioningClient

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
```diff
@@ -833,4 +864,618 @@ class UserAndAssetProvisioningTest extends Specification implements ManagerConta
...
+
+        then: "the connection should be created with the provisioned user and roles"
+        conditions.eventually {
+            assert unprovisionedClient.getConnectionStatus() == ConnectionStatus.CONNECTED
+        }
+
+        and: "the authenticated client should have the correct subject with roles"
+        conditions.eventually {
```
richturner, line 1053: Again test behaviour

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
```diff
@@ -833,4 +864,618 @@ class UserAndAssetProvisioningTest extends Specification implements ManagerConta
...
+        }
+        unprovisionedClient.addMessageConsumer(autoProvisioningResponseTopic, autoProvisioningMessageConsumer);
+        unprovisionedClient.sendMessage(
+            new MQTTMessage<String>(autoProvisioningRequestTopic, ValueUtil.JSON.writeValueAsString(new MTLSProvisioningMessage(null)))
+        )
+
+        then: "we should receive a success response provisioning message"
+        conditions.eventually {
+            assert autoProvisioningResponses.size() == 1
+            assert autoProvisioningResponses.get(0) instanceof SuccessResponseMessage
```
richturner, line 1025: Maybe the autoprovisioning response should contain the realm the device is now provisioned within but not sure how to align this with the subject OU checks you are doing which means the device already needs to know which realm it is connecting to and this is baked into the client cert.

richturner (changes requested):

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
pankalog: Because when the `GuestLoginModule` is first, it invalidates the credentials and assigns `ANONYMOUS_USERNAME`, *without* checking if there are certificates present. So when certificates are present, it will check the username and password field, and when it finds them empty, it will return false and since it is using *SUFFICIENT*, it will fail the entirety of the chain. 

Now, the first login module is the custom one, which adds the required certificate credentials in `commit()`, then `doAuth`s with those credentials, authenticating the user. Of course, if nothing has been found until the `GuestLoginModule`, credentials will be invalidated and the guest roles will be assigned.

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
pankalog: This is behavior consistent with the current way the broker controls anonymous logins, with the only difference being the added `X509CertificateCredential`. They're allowed to login and have the same permissions as any anonymous user, but they can autoprovision. What should be the behavior for this?

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
pankalog: You don't need to get the Haproxy server certs.

The configured truststore contains a list of CAs/certificates which I (the server) am allowed to authorize ("Is the certificate the client sent signed by a CA I trust?").

The configured keystore contains your own (the server's) keypair. The client will check that to prove you're who you say you are ("Here is my certificate; prove I own the private key")
.  

As long as the server certificate (my) keypair is issued by a CA that the client trusts (which can be arbitrarily controlled by the user), then the device can connect. There's no "the CA that signed the server needs to sign me as well".

In fact, you cannot quite use the LetsEncrypt certificates for signing client certificates. They're not made for signing other certificates (meaning they're not allowed to be CAs, they don't have the correct EKUs and extensions). 

When the general brushstrokes of this PR are nailed down, I will create a documentation guide, and also a bash script that is able to generate all needed files for mTLS to work.

model/src/main/java/org/openremote/model/provisioning/MTLSProvisioningMessage.java
pankalog: So how am I supposed to indicate that there exists a different type of `ProvisioningMessage`? I was thinking of removing the `req` parameter since it is not used, but we would still need to have `MTLSProvisioningMessage` so that `{"type":"mtls"}` can be parsed properly, right?

richturner: Indeed what I'm saying is keep the class for type descrimination but there is no payload in the message so yes the `req` parameter is not needed/used which I see you've now removed so all good

model/src/main/java/org/openremote/model/provisioning/ProvisioningUtil.java
richturner, line 1: Unused imports please check each file in your PR for these simple issues

manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java
```diff
@@ -382,12 +404,41 @@ protected void processX509ProvisioningMessage(RemotingConnection connection, Top
...
+            }
+        }
+
         // Get/create service user
         User serviceUser;
 
         try {
             LOG.finest("Checking service user: " + uniqueId);
-            serviceUser = getCreateClientServiceUser(realm, identityProvider, uniqueId, matchingConfig);
+            serviceUser = getCreateClientServiceUser(realm, identityProvider, uniqueId, isMTLS ? "" : null, matchingConfig);
```
richturner: What value does empty string provide?

manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java
richturner: We discussed looking at the MQTT connect username to determine the realm

manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java
```diff
@@ -475,8 +526,9 @@ protected X509ProvisioningConfig getMatchingX509ProvisioningConfig(RemotingConne
             .orElse(null);
     }
 
-    public static User getCreateClientServiceUser(String realm, ManagerKeycloakIdentityProvider identityProvider, String uniqueId, ProvisioningConfig<?, ?> provisioningConfig) throws RuntimeException {
-        String username = (PROVISIONING_USER_PREFIX + uniqueId);
+    public static User getCreateClientServiceUser(String realm, ManagerKeycloakIdentityProvider identityProvider, String uniqueId, String userPrefix, ProvisioningConfig<?, ?> provisioningConfig) throws RuntimeException {
+        String prefix = userPrefix != null ? userPrefix : PROVISIONING_USER_PREFIX;
+        String username = (prefix + uniqueId);
```
richturner: Why are we changing this logic? Useful to be able to identify service users that have been created for provisioning purposes in a consistent way

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +112,84 @@ protected Subject getAuthenticatedSubject(String user,
...
             if (realmAndUsername.length == 2) {
                 realm = realmAndUsername[0];
                 user = realmAndUsername[1];
             }
+        } else if (certs != null && certs.length > 0) {
+            List<X509Certificate> clientAuthCerts = Arrays.stream(certs).filter(e -> {
+                try {
+                    return e.getExtendedKeyUsage().contains(CLIENT_AUTH_EKU_OID);
+                } catch (CertificateParsingException ex) {
+                    throw new RuntimeException(ex);
```
richturner: Log parsing error and return null

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +112,84 @@ protected Subject getAuthenticatedSubject(String user,
...
+                } catch (NullPointerException ignored){return false;}
```
richturner: Guard against NPE and don't use catch

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +112,84 @@ protected Subject getAuthenticatedSubject(String user,
...
+            }).toList();
+
+            if (clientAuthCerts.size() != 1) {
+                String errMsg = "Presented certificate chain contains " + clientAuthCerts.size() +
+                        " certificates with Client Authentication Extended Key Usage. " +
+                        "Expected exactly 1 certificate. Please provide a valid certificate chain. "+ connectionToString(remotingConnection);
+                LOG.log(Level.SEVERE, errMsg);
```
richturner: Warning level

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +112,84 @@ protected Subject getAuthenticatedSubject(String user,
...
+                    throw new LoginException("No realm found matching the username from client certificate.");
+                }
+            }
+
+            if(!TextUtil.isNullOrEmpty(user)) {
+                User dbUser = null;
+                User[] users = identityProvider.queryUsers(new UserQuery().usernames(new StringPredicate(SERVICE_ACCOUNT_PREFIX + user).match(AssetQuery.Match.EXACT)));
+                if (users != null && users.length == 1) {
+                    dbUser = users[0];
+                }else if(users != null && users.length > 1) {
```
richturner: Inconsistent code style, please use spaces after `if/else` statement and after curly braces

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +112,84 @@ protected Subject getAuthenticatedSubject(String user,
...
+                LOG.log(Level.FINE, "Failed to parse subject DN from client certificate: " + dn, e);
+            } catch (Exception e) {
+                LOG.log(Level.SEVERE, "Failed to process given client certificate");
+            }
+
+            if(realm == null){
+                LOG.log(Level.INFO, "Client certificate was provided, but no realm found in certificate subject. " +
+                        "Falling back to the username as a realm name. Subject DN=" + dn);
+
+                if (identityProvider.getRealm(user) != null) {
```
richturner: Check `user != null` to avoid NPE

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
```diff
@@ -155,6 +162,16 @@ public void init(Container container) throws Exception {
         executorService = container.getExecutor();
         timerService = container.getService(TimerService.class);
         assetProcessingService = container.getService(AssetProcessingService.class);
+        persistenceService = container.getService(PersistenceService.class);
+
+        // mTLS values
+        final Path keystoreDirPath = Paths.get("keystores");
+        this.mtlsPort = getInteger(container.getConfig(), OR_MQTT_MTLS_SERVER_LISTEN_PORT, OR_MQTT_MTLS_PORT_DEFAULT);
+        this.mtlsDisabled = getBoolean(container.getConfig(), OR_MQTT_MTLS_DISABLED, true);
```
richturner: Should it also be disabled if issues with paths/truststores

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
```diff
@@ -639,4 +662,47 @@ public void authenticateConnection(RemotingConnection connection, String realm,
...
+    /**
+     * Add mTLS acceptor configuration to the server
+     */
+    protected void addMTLSAcceptor(Container container) throws Exception {
+        if (TextUtil.isNullOrEmpty(this.keystorePath)
+           || TextUtil.isNullOrEmpty(this.truststorePath)
+           || TextUtil.isNullOrEmpty(this.keystorePassword)
+           || TextUtil.isNullOrEmpty(this.truststorePassword)) {
+            LOG.log(INFO, "MQTT mTLS acceptor not being started, as environment variables not configured");
+            return;
```
richturner: Should be moved into disabled logic and logged there

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
richturner: This is an endpoint that does TLS termination the client will want to verify the server's certificate which are managed by the proxy, this is what I'm referring to

richturner (changes requested):

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
pankalog: I'm assuming that there will be an issue with Artemis trying to access the keystores when there are any issues with the keystores and their passwords, so that will fail first.

manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java
pankalog: The client certificates don’t need to be signed by a keypair in the server’s keystore.

I think it's not worth the security implications and the Docker volume mounts that would be needed to have the same certificates for termination. There's also the issues with certificate expiration if we do that, since it then makes devices have to at least also store the CA themselves, which will eventually expire. I think it's reasonable to have the users of this feature have to perform their own setup to use it, and all of this would be documented at length for setting it up.

manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java
pankalog: Because in the older code:

Assume client presents itself with client certificate with DN `CN=device1, OU=building` on topic `provisioning/device1/request`. Provisioning is approved etc. etc., and we get to here. 

The service user is created as `service-account-ps-device1`. 

During authentication, client presents itself with the same CN. we add `service-account-` to that CN, leaving us with `service-account-device1`. The identity provider is then unable to find a service user with username `service-account-device1`, since we used the `service-account-ps-device1` username to provision.

So, we should either have this change here, to allow clients to use the new credentials when logged in, or all service users that want to connect using mTLS, regardless of whether that user was autoprovisioned or not, would need to have `ps-` prepended to their username.

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
pankalog: Moved to its own test

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
pankalog: What would you like to see added to these tests?

pankalog:
There is an issue on the HiveMQ MQTT client:

They have yet to migrate to new Netty 4.2, which means that some method references have changed, and I am hence getting `NoSuchMethodError` exceptions.

```Caused by: java.lang.NoSuchMethodError: 'boolean io.netty.internal.tcnative.SSL.setCurvesList(long, java.lang.String[])'```

Netty version is still the latest 4.1 version, [here's their versions file](https://github.com/hivemq/hivemq-mqtt-client/blob/0177edabac799920426989ea50e232caedf03feb/gradle/libs.versions.toml).

**Comment by pankalog** [comment]
@richturner as noted above, there's an issue with us using a newer version of Netty. The MQTT client we use hasn't been migrated yet to that.

You mentioned something about dependency choices, not sure what that was about? 

If you still want to review this PR and merge once the HiveMQ client has been migrated successfully, that's also an option, and I can make an issue on their GitHub repo for it.

For now, I have moved OpenRemote to the exact Netty version they used, and reverted my commit for fixing some deprecation notices from Netty 4.1 --> 4.2.

File: manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java (hunk 10 above)
Comment (pankalog) [reply manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java]: This was added in `ActiveMQORSecurityManager`.

File: manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java (hunk 13 above)
Comment (richturner) [reply manager/src/main/java/org/openremote/manager/mqtt/MQTTBrokerService.java:704]: Think we're missing my point here, part of mTLS is the standard client validation of the server i.e. SSL. The proxy was doing all of this including getting the SSL certs via lets encrypt but now we have a TLS endpoint outside of the proxy so the client will want to veryify the servers certs.

What server cert does Artemis use by default? we want to ideally use the valid SSL certs that the proxy manages

File: manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java [hunk 30]
```diff
@@ -113,68 +111,67 @@ protected Subject getAuthenticatedSubject(String user,
...
         X509Certificate[] certs = getCertsFromConnection(remotingConnection);
+        final String originalUsername = user;
         if (user != null) {
             String[] realmAndUsername = user.split(":");
             if (realmAndUsername.length == 2) {
                 realm = realmAndUsername[0];
                 user = realmAndUsername[1];
             }
-        } else if (certs != null && certs.length > 0) {
+        } if (certs != null && certs.length > 0) {
```
richturner: New line

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -113,68 +111,67 @@ protected Subject getAuthenticatedSubject(String user,
...
+             * The below logic supports using the Username field from the authentication attempt as a realm override.
+             * This is to support clients that are logging in using client certificates. They cannot change the realm they connect to
+             *  in the certificate DN, so we allow them to specify the realm in the username instead, as an override.
+             */
+            if (identityProvider.realmExists(originalUsername)) {
+                realm = originalUsername;
+            }
+
+
             if(realm == null){
```
richturner: Spacing

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -113,68 +111,67 @@ protected Subject getAuthenticatedSubject(String user,
...
-                    realm = user;
-                }else{
-                    LOG.log(Level.WARNING, "No realm found matching the username. " +
-                            "Failing login attempt. Subject DN=" + dn);
-                    throw new LoginException("No realm found matching the username from client certificate.");
-                }
+                throw new LoginException("No realm found matching the OU of the client certificate, or the Username field.");
             }
 
             if(!TextUtil.isNullOrEmpty(user)) {
```
richturner: Spacing

manager/src/main/java/org/openremote/manager/mqtt/UserAssetProvisioningMQTTHandler.java
```diff
@@ -339,7 +339,7 @@ protected void processX509ProvisioningMessage(RemotingConnection connection, Top
         String uniqueId;
         boolean isMTLS = provisioningMessage instanceof MTLSProvisioningMessage;
 
-        if (provisioningMessage instanceof X509ProvisioningMessage) {
+        if (!isMTLS) {
```
richturner: Don't like this logic as you're assuming if `!isMTLS` it must be `X509` works currently but will break if/when new type introduced.

Log and throw if unknown type provided

test/src/test/groovy/org/openremote/test/provisioning/UserAndAssetProvisioningTest.groovy
richturner: The below additions seem to address this

manager/src/main/java/org/openremote/manager/mqtt/ActiveMQORSecurityManager.java
```diff
@@ -98,13 +110,83 @@ protected Subject getAuthenticatedSubject(String user,
...
+
+            X509Certificate leaf = clientAuthCerts.getFirst();
+
+            String dn = leaf.getSubjectX500Principal().getName();
+            try {
+                user = ProvisioningUtil.getSubjectCN(leaf.getSubjectX500Principal());
+                realm = ProvisioningUtil.getSubjectOU(leaf.getSubjectX500Principal());
+            } catch (Exception e) {
+                LOG.log(Level.SEVERE, "Failed to process given client certificate");
+            }
```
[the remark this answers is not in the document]
pankalog: This already happens by Artemis, if artemis found any issues with validity then this code wouldn't be invoked.

docker-compose.yml
```diff
@@ -59,7 +61,10 @@ services:
   manager:
 #    privileged: true
     restart: always
-    image: openremote/manager:${MANAGER_VERSION:-latest}
+#    image: openremote/manager:${MANAGER_VERSION:-latest}
+    image: openremote-manager
+    build:
+      context: manager/build/install/manager
```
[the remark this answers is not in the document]
pankalog: Will be removed when ready to merge

PR #2265, opened 2025-11-18
Improve DefaultMQTTHandler multithreading and test reliability

richturner:
This removes the manual synchronisation that was in `DefaultMQTTHandler` and replaced it with `ConcurrentHashMap`
Also improved test robustness by cleaning up agents between each test.

### DIFF
--- a/test/src/test/groovy/org/openremote/test/protocol/mqtt/MQTTClientProtocolTest.groovy
+++ b/test/src/test/groovy/org/openremote/test/protocol/mqtt/MQTTClientProtocolTest.groovy
@@ -397,11 +397,11 @@ class MQTTClientProtocolTest extends Specification implements ManagerContainerTr
         }
 
         cleanup: "the agent and assets are deleted"
-        if (agent != null) {
-            assetStorageService.delete([agent.id])
-        }
         if (asset != null) {
             assetStorageService.delete([asset.id])
+        }
+        if (agent != null) {
+            assetStorageService.delete([agent.id])
         }
         conditions.eventually {
             assert brokerService.getUserConnections(keycloakTestSetup.serviceUser.id).isEmpty()
@@ -532,14 +532,14 @@ class MQTTClientProtocolTest extends Specification implements ManagerContainerTr
 
 
         cleanup: "the agent and assets are deleted"
-        if (agent != null) {
-            assetStorageService.delete([agent.id])
-        }
         if (asset != null) {
             assetStorageService.delete([asset.id])
         }
         if (testThing != null) {
             assetStorageService.delete([testThing.id])
+        }
+        if (agent != null) {
+            assetStorageService.delete([agent.id])
         }
         conditions.eventually {
             assert brokerService.getUserConnections(keycloakTestSetup.serviceUser.id).isEmpty()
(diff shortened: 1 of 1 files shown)


test/src/test/groovy/org/openremote/test/protocol/mqtt/MQTTClientProtocolTest.groovy
```diff
@@ -519,6 +527,14 @@ class MQTTClientProtocolTest extends Specification implements ManagerContainerTr
             assert asset.getAttribute("temperature").flatMap { it.value }.map { it == 21.5 }.orElse(false)
             assert asset.getAttribute("switchStatus").flatMap { it.value }.map { it == false }.orElse(false)
         }
+
+        when: "the agent is deleted"
+        assetStorageService.delete([agent.id, testThing.id, asset.id])
+
+        then: "the connection should be removed"
+        conditions.eventually {
+            assert brokerService.getUserConnections(keycloakTestSetup.serviceUser.id).isEmpty()
```
ebariaux, line 545: For consistency compared to above code, either use a one-liner in both tests or use connections variable in both

test/src/test/groovy/org/openremote/test/protocol/mqtt/MQTTClientProtocolTest.groovy
```diff
@@ -712,6 +728,14 @@ class MQTTClientProtocolTest extends Specification implements ManagerContainerTr
             assert asset.getAttribute("pressure").flatMap { it.value }.map { it == 1000 }.orElse(false)
             assert asset.getAttribute("uvIndex").flatMap { it.value }.map { it == 4.0 }.orElse(false)
         }
+
+        when: "the agent is deleted"
+        assetStorageService.delete([agent.id, testThing1.id, testThing2.id, testThing3.id, asset.id])
+
+        then: "the connection should be removed"
+        conditions.eventually {
+            assert mqttBrokerService.getUserConnections(keycloakTestSetup.serviceUser.id).isEmpty()
```
ebariaux, line 758: One liner or use variable

test/src/test/groovy/org/openremote/test/protocol/mqtt/MQTTClientProtocolTest.groovy
```diff
@@ -396,6 +395,15 @@ class MQTTClientProtocolTest extends Specification implements ManagerContainerTr
             assert connection != null
             assert defaultMQTTHandler.sessionSubscriptionConsumers.containsKey(getConnectionIDString(connection))
         }
+
+        when: "the agent is deleted"
+        assetStorageService.delete([agent.id, asset.id])
```
ebariaux: If the test fails sooner, then I suppose delete will not be performed
Do we want to add something in cleanup: as safeguard ?

test/src/test/groovy/org/openremote/test/protocol/mqtt/MQTTClientProtocolTest.groovy
richturner: Will adjust

test/src/test/groovy/org/openremote/test/protocol/mqtt/MQTTClientProtocolTest.groovy
richturner: Yeah fair point, I avoided using a cleanup block as I wanted to ensure the connections were closed before starting the next test, might just to do that at the start of each test

PR #2271, opened 2025-11-20
Improve When-Then JSON Rules UI for scaling to larger instances

### LINKED ISSUE #2264: When-then JSON rules UI slow with 50.000+ assets
### Description
As stated in #2233, the When-then JSON Rules page is not scalable to large OpenRemote instances.
When more than 50.000 assets of the same asset type are present in the system, (such as a light asset)
the dropdown where you select a specific asset is very slow with rendering.

### To Reproduce
1: Locally (or remotely) start a Manager instance with 50.000 assets.
You can use a custom setup, or use the newly introduced `load2` setup from #2210.
2. Visit `<your url>/manager/#/rules`, and make sure you are in the same realm as the 50k assets.
3. Create a new When-then JSON rule
4. In the creation window, add a When (or Then) clause, and select your asset type. (such as the light asset)
You notice that the first dropdown says "loading" for several seconds.

### Expected behavior
The time it takes to load the When and Then clauses should be decreased to at most 1 second

### Screenshot
Image

### Additional context
This is caused due to the frontend requesting **ALL** assets through the HTTP API.
So if you have 50k light assets, it is fetching 50k assets, being around 20MB of data in total.
With a metered connection, this is terrible performance.
And of course, it's bad for the UX overall.

MartinaeyNL:
## Description
Applied remote fetching of data when searching assets through the When-Then JSON Rules UI.
Before, it fetched **all** assets on initial creation of the rule, which didn't scale to 50.000+ assets.
Fixes #2264

## Changelog
- When-Then JSON Rules UI now queries assets remotely after typing in the search field.
This only applies to asset types with more than 100 assets.
- Improved and added UI tests for the When-Then JSON rules.

## Checklist

### DIFF
--- a/ui/app/shared/build.gradle
+++ b/ui/app/shared/build.gradle
@@ -172,13 +172,15 @@ tasks.register('installDist', Copy) {
 }
 
 tasks.register('processTranslations', JavaCompile) {
+    dependsOn(":agent:classes")
     dependsOn(":model:classes")
     outputs.upToDateWhen { false } // Always invalidate
 
+    def agentSourceSet = project(":agent").sourceSets.main
     def modelSourceSet = project(":model").sourceSets.main
 
-    source = modelSourceSet.allJava
-    classpath = modelSourceSet.compileClasspath
+    source = files(agentSourceSet.allJava, modelSourceSet.allJava)
+    classpath = files(agentSourceSet.compileClasspath, modelSourceSet.compileClasspath)
     destinationDirectory = layout.buildDirectory.dir("generated/i18n") // unused
     options.compilerArgs += [
         "-proc:only", // Only run the processor


MartinaeyNL:
Ready for a first review!

In the meantime, I'll test this a bit more extensively, and make sure no obvious bugs arise.
Anyone should feel free to test themselves as well, and provide a first code review 😉

wborn:
Why does the build fail @copilot ?

MartinaeyNL:
Resolved the inconsistencies in the UI tests, so it's ready for a proper review now 👍

ui/component/or-rules/src/json-viewer/or-rule-action-attribute.ts
```diff
@@ -103,50 +107,57 @@ export class OrRuleActionAttribute extends translate(i18next)(LitElement) {
...
+
+        // Set list of displayed assets, and filtering assets out if needed.
+        // If <= 25 assets: display everything
+        // If between 25 and 100 assets: display everything with search functionality
+        // If >= 100 assets: only display if in line with search input
+        const assets: Asset[] = this._cache ? this._cache.assets : [];
+        const searchable = assets.length > 25;
+        if(searchable && this._selected) {
+            idOptions.set(this._selected.id!, this._selected.name!);
+        }
```
Ekhorn, line 129: The `or-rule-asset-query` rendering logic does essentially identical things, but with a different query which we can pass as an argument.

I would prefer this wrapped in into a separate function including the logic around when you show the `searchProvider` and the provider itself.

Though not sure exactly what those functions would be.

ui/component/or-rules/src/json-viewer/or-rule-action-attribute.ts
```diff
@@ -238,9 +249,48 @@ export class OrRuleActionAttribute extends translate(i18next)(LitElement) {
...
-        })
+    /**
+     * Fetches assets using the {@link assetProvider} from the parent component.
+     * This is often linked to the OpenRemote HTTP API to request assets from using an {@link AssetQuery} object.
+     * @param type - The asset type name to filter by
+     * @param search - The asset name to filter by (acts as a search)
+     * @param idValue - Selected asset ID to query along
+     * @protected
+     */
+    protected async loadAssets(type: string, search?: string, idValue?: string): Promise<Asset[] | undefined> {
```
Ekhorn, line 260: The `loadAssets` method could be instead a function reused between `or-rule-action-attribute` and `or-rule-asset-query`.

```suggestion
/**
 * Fetches assets using the {@link assetProvider} from the parent component.
 * This is often linked to the OpenRemote HTTP API to request assets from using an {@link AssetQuery} object.
 * @param type - The asset type name to filter by
 * @param search - The asset name to filter by (acts as a search)
 * @param idValue - Selected asset ID to query along
 * @protected
 */
function async loadAssets(type: string, provider: AssetProvider, search?: string, idValue?: string): Promise<Asset[] | undefined>
```

We can make these components implement the following interface and then pass `this` to the function.

```ts
export interface AssetProvider {
   provider: (type: string, query?: AssetQuery) => Promise<Asset[] | undefined>,
   cache: {query: AssetQuery, assets: Asset[]};
   loading: boolean
}
```

The only issue with this might be that the `cache` access scope might need to change which seems like a deal breaker.

Inheritance would then I guess be better.. let's discuss?

Ekhorn (commented):
Looking good. I would prefer to extract and share some of the logic, but let's discuss.

ui/component/or-rules/src/json-viewer/or-rule-action-attribute.ts
MartinaeyNL: Inheritance would indeed make sense here.
Because, like you mentioned, you could make a child class with the `assetProvider` property,
and copy the function from here. The `cache` variable can then be passed through a parameter,
instead of referencing the class.

But I think it's too overdone to apply inheritance for this single function,
while the main use case of this function is to update the `cache` properly.
The `assetProvider` (or logic to fetch the assets) is already located outside of this function.
Moving logic to more different files would only make it more difficult to maintain.

As long as I don't duplicate it more often, this is OK currently.

ui/component/or-rules/src/json-viewer/or-rule-action-attribute.ts
MartinaeyNL: Yes the `or-rule-asset-query` and `or-rule-asset-attribute` classes are very similar.
And lots of the code is copied over from one another 😂 

Similar to my other comment on your review, I think it's too much of a mess to apply inheritance at this stage.
Having a base class would make the most sense, and should've been the approach from the start.
At this stage, it would only make the code more complex, and less readable.

ui/component/or-rules/src/json-viewer/or-rule-asset-query.ts
```diff
@@ -911,9 +926,48 @@ export class OrRuleAssetQuery extends translate(i18next)(LitElement) {
...
+            const assets = responses.filter(value => !!value).flat();
+            const cachedIds = this._cache?.assets.map(asset => asset.id) ?? [];
+            this._cache = {
+                query: query,
+                assets: [...(this._cache?.assets ?? []), ...(assets?.filter(a => !cachedIds.includes(a.id)) ?? [])]
+            };
+            return assets;
+        }
+        return this._cache?.assets;
     }
```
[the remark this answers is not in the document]
MartinaeyNL: (see the comments I placed under Ekhorn his review)

Ekhorn (approved):

PR #2272, opened 2025-11-20
Filter repo.osgeo.org content

wborn:
This makes Gradle only download the artifacts we use from this repo. Filtering repo content prevents potential security issues. It should also make builds with empty caches faster as it will use Maven Central and its CDN immediately for downloading most artifacts. This should also reduce the number of download issues (Bad gateway) we have with this Maven repository.

## Checklist

### DIFF
--- a/project.gradle
+++ b/project.gradle
@@ -170,6 +170,11 @@ repositories {
     // Needs to be above central due to org.geotools:gt-main issues
     maven {
         url = "https://repo.osgeo.org/repository/release/"
+        content {
+            includeGroup "javax.media"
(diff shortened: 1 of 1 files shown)


ebariaux:
An alternative / addition to this could be that we always use our own artefacts repo (on AWS) that caches public packages and also hosts our own custom ones.

wborn:
Yes there are other ways but this seemed the least effort. :wink:
Another alternative/addition could be to test using GitHub Actions to cache/restore the Gradle cache.

ebariaux (approved):

PR #2279, opened 2025-11-24
Logging config enhancements for FileHandler

richturner:
Adds the ability to configure the logging FileHandler to get the logfile output path from an environment variable.

Introduced `LoggingFileHandler` which replaces `or.logging.dir` string with `OR_LOGGING_DIR` environment variable.

### DIFF
--- a/container/src/main/java/org/openremote/container/util/LogUtil.java
+++ b/container/src/main/java/org/openremote/container/util/LogUtil.java
@@ -32,7 +32,6 @@ import java.util.Locale;
 import java.util.logging.LogManager;
 
 import static java.lang.System.Logger.Level.INFO;
-import static org.openremote.model.Container.OR_DEV_MODE;
 import static org.openremote.model.util.Config.OR_DEV_MODE;
 
 /**


container/src/main/java/org/openremote/container/Container.java
```diff
@@ -124,6 +124,9 @@ public Container(Map<String, String> config, Iterable<ContainerService> services
         SCHEDULED_EXECUTOR = new ContainerScheduledExecutor("ContainerScheduledExecutor", scheduledExecutorThreads);
         EXECUTOR = new ThreadPoolExecutor(executorThreadsMin, executorThreadsMax, 60L, TimeUnit.SECONDS, new SynchronousQueue<Runnable>(), new ContainerThreadFactory("ContainerExecutor"), new ThreadPoolExecutor.CallerRunsPolicy());
 
+        LOG.log(INFO, EXECUTOR);
+        LOG.log(INFO, SCHEDULED_EXECUTOR);
+
```
MartinaeyNL, line 122: Is this logging statement intentional?

richturner: Yes the `toString()` implementations of these classes provides a nicely formatted string with details of the thread pool which is useful to know.

MartinaeyNL:
I'm happy to approve this. However, it looks like a compile error came through (see CI/CD)

MartinaeyNL (approved):
Thanks!

PR #2280, opened 2025-11-24
Metrics improvements

richturner:


.ci_cd/aws/cloudformation-create-dashboard.yml
```diff
@@ -42,12 +42,15 @@ Resources:
               "type": "metric",
               "properties": {
                 "metrics": [
+                  [ { "expression": "m8-m6-m5-m4-m3-m2-m1", "label": "Other", "id": "e1", "region": "eu-west-1" } ],
```
denniskuijs, line 45: `m7` is not included here. Is this intended?

.ci_cd/aws/cloudformation-create-dashboard.yml
```diff
@@ -42,12 +42,15 @@ Resources:
...
                   [ "CWAgent-Prometheus", "or_attributes_total", "InstanceName", "a", "source", "RulesEngine", { "id": "m1", "yAxis": "left", "label": "Rules" } ],
                   [ "...", "AgentService", { "id": "m3", "yAxis": "left", "label": "Agents" } ],
                   [ "...", "DefaultMQTTHandler", { "id": "m2", "yAxis": "left", "label": "MQTT" } ],
```
denniskuijs, line 47: Sorry for my perfectionism! Can you move `m2` above `m3` so the sequence is logically ordered?

denniskuijs (changes requested):

.ci_cd/aws/cloudformation-create-dashboard.yml
richturner: Yes `m7` is an incompatible metric here we are using the total no. of attribute events and substracting the known event sources to determine the `other` count.

.ci_cd/aws/cloudformation-create-dashboard.yml
richturner: This is how it comes out of the dashboard editor in AWS but I've adjusted it for you 😄

PR #2286, opened 2025-11-26
Improved asset selection on Insights UI for scaling to larger instances 

### LINKED ISSUE #2285: Configuring a Map or Table widget slow with 50.000+ assets
### Description
When configuring either the **Map**- or **Table** widget inside an OpenRemote instance of 50.000+ assets,
the modal of selecting the asset becomes quite slow. This is because it tries to load **all** assets of a specific type,
which is a very large payload to request.

### To Reproduce
1. Run an OpenRemote Manager instance with 50.000+ assets.
2. Go to the Insights page
3. Create a new Dashboard with a Table (or Map) widget.
4. Select the widget to configure it, and turn off "All assets of this type".
5. Click on the button "0 assets" below it, which opens the modal to multi-select assets
6. It now takes a long time until any assets appear.

### Expected behavior
The modal should load and show at least a small amount of assets within 1 second.

### Additional context
The simplest solution to this is by applying pagination similar to the Asset Tree.
So it would initially show 100 assets by default, and show a "Load more" button to load the rest.
However, there is no search field, so looking for a specific asset is currently difficult.
Along with that, do we load the full tree (so all asset types), or do we only load 1 asset type without any structure.

MartinaeyNL:
## Description
I've simplified and improve the asset selection for the Map- and Table widgets on larger OpenRemote instances.
Because, as described in the original issue, the selector could not handle more than 1000 assets. (both UX and performance)
Now, the default asset tree is used, and a search bar has been added.

This was also the most suitable approach after I've discussed it with @remyberden.

Fixes #2285 

## Changelog
- Applied map clustering to the Map widget; it now has the same behavior of the Map page.
- Removed custom data provider for asset tree selector in the Insights UI, now showing default asset tree.
- Added search to the asset tree selector in the Insights UI, without the filter options, to the asset selector.

## Checklist

### DIFF
--- a/ui/component/or-asset-tree/src/index.ts
+++ b/ui/component/or-asset-tree/src/index.ts
@@ -1271,9 +1271,9 @@ export class OrAssetTree extends subscribe(manager)(LitElement) {
 
         try {
             console.debug(`Querying assets using filter '${this._filterInput.nativeValue}'...`);
-            const promises = assetQueries.map(q => manager.rest.api.AssetResource.queryAssets(q));
+            const promises = assetQueries.map(q => manager.rest.api.AssetResource.queryAssetTree(q));
             const responses = await Promise.all(promises);
-            foundAssets = responses.flatMap(r => r.data);
+            foundAssets = responses.flatMap(r => r.data.assets ?? []);
             foundAssetIds = foundAssets.map(a => a.id!);
             console.debug(`The filter query found ${foundAssets.length} assets!`);
 
--- a/ui/component/or-asset-viewer/src/index.ts
+++ b/ui/component/or-asset-viewer/src/index.ts
@@ -940,6 +940,7 @@ async function getAssetChildren(parentId: string, childAssetType: string): Promi
 
     try {
         response = await manager.rest.api.AssetResource.queryAssets({
+            types: [childAssetType],
             parents: [
                 {
                     id: parentId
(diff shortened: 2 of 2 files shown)


MartinaeyNL:
Turned out to be a relatively simple change! 👍 
Ready for review.

MartinaeyNL:
Converting this back to draft, after I found some performance issues during testing.

MartinaeyNL:
Resolved the performance issues for the Map widget.
It now uses Map marker clustering, like on other places, which greatly improves the UX and browser performance.

Ready to review again 👍

ui/component/or-dashboard-builder/src/widgets/map-widget.ts
```diff
@@ -83,131 +83,128 @@ export class MapWidget extends OrAssetWidget {
...
+                    };
+                    break;
+                }
+                case "text": {
+                    marker.colours = this.widgetConfig.textColors.reduce((colors: any, [key, val]) => {
+                        colors[key] = val.replace("#", "");
+                        return colors;
+                    }, { type: "string" });
+                    break;
+                }
```
Ekhorn, line 186: The text option can't be saved.

Looking on test4 which was last deployed to nov 18, this also occurred there, but I guess it can't hurt to fix this or at least write an issue.

Ekhorn: Created an issue https://github.com/openremote/openremote/issues/2296

Ekhorn (approved):

PR #2304, opened 2025-12-03
Fix load2 Java compilation errors

wborn:
* Fix load2 imports
* Remove redundant load2 imports
* Make sure Java code is always build to prevent this from happening
* Fix Gradle deprecation


Ekhorn (approved):
Thanks!

Issue #2197, opened 2025-10-17
Move Energy related code module to separate repo


ebariaux:
## User Story
As a development team, I want to extract a modular component from our main repository into a standalone repository, so that we can improve code maintainability, reduce dependencies, and enable independent versioning.

## Description
The first step was to extract all energy related code into a separate module but still within the main repository, this has been covered in: #1726 

Phase 2 - Repository Extraction:
- Move the fully decoupled module to a separate repository
- Set up CI/CD pipeline for automated artifact publishing
## Acceptance Criteria

Phase 2 - Repository Extraction:
- [ ] 1. Create a new GitHub repository for the module
- [ ] 2. Migrate module code to the new repository
- [ ] 3. Set up Maven/Gradle build configuration
- [ ] 4. Implement CI/CD pipeline for automated testing
- [ ] 5. Configure artifact publishing to Maven
- [ ] 6. Create README and documentation
- [ ] 7. Establish versioning strategy
- [ ] 8. Perform integration testing with main repository

## Implementation Notes

Phase 2 Considerations:
- Use GitHub Actions for CI/CD
- Publish to Maven
- Implement semantic versioning
- Create release workflows
- Set up automated testing for each pull request

PR #1412, opened 2024-09-12
Updated forecast solar service

igorrutka:
Changelog Solar forecast Service:

- Improved logging
- Updating input attributes, needed for the solar forecast, updates the forecast without the need to unchecking and checking the "Include forecast solar service" checkbox
- "Set actual solar power value with forecast" checkbox updates the power attribute also on button click
- "Power forecast" attribute updates every 15-minutes at 00,15,30,45 minutes and stores the correct forecast values

Solar forecast:
- Forecast updates 1 once an hour at the beginning of a new hour
- Only forecast values after current time are updated
- 0 values at nighttime are now at constant intervals of 00,15,30,45 minutes
- Correct handling of transition between nighttime 0 values to non-zero sunlight values

PR #2309, opened 2025-12-04
Move energy code into separate extension

### LINKED ISSUE #1726: Move Energy related code to extensions repo
## User Story
As a development team, I want to extract an extension from our main repository into a separate extension and eventually a standalone repository, so that we can improve code maintainability, reduce dependencies, and enable independent versioning.

Phase 2 is covered in #2197 

## Description
Our extraction strategy will follow a two-phase approach:

Phase 1 - Internal Modularization:
- Create a separate extension within the main repository
- Remove all direct dependencies to the main repository code
- Use only the main repository's published artifact
- Ensure complete decoupling and standalone functionality
- jar from this module is published just as other "main jars" (e.g. model)
- jar is included in built docker image

This approach minimizes risk and allows for incremental, controlled refactoring.

## Acceptance Criteria
Phase 1 - Internal Modularization:
- [ ] 1. Use energy related code(also tests) as first extension to isolate
- [ ] 2. Create a new extension in the main repository
- [ ] 3. Refactor code to remove internal dependencies
- [ ] 4. Use only main repository's published artifact
- [ ] 5. Verify all tests pass with new extension structure
- [ ] 6. Conduct code review to ensure complete decoupling

## Implementation Notes
Phase 1 Considerations:
- Also migrate tests of energy related code
- Minimize scope of extension
- Ensure clear separation of concerns

wborn:
Moves energy related code (assets, services, tests) into a separate extension.

Fixes #1726

TODO:

- [x] Fix Demo setup dependency issues
- [x] Agree upon naming (~add-on~/~bundle~/**extension**/~module~/~plug-in~/~block~/~pack(age)~/~component~/~kit~/~unit~/~integration~?)
- [x] Test if everything still works
- [x] Test if code is self-contained by removing code and using published artifacts


wborn:
The Demo setup has a dependency on the Energy code so I also extracted it as an extension. We could do something similar for the load tests.

To test that the code is self contained you can remove the `extensions` dir and it will still all build correctly.

I also made a small PoC to see if the extensions code can be extracted to a separate repo (#2197).

To test this:

1. Build this branch and publish it to your local Maven repository using:
  `gradle clean installDist publishToMavenLocal`
2. Clone https://github.com/wborn/extensions (based on the extensions dir):
  `git clone git@github.com:wborn/extensions.git`
3. Build the extensions:
  `cd extensions`
  `gradle clean build`

extension/demo-setup/src/main/java/org/openremote/extension/demosetup/DemoSetupTasks.java
wborn, line 1: We might also want to update the Java package names. WDYT?

wborn:
> I think it makes sense, but I see no energy assets anymore in the Empty intellij profile. Is this intentional?

IIRC having dependencies on extensions in the manager would introduce dependency cycles. That is because the extensions depend on manager APIs.

If you change the classpath in the run configuration to the  energy extension, the energy assets will show again:

Screenshot from 2025-12-12 15-13-03

project.gradle
```diff
@@ -136,7 +136,7 @@ if (project == rootProject) {
                     workingDirectory = projectDir.toString()
                 }
                 "Demo Setup"(Application) {
-                    moduleName = getProject().idea.module.name + ".setup.demo"
+                    moduleName = getProject().idea.module.name + ".extensions.demo-setup.main"
```
pankalog: Bit of a conceptual question here; shouldn't all extensions automatically be enrolled as run configurations? Since "including the source code" of each extension essentially requires changing the classpath, shouldn't we register every extension as a run configuration in IntelliJ? This could help with easily debugging what is/isn't in an extension.

extension/demo-setup/src/main/java/org/openremote/extension/demosetup/DemoSetupTasks.java
pankalog: I would say that the code of extensions should be moved to the `org.openremote.extension` package.

project.gradle
wborn: Yes it should be easy to configure what extensions are used during development. How that should work is something to discuss in https://github.com/openremote/openremote/issues/2327. In this PR I just made sure the existing functionality keeps working. Perhaps Gradle can also resolve the active extension dependencies based on what is configured in a property/env var/file.

extension/demo-setup/src/main/java/org/openremote/extension/demosetup/DemoSetupTasks.java
ebariaux: I was first confused with the folder called extensions (with s) and the package extension (no s).
Any reason for the discrepancy ?

extension/energy/build.gradle
```diff
@@ -0,0 +1,90 @@
...
+apply plugin: "signing"
+
+base {
+    archivesName = "openremote-${project.name}-extension"
+}
+
+dependencies {
+    api resolveProject(":manager")
+    api resolveProject(":model")
+
```
ebariaux, line 16: Dependencies such as camel or jackson used in the code should be explicitly defined in the module and not rely on transitivity from other jars

extensions/energy/src/test/groovy/org/openremote/extension/energy/ForecastSolarServiceTest.groovy
```diff
@@ -191,17 +186,11 @@ class ForecastSolarServiceTest extends Specification implements ManagerContainer
         newSolarAsset2.setIncludeForecastSolarService(false);
         newSolarAsset2 = assetStorageService.merge(newSolarAsset2)
 
-        then: "the assetId shouldn't be present in the calculationFutures"
-        conditions.eventually {
-            assert forecastSolarService.calculationFutures.get(newSolarAsset2.getId()) == null
-        }
-
-        when: "an asset updated it's includeForecastSolarService to true"
+        and: "an asset updated it's includeForecastSolarService to true"
```
ebariaux:
```suggestion
        and: "an asset updated its includeForecastSolarService to true"
```

extension/energy/src/test/groovy/org/openremote/extension/energy/ForecastSolarServiceTest.groovy
```diff
@@ -144,10 +142,8 @@ class ForecastSolarServiceTest extends Specification implements ManagerContainer
         def assetProcessingService = container.getService(AssetProcessingService.class)
         def forecastSolarService = container.getService(ForecastSolarService.class)
 
-        expect: "a Future for calculation should exist and the asset should have filled in value for power and powerForecast"
+        expect: "the asset should have filled in value for power and powerForecast"
         conditions.eventually {
```
ebariaux, line 146: I don't know enough about the business logic underneath but why remove some of those tests ?

wborn: This test doesn't run because it is unstable; the class got annotated with `@Ignore` in https://github.com/openremote/openremote/commit/c60dfa273aac1723a94543b75c749464c6512f03.

I  tested if the test would run when `@Ignore` is removed. Then I noticed there were compilation failures because the `calculationFutures` got removed in https://github.com/openremote/openremote/pull/1412 so I updated the test for this. It's probably still unstable though.

extension/demo-setup/src/main/java/org/openremote/extension/demosetup/DemoSetupTasks.java
wborn: Now it's more consistent and using `extension` everywhere.

ebariaux (approved):
LTGM

MartinaeyNL:
We were waiting for a new release (1.13.0), which went live about an hour ago.
So let's merge this 💪 

I got approval from @wborn to press the button haha

PR #2335, opened 2025-12-10
Update codeql-action/upload-sarif to version 4.31.7

denniskuijs:


wborn (commented):
Perhaps also update the [anchore/scan-action](https://github.com/anchore/scan-action) because that action generates the JSON that is uploaded by the upload-sarif action.

PR #2339, opened 2025-12-11
Free up disk space on the GitHub runner

denniskuijs:


denniskuijs:
@wborn I see that Copilot is added as a co-author. To avoid discussion, we should remove it from the commit description, if possible.

PR #2356, opened 2025-12-17
Backend test robustness improvement

richturner:
This improves the container re-use logic and merges all agents before other assets; it also handles the scenario where agents are decendents of other assets by ensuring those ancestors are merged first.

### DIFF
--- a/container/src/main/java/org/openremote/container/web/CORSConfig.java
+++ b/container/src/main/java/org/openremote/container/web/CORSConfig.java
@@ -118,7 +118,11 @@ public class CORSConfig {
         String allowedOriginsStr = Config.getString(OR_WEBSERVER_ALLOWED_ORIGINS, null);
 
         if (!TextUtil.isNullOrEmpty(allowedOriginsStr)) {
-            allowedOrigins.addAll(Arrays.stream(allowedOriginsStr.split(",")).toList());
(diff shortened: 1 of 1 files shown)


test/src/main/groovy/org/openremote/test/ContainerTrait.groovy
```diff
@@ -219,8 +219,40 @@ trait ContainerTrait {
...
+                            LOG.info("Re-inserting ${agents.size()} agents(s)")
+                            agents.forEach { a ->
+                                a.version = 0
+
+                                if (a.path.size() > 1) {
+                                    for (final def assetId in a.path) {
+                                        if (assetId == a.id) {
+                                            continue
+                                        }
+                                        def ancestorIndex = assets.findIndexOf {it.id == assetId}
```
ebariaux: Coherent with spacing above
```suggestion
                                        def ancestorIndex = assets.findIndexOf { it.id == assetId }
```

wborn:
Will it also close any of the [open test issues](https://github.com/openremote/openremote/issues?q=is%3Aissue%20state%3Aopen%20label%3ATest) @richturner?

richturner:
Difficult to say for certain but it should help unless there is some other timing issue that needs addressing, lets keep those items open for a while

PR #2366, opened 2025-12-17
Hotfix for wildcard CORS configuration when OR_HOSTNAME and/or OR_ADDTIONAL_HOSTNAMES also set

richturner:


container/src/main/java/org/openremote/container/web/CORSConfig.java
```diff
@@ -118,7 +118,11 @@ public static Set<String> getCorsAllowedOriginsFromConfig() {
         String allowedOriginsStr = Config.getString(OR_WEBSERVER_ALLOWED_ORIGINS, null);
 
         if (!TextUtil.isNullOrEmpty(allowedOriginsStr)) {
-            allowedOrigins.addAll(Arrays.stream(allowedOriginsStr.split(",")).toList());
+            if (allowedOriginsStr.equals(DEFAULT_CORS_ALLOW_ALL)) {
```
ebariaux, line 121: You're testing for equality, so what if allowedOriginStr is something like "*, xyz" ?
Would not match this and still result in an allow all + something else

ebariaux: Maybe you want to split first and check if any of the elements is the ALLOW_ALL ?

richturner: The main point here is to handle wildcard configuration any other value that is not a valid URI will fail as expected

richturner: So `OR_WEBSERVER_ALLOWED_ORIGINS` can be a wildcard `*` or comma separated list of valid URIs

ebariaux: Was wondering if we wanted to handle case of `OR_WEBSERVER_ALLOWED_ORIGINS` being a comma separated list of valid URIs and also include `*`.
In that case, do we want to default to just `*` ?

ebariaux: Discussed and this is considered an invalid configuration, we don't automatically fallback to only wildcard in that scenario.

richturner: As discussed we can always revisit this in future

PR #2385, opened 2026-01-06
UI bugfix for fields being cleared while on the user creation page

### LINKED ISSUE #2244: Fields are cleared when switching browser tabs during user creation
**Describe the bug**
When switching between browser tabs during user creation, some input fields such as `username`, `e-mail` and `realm_roles` are cleared.

**To Reproduce**
1. Go to an OpenRemote Manager.
2. Navigate to the **Users** page and click on **Add User**.
3. Fill in all the required fields.
4. Switch to another browser tab.
5. Go back to the OpenRemote Manager.
6. Several fields are cleared and lose their values.

**Expected behavior**
All values should be visible after returning from a different browser tab.

**Screenshots**
Image

MartinaeyNL:
## Description
This PR fixes #2244, where the fields on the `/users` page of the Manager UI were cleared after switching tabs.
In the code there was a statement that reset the `creationState` variable, causing all previously entered details to be removed.
Now, I only check if there was a change in user type (`serviceAccount` yes/no), since that was the original use case for it.

## Changelog
- Bugfix for fields on user page being reset when the internal state updates.
This is caused for example by switching browser tabs.

## Checklist

### DIFF
--- a/ui/app/manager/test/users.test.ts
+++ b/ui/app/manager/test/users.test.ts
@@ -78,7 +78,7 @@ test(`Verify browser behavior while creating regular users`, async ({ page, user
     await usersPage.toHavePermissions(...permissions);
     await page.evaluate(() => {
         document.dispatchEvent(new Event('visibilitychange'));
-    })
+    });
     await expect(await page.getByRole('textbox', {name: /username/i }).inputValue()).toBe("mycustomusername");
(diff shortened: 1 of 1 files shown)


wborn (approved):
Thanks for fixing this annoying bug!

PR #2396, opened 2026-01-08
Only close gateway tunnels for interconnect connection status change in the same realm

richturner:


manager/src/main/resources/logging.properties
```diff
@@ -79,6 +79,7 @@ org.apache.camel.impl.converter.DefaultTypeConverter.level=WARNING
 org.apache.camel.impl.DefaultShutdownStrategy.level=SEVERE
 org.jeasy.rules.level=WARNING
 org.apache.activemq.audit.level=WARNING
+io.netty.level=SEVERE
```
MartinaeyNL, line 82: Double check if this severity change is intentional?

richturner: Yes netty outputs a full stack trace at WARNING which is very verbose and we already log enough info on connection failures

richturner:
Thanks for the review @MartinaeyNL LGTM and further updates can be handled in other PRs

PR #2420, opened 2026-01-14
UI fixes for Insights page with inconsistencies in attribute order

### LINKED ISSUE #2125: Line Chart attributes order issue
**Describe the bug**
In the line chart legend, the order of the attributes is inconsistent when modifying the figure. 

**To Reproduce**
1. Go to the insights page
2. Create a line chart and add 1 attribute and click 'save'
4. Modify the line chart again by adding a second attribute and click 'save'
5. Repeat the above step multiple times and see how the order of attributes changes. Some of the attributes are rearranged in alphabetical order but others are not. This inconsistency combined with the color picker issue https://github.com/openremote/openremote/issues/2123 makes setting a consistent color theme through multiple line charts a tedious process. 

**Expected behavior**
The second attribute should be visible.

MartinaeyNL:
## Description
This PR resolves the inconsistencies in how attributes are sorted on the Insights page.
Before, they were sometimes based on when the attribute was added onto the widget,
and sometimes they were alphabetically sorted.

This is now made consistent; so all attributes on the Insights page will always be alphabetically sorted
based on the label of the attribute. (either through meta items, or the attribute name itself)

Fixes #2125 

## Changelog
- All widget settings panels with an "add attribute" window now has alphabetically sorted attributes.
- Chart widget legend now has alphabetically sorted attributes.

## Checklist


ui/app/manager/test/insights.test.ts
```diff
@@ -0,0 +1,75 @@
...
+import {preparedAssetsForInsights as assets} from "./fixtures/data/assets.js";
+import { expect } from "@openremote/test";
+
+test.use({storageState: userStatePath});
+
+test("Create a Line Chart widget", async ({manager, shared, page, insightsPage}) => {
+
+    await shared.interceptResponse<Dashboard>("**/dashboard", (dashboard) => {
+        if (dashboard) manager.dashboards.push(dashboard.id!);
+    });
```
Ekhorn, line 12: This is a little inconsistent with how we do rest of our setup in `manager.setup`. I would consider moving it into there if that isn't an issue. If you have better ideas for `manager.setup` please consider sharing them, so we can pick that up as it currently doesn't always feel as flexible.

ui/app/manager/test/insights.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    await expect(insightsPage.getWidgets()).toHaveCount(0);
+    await insightsPage.dragAndDropWidget("Line Chart");
+    await expect(insightsPage.getWidgets()).toHaveCount(1);
+
+    // Make widget larger by resizing it using the handle
+    const chartWidget = insightsPage.getWidgets({ hasText: "Line Chart" }).first();
+    const box = await chartWidget.boundingBox();
+    const startX = box!.x + box!.width - 4;
+    const startY = box!.y + box!.height - 4;
+    await page.mouse.move(startX, startY);
```
Ekhorn: Just checking if you have tried the `hover` API?

```suggestion
    await page.hover("handle-element-example");
```

ui/app/manager/test/insights.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    await expect(insightsPage.getWidgetAttributes().getByText("Power")).toHaveCount(1);
+    await expect(insightsPage.getWidgetAttributes()).toHaveCount(3);
+
+    // Save the dashboard
+    await page.route("**/dashboard", async (route, request) => {
+        await route.continue();
+        const response = await request.response();
+        expect(response?.status()).toBe(200);
+        expect(await response?.json()).toBeDefined();
+    });
```
Ekhorn: Is it not possible to use the shared fixture for intercepting a response? Which you have used above? I guess we need to return the response object and not the payload.

```suggestion
   await shared.interceptResponse<Dashboard>("**/dashboard", (response => {
        expect(response?.status()).toBe(200);
        expect(await response?.json()).toBeDefined();
    });
```

ui/app/manager/test/insights.test.ts
Ekhorn: I see you have in one of your fixtures.

ui/app/manager/test/insights.test.ts
Ekhorn: The whole gathering of DB changes to clean up is a mess currently I'll pick it up at a later stage.

ui/app/manager/test/insights.test.ts
MartinaeyNL: I want to move the cursor to a specific coordinate (or offset to the bottom left) so it can drag the widget larger.
The `hover` API does not support this as far as I know.
You're correct that I do use it at a few other places.

ui/app/manager/test/insights.test.ts
MartinaeyNL: `interceptResponse` didn't work here, as I need the `Response` object instead of the JSON object, for checking `OK 200`.
See here the function inside the `shared.ts` file:
```typescript
async interceptResponse<T>(url: string, cb: (body?: T) => void) {
        await this.page.route(
            url,
            async (route, request) => {
                await route.continue();
                const response = await request.response();
                cb(await response?.json()); // Returns the JSON, not the Response object I'm looking for to make sure 200 is returned.
            },
            { times: 1 }
        );
    }
```

ui/app/manager/test/insights.test.ts
MartinaeyNL: I copied over / inspired most of the code based on other files.
For me, the `manager.setup()` function does not have any interceptors that it defines.
The `assets` fixture file does however.

We discussed this in person, and we agreed to improve the `shared` fixtures in a separate issue.

ui/app/manager/test/insights.test.ts
Ekhorn: Yes so we can modify it to return the actual response rather than the body IMO :)

MartinaeyNL: Done.

MartinaeyNL:
Processed the requested changes. Feel free to take a look @Ekhorn :)

ui/app/manager/test/insights.test.ts
MartinaeyNL: Ended up replacing it with "hovering the handle" as you requested 😉 
My bad for not reading your comment correctly haha

Issue #2389, opened 2026-01-07
Reduce manager app bundle size


Ekhorn:
## User Story
As a [user] I want the manager app bundle size to be reduced, so that it loads faster on slower connections.

## Description

The manager app bundle size is currently 7.40 MB due too several libraries that bloat the bundle.

<img width="727" height="52" alt="Image" src="https://github.com/user-attachments/assets/0e88de85-6dfb-462b-8288-5c64be524307" />

Other options could be to apply tree-shaking.

This should also help address https://github.com/openremote/openremote/issues/1988

I expect this issue to be a longer ongoing effort, which we may want to split in smaller issues.

## Acceptance Criteria

> Important: all changes must indeed reduce the bundle size, use Rsdoctor and check the bundle when loading the page. Also of course check for breaking changes.

- [ ] 1. Change [`moment.js`](https://bundlephobia.com/package/moment@2.29.4) to [`luxon`](https://bundlephobia.com/package/luxon@3.7.2) or [`dayjs`](https://bundlephobia.com/package/dayjs@1.11.19) (intermediate step https://github.com/openremote/openremote/pull/2429)
- [x] 2. Replace [`moment-timezone`](https://bundlephobia.com/package/moment-timezone@0.6.0) with standard browser API (ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf) ~~is bundled correctly (it should only be adding 3kb (gzip), ref https://momentjs.com/timezone/).~~ #2428 
- [ ] 3. Get rid off (entire) lodash imports
- [x] 4. Minify our Lit CSS & HTML string literals (ref https://github.com/openremote/openremote/pull/2435)
- [ ] #2537
- [x] 6. We can remove https://github.com/openremote/openremote/issues/2240

## Implementation Notes

The following PR introduced tooling to figure out the app bundle size, see https://github.com/openremote/openremote/pull/2345 (this has now been merged by #2426)

<img width="2436" height="490" alt="Image" src="https://github.com/user-attachments/assets/f84dc1f9-c0cb-4e33-95e5-7768af08f599" />

Ekhorn:
So far we shaved of 1MB from the manager bundle.

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/44a0d5eb-0f03-4200-a365-11361f70f465" />

[...]

PR #2430, opened 2026-01-19
UI: Replace global lodash dependency with individual packages

MartinaeyNL:
## Description
This PR removes our import of the `lodash` package, which ships a ton of functions we don't use.
Instead, I opt to use the individual packages, decreasing the amount of code we depend on.
Unfortunately it doesn't decrease our package size, as `@jsonforms/core` requires it, but it might help in the future.

- Ref #2389 

## Checklist
- [x] ~~2. Tests are written and all tests pass~~

### DIFF
--- a/ui/component/core/package.json
+++ b/ui/component/core/package.json
@@ -35,7 +35,7 @@
     "url-search-params-polyfill": "^8.1.0"
   },
   "devDependencies": {
-    "@types/lodash.transform": "^4.6.9",
+    "@types/lodash.transform": "^4",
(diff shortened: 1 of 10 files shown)


Ekhorn (changes requested):
`or-survey` still has a dependency on `lodash-es`.

Then I think we should attempt tree-shaking `jsonforms/core`.

I see `html-webpack-plugin` (and its transitive deps also have: `pretty-error`, `renderkid`) a dependency on `lodash`, but of course this shouldn't matter for the bundle size.

PR #2452, opened 2026-01-23
Fix default text color example not being shown

### LINKED ISSUE #2296: The map marker config 'text' option in the map widget doesn't work
**Describe the bug**

The map marker config in the map widget can't be saved nor are the default example values added.

**To Reproduce**
1. Add a light asset and add a location
2. Add map widget
3. Select light asset type and the notes attribute
4. Doesn't show the default example values nor does it allow saving the text marker config

**Expected behavior**

The examples to be added and the save button to store the config.

**Screenshots**

Image

**Additional context**

The default config contains the following https://github.com/openremote/openremote/blob/b0e642001d2af7198e1279a022ff1c644043e520/ui/component/or-dashboard-builder/src/widgets/map-widget.ts#L54

Ekhorn:
## Description

This PR ensures the default `textColors` configuration options (thresholds) are added once a text attribute is selected in the map widget. It also implements the missing handler for this option to propagate the changes to allow them to be saved.

I've removed the lock on the first threshold for the text colors options, to allow people to not always be stuck with at least 1 case where the marker color can change.

- Closes #2296 

## Changelog

- Fix default examples not shown for the map widget marker text config option
- Fix changes propagating changes for the map widget marker text config option
- Unlocks the first threshold for the marker text config option (to be removable)
- Change first example value from `example` to `example1`

## Checklist


ui/component/or-dashboard-builder/src/settings/map-settings.ts
Ekhorn, line 1: I don't feel this should be merged, because the first threshold value can't be removed, which doesn't make sense to me.

Image

Ekhorn: Made some changes I like it now.

ui/component/or-dashboard-builder/src/panels/thresholds-panel.ts
Ekhorn, line 1: I'll write a little test for this 😃

ui/app/manager/test/insights.test.ts
```diff
@@ -73,6 +78,105 @@ test("Create a Line Chart widget", async ({manager, shared, page, insightsPage})
...
+    // Update attribute to match the first threshold
+    await manager.sendWebSocketEvent({
+        eventType: "attribute",
+        ref: { id: manager.assets[0].id, name: "notes" },
+        value: "example1",
+    });
+    await page.getByTitle("Refresh", { exact: true }).click();
+    await expect(page.locator(".or-map-marker")).toBeVisible();
+
+    const first = await page.locator('or-icon[icon="or:marker"]').evaluate(getRGBColor);
```
Ekhorn: This is flaky, consider polling here.

```ts
    const first = await page.locator('or-icon[icon="or:marker"]').evaluate(getRGBColor);
    expect(rgbToHex(first)).toBe("4caf50");
```

ui/app/manager/test/fixtures/pages/insights.ts
```diff
@@ -40,36 +40,71 @@ export class InsightsPage implements BasePage {
...
 
         // Manual drag and drop
         const count = await this.getWidgets().count();
-        await this.page.locator("#sidebar .grid-stack-item", {hasText: type}).first().hover();
-        await this.page.mouse.down();
-        await this.page.mouse.move(-200, 0, { steps: 5 }); // Move tiny bit left, so the hover event is triggered
-        await this.page.locator(".maingrid").hover({ position: { x: coordinates[0], y: coordinates[1] }});
-        await this.page.mouse.up();
+        await this.page.dragAndDrop(`#sidebar .grid-stack-item :has-text("${type}")`, ".maingrid", {
+            sourcePosition: { x: 10, y: 10 }, // Depending on where you grab the widget the drop
```
MartinaeyNL: Maybe add a small clarification comment that this uses an "offset" of the element, not the raw screen/viewport position.

ui/app/manager/test/fixtures/pages/insights.ts
```diff
@@ -40,36 +40,71 @@ export class InsightsPage implements BasePage {
...
         await expect(this.getWidgets()).toHaveCount(count + 1);
     }
 
+    /**
+     * Resize a widget to the specified grid cell coordinates
+     * @param widget The locator of the widget
+     * @param from The grid cell coordinates of the widget
+     * @param to The grid cell coordinates to resize the widget to
+     */
+    async resizeWidgetAtTo(widget: Locator, [gridX, gridY] = [0, 0], [cellsWidth, cellsHeight] = [8, 8]) {
```
MartinaeyNL: Shouldn't `gridX` and `gridY` here be automatically set based on `widget`, instead of being a parameter?

As in, it feels weird to say `resize widget from X to Y`, while `X` should be the current position of the widget.
So preferably you'd get the coordinates using a function similar to `getGridCellDimensions()`.

ui/component/or-dashboard-builder/src/settings/map-settings.ts
MartinaeyNL, line 1: Can we hide the thresholds panel if it's empty?

Image

ui/app/manager/test/fixtures/pages/insights.ts
Ekhorn: Yeah forgot to properly update this comment thx, will do!

ui/app/manager/test/fixtures/pages/insights.ts
Ekhorn: Well if there is a Gridstack API I can use to determine the location of the current widget, then yes..

ui/component/or-dashboard-builder/src/settings/map-settings.ts
Ekhorn: Yep! done.

ui/app/manager/test/fixtures/pages/insights.ts
Ekhorn: Figured out a somewhat clean way finally.. The grid stack nodes have x and y coordinates we now evaluate those from the parent of the widget container element.

PR #2458, opened 2026-01-27
Use PostgreSQL latest-slim image

wborn:
This will speed up builds and deployments as the slim image is 4.5GB smaller. It will also prevent storage issues on VMs.

Depends on https://github.com/openremote/postgresql/pull/22

## Checklist

### DIFF
--- a/setup/src/load2/java/org/openremote/setup/load2/KeycloakSetup.java
+++ b/setup/src/load2/java/org/openremote/setup/load2/KeycloakSetup.java
@@ -89,8 +89,9 @@ public class KeycloakSetup extends AbstractKeycloakSetup {
 
         // Wait until all users created
         int waitCounter = 0;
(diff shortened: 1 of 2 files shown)


wborn:
> Doesn't this mean anyone who clones this repository, tests it out locally,
> or deploys it to a production instance using a clone, would use the slim image?
> 
> I would personally leave it out / remove it from the `docker-compose.yml` file,
> and only use it in the development profiles (`dev-testing.yml`, ...) and the `.env` files of our testing deployments.

The best practice is to always set `POSTGRESQL_VERSION` anyhow to keep deployments reproducible. 
When you set it you can decide between using the fat/slim image.

The latest image should only be used for testing. IMHO the slim image should work just as well as the fat image.
If not, we should not spend time on maintaining or releasing it.

PR #2475, opened 2026-01-28
Gateway API refactor and tunnel robustness improvements

richturner:
A summary of changes included in this PR:

- `JSchGatewayTunnelFactory` replaced by `MINAGatewayTunnelFactory` to provide a more robust SSH tunnelling client that automatically handles reconnection of SSH remote forwarding sessions
- Gateway instance client logic moved into its' own class `GatewayClientConnector` declutering `GatewayClientService` and avoiding the use of Maps with realm as key
- `setInitFutureSupplier` method added to `AbstractNettyIOClient` this makes `GatewayIOClient` obsolete
- `AbstractINettyOClient:startChannel()` changed to return `CompletableFuture<Void>` rather than `Future<Void>`
- Gateway API version introduced
- Gateway API enhancements - Forwards and backwards compatibility maintained for previous gateway versions (some deprecation, final removal to be scheduled after sufficient time to migrate)

### DIFF
--- a/manager/src/main/java/org/openremote/manager/gateway/GatewayClientConnector.java
+++ b/manager/src/main/java/org/openremote/manager/gateway/GatewayClientConnector.java
@@ -321,7 +321,7 @@ public class GatewayClientConnector implements AutoCloseable {
                     LOG.finest("Gateway tunnel creation request received but gateway tunnel factory is not available: realm=" + connection.getLocalRealm());
                     return;
                 }
-                if (tunnelHostname == null || tunnelPort == 0) {
+                if (tunnelHostname == null) {
                     // If we don't have tunnel hostname and port already then this is a legacy manager so try and get from
                     // the start request event - it never changes in legacy manager so we can safely do this once
                     // TODO: Remove this once enough time has passed since this commit was made


manager/src/main/java/org/openremote/manager/gateway/GatewayClientConnector.java
```diff
@@ -0,0 +1,692 @@
...
+                                if (Math.abs(value - oldValue) > Math.abs(delta)) {
+                                    LOG.finest(() -> "Gateway client for '" + connection.getLocalRealm() + "' delta setting has allowed attribute event: " + ev.getRef());
+                                    return true;
+                                }
+                            }
+                        }
+                        if (filter.getDurationParsed().isPresent()) {
+                            boolean allow = filter.getDurationParsed().map(durationMillis -> {
+                                Long lastSendMillis = attributeTimestamps.get(ev.getRef());
+                                if (lastSendMillis == null || timerService.getCurrentTimeMillis() - lastSendMillis > durationMillis) {
```
pankalog, line 600: The attributeTimestamps map is only initialized when connection.getAttributeFilters() is not null or empty, but the code can access it:

`if (attributeTimestamps != null && filter.getDurationParsed().isPresent()) {`

manager/src/main/java/org/openremote/manager/gateway/GatewayClientConnector.java
```diff
@@ -0,0 +1,692 @@
...
+                        LOG.log(Level.INFO, "Request from central manager to create/update an asset failed: Realm=" + connection.getLocalRealm() + ", Asset ID=" + asset.getId(), e);
+                    }
+                }
+            }
+            case GatewayTunnelStartRequestEvent startRequestEvent -> {
+                if (tunnelFactory == null) {
+                    LOG.finest("Gateway tunnel creation request received but gateway tunnel factory is not available: realm=" + connection.getLocalRealm());
+                    return;
+                }
+                if (tunnelHostname == null || tunnelPort == 0) {
```
pankalog: NPE if tunnelPort is null

`if (tunnelHostname == null || tunnelPort == null || tunnelPort == 0)`

manager/src/main/java/org/openremote/manager/gateway/GatewayClientConnector.java
richturner: Been through that one with @ebariaux if it gets to that part of the code then predicatesWithFilters is empty and so the filtering logic will not trigger

manager/src/main/java/org/openremote/manager/gateway/GatewayClientConnector.java
richturner: Although unlikely indeed the unboxing could trigger an NPE, will remove the port check as hostname check is sufficient

PR #2482, opened 2026-01-31
Fix console API 500 errors

ebariaux:


ebariaux:
Deployed on test2 and tested OK with iOS console.
@MartinaeyNL can you approve PR so I can merge ? Thx

PR #2512, opened 2026-02-09
or-scheduler: migrate to Vaadin UI

### LINKED ISSUE #2362: Migrate or-scheduler component to vaadin design
## User Story
As a developer I want the Vaadin design for the scheduler to be applied, so that we can close the chapter on this.

## Description

Discussed with @pierrekil and @remyberden to update this component to the new design in isolation.

## Acceptance Criteria
- [ ] 1. Introduce missing fields to Vaadin input
- [ ] 2. Replace all fields in `or-scheduler` with Vaadin input.
- [ ] 3. Update the dialog component to match with the new design.

## Implementation Notes

Will need to extract the relevant code from https://github.com/openremote/openremote/pull/2340

Ekhorn:
## Description

- Closes #2362

## Changelog

- The `or-scheduler` component is now using the new Vaadin UI
- The `manager` app now applies the default theme from `@openremote/theme`
- The `or-scheduler` dialog now has a configurable "delete schedule" button by setting `removable`
- Fix circular dependency in `@openremote/or-vaadin-components/or-vaadin-input` and `@openremote/or-vaadin-components/util` by moving the Value Input Provider code to a separate `value-input-provider.ts` file.
- Adds `or-scheduler` to Storybook
---
- **Breaking**: Renamed `or-vaadin-numberfield` to `or-vaadin-number-field`
- **Breaking**: Renamed `or-vaadin-passwordfield` to `or-vaadin-password-field`
- **Breaking**: Renamed `or-vaadin-textfield` to `or-vaadin-text-field`
- **Breaking**: Renamed `or-vaadin-textarea` to `or-vaadin-text-area`

## Findings during Vaadin migration

- **Vaadin Docs**: There is a lack of documentation on how their Web component mixins and theming system work, how to overwrite styles etc. I've found that the newer `Aura` theme is more in line with what we're looking for (providing more variables, less hardcoded styles and using less runtime code to apply styles, I think ??), but I don't think they've created a Figma template yet ??
- **Vaadin fields issues**:
  - **Default width**: Fields seem to use a default width on the field container. This can be difficult to work with as setting the variable for this field doesn't always overwrite the width properly. I've set the default to be `--vaadin-field-default-width: 100%;`. I have not found a way to change the styling through CSS selectors at the time of writing.
- **Tests**:
  - **Time field** (likely others): Using `.fill` action doesn't close the field dropdown menu, requires an explicit `.press("Enter")` action.
- **Components**:
  - **Vaadin Dialog**: The contents of the dialog are hard to make responsive .. due to slots ?? The Vaadin docs also seem to have this issue with fields extending past the dialog contents. 
    image
    The Vaadin docs do mention some tips to [make things responsive](https://vaadin.com/docs/latest/designing-apps/responsiveness).
    Also the [Vaadin dialog](https://vaadin.com/docs/latest/components/dialog) doesn't use the [`<dialog>` html element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) which I would've expected.
  - **Vaadin Checkbox Group**: The `@value-changed` event does not get fired if we use our own elements that extend `vaadin-checkbox`. So I've added the `import "@vaadin/checkbox"` to `or-vaadin-checkbox-group` to have the `<vaadin-checkbox>` tag available by default. The `Vaadin Radio Group` is also meant to use the Vaadin "native" elements (`vaadin-radio-button`). The group fields will emit a warning in the console if they can't find their children. Also the Radio group field works with the `@change` event. I've mentioned how to use these in the `@openremote/or-vaadin-components` README.

## Checklist

### DIFF
--- a/ui/component/or-scheduler/test/or-scheduler.test.ts
+++ b/ui/component/or-scheduler/test/or-scheduler.test.ts
@@ -109,7 +109,7 @@ ct.describe("Default event type should", () => {
         await vaadinDialog.getDialog().getByRole("button", { name: "Save schedule" }).click();
 
         const actual = (await promise).value;
-        expect(actual).toStrictEqual({ ...expected, recurrence: undefined });
+        expect(actual).toStrictEqual(expected);
         await expect(component.getByRole("button", { name: "Default" })).toHaveCount(1);
     });
 });


ui/component/or-scheduler/src/index.ts
```diff
@@ -320,112 +357,152 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
+            this._normalizedSchedule,
+            this._rrule,
+        ];
         return html`
-            <or-mwc-input outlined .type="${InputType.BUTTON}" label="${this.timeLabel()}" @or-mwc-input-changed="${() => this.showDialog()}"></or-mwc-input>
+            <or-vaadin-button @click="${() => this._dialog!.open()}">${this.timeLabel()}</or-vaadin-button>
+            <or-vaadin-dialog id="scheduler" header-title="${i18next.t(this.header)}" @closed="${() => this._dialog!.close()}"
+                ${dialogHeaderRenderer(this.getDialogHeader, [])}
+                ${dialogRenderer(this.getDialogContent, dependencies)}
+                ${dialogFooterRenderer(this.getDialogFooter, dependencies)}
```
Ekhorn:
```suggestion
                ${dialogFooterRenderer(this.getDialogFooter, [])}
```

ui/component/or-scheduler/src/index.ts
```diff
@@ -466,21 +543,23 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
-                        <or-mwc-input style="width: 60px;" min="1" max="9" .value="${interval}" .type="${InputType.NUMBER}"
-                            @or-mwc-input-changed="${(e: OrInputChangedEvent) => this.setRRuleValue(e.detail.value, "interval")}"></or-mwc-input>
+                        <or-vaadin-number-field min="1" max="9" step-buttons-visible style="width: 106px" .value="${interval}"
+                            @change="${(e: any) => this.setRRuleValue(e.target.value, "interval")}">
+                        </or-vaadin-number-field>
                     `)}
-                    <or-mwc-input style="flex: 1;" .value="${frequency.toString()}" .type="${InputType.SELECT}" .options="${frequencies}"
-                        @or-mwc-input-changed="${(e: OrInputChangedEvent) => this.setRRuleValue(e.detail.value, "freq")}"></or-mwc-input>
+                    <or-vaadin-select style="flex: 1;" .value="${frequency.toString()}" .items="${frequencies.map(([value, label]) => ({ value, label }))}"
+                        @change="${(e: any) => this.setRRuleValue(e.target?.value, "freq")}">
```
Ekhorn: Fix all any types.

```suggestion
                        @change="${(e: ...) => this.setRRuleValue(e.target?.value, "freq")}">
```

ui/component/or-scheduler/src/index.ts
```diff
@@ -466,21 +543,23 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
                 </div>
-                <div>${this.getByRulePart("bymonth", InputType.CHECKBOX_LIST, Object.entries(MONTHS))}</div>
+                <div>${this.getByRulePart("bymonth", InputType.CHECKBOX_LIST, Object.entries(MONTHS) as [string, string][])}</div>
                 <div>
                     ${this.getByRulePart("byweekno", InputType.SELECT, byWeekNoOptions)}
                     ${this.getByRulePart("byyearday", InputType.SELECT, byYearDayOptions)}
                     ${this.getByRulePart("bymonthday", InputType.SELECT, byMonthDayOptions)}
                 </div>
-                <div>${this.getByRulePart("byweekday", InputType.CHECKBOX_LIST, Object.keys(Days))}</div>
+                <div>${this.getByRulePart("byweekday", InputType.CHECKBOX_LIST, Object.keys(Days).map(v => [v, v]))}</div>
```
Ekhorn: Fix the entries generation and labels to sentence case

```suggestion
                <div>${this.getByRulePart("byweekday", InputType.CHECKBOX_LIST, Object.keys(Days))}</div>
```

ui/component/or-scheduler/src/index.ts
```diff
@@ -494,24 +573,23 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
-                                  .label="${part}"
-                                  multiple
-                                  @or-mwc-input-changed="${(e: OrInputChangedEvent) => {
-                                      this.setRRuleValue(e.detail.value, part);
-                                  }}"></or-mwc-input>`;
+            : this._rrule?.origOptions[part]
+        ) ?? [];
+
+        return type === InputType.CHECKBOX_LIST ? html`
+          <or-vaadin-checkbox-group .value="${value}" @value-changed="${(e: CheckboxGroupValueChangedEvent) => { if (!Util.objectsEqual(e.detail.value, value, true)) { this.setRRuleValue(e.detail.value, part) } }}">
```
Ekhorn: Consider handling this inside our own custom checkbox group by default?

```suggestion
          <or-vaadin-checkbox-group .value="${value}" @value-changed="${(e: CheckboxGroupValueChangedEvent) => this.setRRuleValue(e.detail.value, part)}">
```

ui/component/or-scheduler/test/fixtures/index.ts
```diff
@@ -18,16 +18,21 @@
...
+    vaadinDialog: VaadinDialog;
+    vaadinInput: VaadinInput;
 }
 
 export const ct = base.extend<ComponentFixtures>({
     // Components
     mwcDialog: withPage(MwcDialog),
     mwcInput: withPage(MwcInput),
+    vaadinDialog: withPage(VaadinDialog),
+    vaadinInput: withPage(VaadinInput),
```
Ekhorn:
```suggestion
import { VaadinDialog, VaadinInput } from "../../../or-vaadin-components/test/fixtures";
import { ct as base, type SharedComponentTestFixtures, withPage } from "@openremote/test";
export { expect } from "@openremote/test";

interface ComponentFixtures extends SharedComponentTestFixtures {
    vaadinDialog: VaadinDialog;
    vaadinInput: VaadinInput;
}

export const ct = base.extend<ComponentFixtures>({
    // Components
    vaadinDialog: withPage(VaadinDialog),
    vaadinInput: withPage(VaadinInput),
```

ui/component/or-vaadin-components/src/or-vaadin-input.ts
Ekhorn, line 1: Either revert or implement these fields.

ui/component/theme/src/components/or-vaadin-button.css
Ekhorn, line 1: Unused remove?

ui/component/theme/src/components/or-vaadin-icon.css
Ekhorn, line 1: Unused remove?

.yarnrc.yml
```diff
@@ -3,5 +3,6 @@ nodeLinker: node-modules
 npmMinimalAgeGate: "1w"
 npmPreapprovedPackages:
   - "@openremote/*"
+  - "@rspack/*"
```
Ekhorn:
```suggestion
```

ui/component/or-vaadin-components/test/fixtures/index.ts
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Get the underlying native input element of the or-mwc-input element
+     * @param type The {@link InputType} to look for
+     * @param locator The locator to start from
+     * @returns The locator to the underlying <input> element inside the or-mwc-input
+     */
+    getInputByType(type: `${InputType}`, locator?: Locator) {
+        return (locator ?? this.page).locator(`or-vaadin-input[type=${type}] #component > input`);
+    }
+
```
Ekhorn: Unused?
```suggestion
```

ui/component/or-scheduler/src/index.ts
```diff
@@ -320,112 +357,152 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
+                    gap: 2px;
+                }
+
+                @media only screen and (max-width: 768px) {
+                    .period {
+                        flex-wrap: wrap;
+                    }
+                }
+            </style>
+            <div style="display: flex; flex-direction: column; gap: var(--lumo-space-s); max-width: 85vw">
```
Ekhorn: Figure out what the wrapping dialog elements do to fit within the screen.

```suggestion
            <div style="display: flex; flex-direction: column; gap: var(--lumo-space-s);">
```

ui/component/or-scheduler/src/index.ts
Ekhorn: There's not much of a reason to do this.

ui/app/manager/src/index.ts
```diff
@@ -50,6 +50,10 @@ import {pageAlarmsProvider} from "./pages/page-alarms";
 import { ManagerAppConfig } from "@openremote/model";
 import {pageGatewayTunnelProvider} from "./pages/page-gateway-tunnel";
 
+// Import theme package for importing type definitions
+import "@openremote/theme";
+import themeCss from "@openremote/theme/default.css";
```
MartinaeyNL: This reminded me of the comment you made on my own PR.
Maybe you can already apply this here?

> To get rid of this can't we add `typeRoots` to the `tsconfig.json` file in the `@openremote/theme` package. Just like you have done for the `or-gauge`?
> 
> ```
> typeRoots: [ "./@types" ]
> ```

ui/app/shared/locales/en/or.json
```diff
@@ -34,6 +34,18 @@
...
+  "march": "March",
+  "april": "April",
+  "may": "May",
+  "june": "June",
+  "july": "July",
+  "august": "August",
+  "september": "September",
+  "october": "October",
+  "november": "November",
+  "december": "December",
```
MartinaeyNL, line 41: I would move this either above the "days of the week" or below them,
as they're now in between "monday" and the short "MO" version.

ui/app/shared/locales/nl/or.json
```diff
@@ -34,6 +34,18 @@
...
+  "march": "maart",
+  "april": "april",
+  "may": "mei",
+  "june": "juni",
+  "july": "juli",
+  "august": "augustus",
+  "september": "september",
+  "october": "oktober",
+  "november": "november",
+  "december": "december",
```
MartinaeyNL, line 41: (same comment here as on the English translation file `/shared/locales/en/or.json`

ui/component/or-icon/src/or-icon-set.ts
```diff
@@ -4,7 +4,8 @@ const orIconSet = {
...
+        "marker": "<ellipse fill=\"#000\" opacity=\"0.2\" ry=\"1.5\" rx=\"4\" cy=\"22.5\" cx=\"12\" /><path style='pointer-events: var(--or-icon-pointer-events, none);' d=\"m11.999999,0a8.437501,8.344057 0 0 1 8.437501,8.344057c0,6.500779 -5.798864,6.280799 -8.437501,14.412461c-2.638637,-8.131662 -8.437501,-7.911683 -8.437501,-14.412461a8.437501,8.344057 0 0 1 8.437501,-8.344057z\" />",
+        "trash": "<path d=\"M6 9V19C6 19.5523 6.44772 20 7 20H17C17.5523 20 18 19.5523 18 19V9\" stroke=\"#E21D12\" stroke-width=\"2\"/><path d=\"M14 11L14 16\" stroke=\"#E21D12\" stroke-width=\"2\" stroke-linecap=\"round\"/><path d=\"M10 11V16\" stroke=\"#E21D12\" stroke-width=\"2\" stroke-linecap=\"round\"/><path d=\"M5 6H8.5H19\" stroke=\"#E21D12\" stroke-width=\"2\" stroke-linecap=\"round\"/><path d=\"M8.09516 4.02871L7 6H17L15.9048 4.02871C15.5521 3.39378 14.8829 3 14.1565 3H9.84347C9.11714 3 8.4479 3.39378 8.09516 4.02871Z\" fill=\"#E21D12\"/>"
```
MartinaeyNL, line 8: Just for my information; is this custom `trash` SVG what we ended up using in the end?
As in, I assume this is a copy-paste SVG from what Remy made?

ui/component/or-scheduler/src/index.ts
```diff
@@ -94,12 +124,18 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
+
     @property({ type: Object })
     public schedule?: CalendarEvent = this.defaultSchedule;
 
     @property()
     public timezoneOffset = 0;
 
+    @state()
+    protected _count = 1;
+
```
MartinaeyNL, line 161: Based on the variable name I can't tell what this would do 😅
Maybe add a small 1 line comment? (or rename the variable, whatever makes sense)

ui/component/or-scheduler/src/index.ts
```diff
@@ -320,112 +355,153 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
-                    .mdc-dialog .mdc-dialog__content {
-                        padding: 8px 16px !important;
-                    }
-
-                    @media only screen and (max-width: 1279px) {
-                        .mdc-dialog__surface {
-                            overflow-x: auto !important;
-                            overflow-y: auto !important;
-                        }
+    protected getDialogHeader(): TemplateResult {
```
MartinaeyNL: Maybe nitpicking here, but I think it'd be nice to prefix protected and private functions using underscores, like `_myFunction()`.

ui/component/or-scheduler/src/index.ts
```diff
@@ -320,112 +355,153 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
+                }
 
-        return html`
-            <div style="min-width: 635px; display:grid; flex-direction: row;">
+                @media only screen and (max-width: 768px) {
+                    .period {
+                        flex-wrap: wrap;
+                    }
+                }
+            </style>
```
MartinaeyNL, line 478: I think this can be cleaned up right? With our discovery of styles inside `@openremote/theme` friday?

I would also consider to make this "custom checkbox group with texts" a separate component
inside the `/ui/component/or-vaadin-components` directory. As I assume we're going to use this more often?
And if not, it's nice to separate it out of the scheduler codebase here.

ui/component/or-scheduler/src/index.ts
MartinaeyNL, line 1: Let's skip the mobile view of the scheduler for now 😂 
We can iterate on this later, not a priority.

Image

ui/component/or-scheduler/src/index.ts
```diff
@@ -320,112 +355,153 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
+                }
+                if (this._eventType === EventTypes.default) {
+                    delete this._normalizedSchedule;
+                } else if (this._eventType === EventTypes.recurrence) {
+                    this._normalizedSchedule!.recurrence = this.getRRule();
+                }
+                const schedule = this.applyTimezoneOffset(this._normalizedSchedule ?? this.defaultSchedule, true);
+                this.dispatchEvent(new OrSchedulerChangedEvent(schedule));
+                this._dialog!.close();
+            }}"><or-translate value="schedule.save"></or-translate></or-vaadin-button>
```
MartinaeyNL: I would make these `@click` callback events separate functions.

ui/component/or-scheduler/src/index.ts
```diff
@@ -320,112 +355,153 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
-                    @media only screen and (max-width: 1279px) {
-                        .mdc-dialog__surface {
-                            overflow-x: auto !important;
-                            overflow-y: auto !important;
-                        }
+    protected getDialogHeader(): TemplateResult {
+        return html`
+            <vaadin-button theme="tertiary" @click="${() => this._dialog!.close()}">
+                <vaadin-icon icon="lumo:cross"></vaadin-icon>
+            </vaadin-button>
```
MartinaeyNL: Although it's only a `@click` with a dialog close, I would still move it to a dedicated function.

ui/component/or-scheduler/src/index.ts
```diff
@@ -320,112 +355,153 @@ export class OrScheduler extends translate(i18next)(LitElement) {
     }
 
     protected render() {
+        const timeLabel = this.timeLabel();
         return html`
-            <or-mwc-input outlined .type="${InputType.BUTTON}" label="${this.timeLabel()}" @or-mwc-input-changed="${() => this.showDialog()}"></or-mwc-input>
+            <or-vaadin-button @click="${() => this._dialog!.open()}">${timeLabel?.charAt(0).toUpperCase()}${timeLabel?.slice(1)}</or-vaadin-button>
```
MartinaeyNL: Although it's only a `@click` with a dialog open, I would still move it to a dedicated function.

ui/component/or-scheduler/src/index.ts
```diff
@@ -460,27 +536,30 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
-            .map(([k,v]) => [String(FrequencyValue[k as Frequency]), i18next.t(v, { count: interval })])
-            .filter(([k]) => this.isAllowedFrequency(FrequencyValue[k as Frequency] as any))
+            .map(([k, v]) => ({ value: String(FrequencyValue[k as Frequency]), label: i18next.t(v, { count: interval }) }))
+            .filter(({ value }) => this.isAllowedFrequency(FrequencyValue[value as Frequency] as any))
 
         return html`
             <div id="recurrence" class="section">
                 <label class="title"><or-translate value="schedule.repeat"></or-translate></label>
-                <div class="layout horizontal" style="display: flex; gap: 8px;">
+                <div style="display: flex; gap: 8px; margin-bottom: var(--lumo-space-l)">
```
MartinaeyNL: As I'm reviewing this code on a higher level, looking towards UI development for OR in general,
I was thinking about this `var(--lumo-space-1)` variable inside our codebase.
Dropping my thoughts here:

I think it makes the most sense if we have our own `--or-***` CSS variables for these kind of applications.
However, if you'd have to make a custom CSS variable for every variable that we use, it would create overhead.
I don't know whether using a component specific `--lumo-dialog-spacing-xs` would help, but I don't know what the best approach would be.

I tend to think that all lumo (or Vaadin) specific variables would have to be applied within the Vaadin components themselves (like `or-vaadin-button`), and all OR components would use our own variables.
But I have no clue if that is realistic at all. Let me know what you think!

ui/component/or-scheduler/src/index.ts
```diff
@@ -491,59 +570,63 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
-                                  }}"></or-mwc-input>`;
+            : this._rrule?.origOptions[part]
+        ) ?? [];
+
+        return type === InputType.CHECKBOX_LIST ? html`
+          <or-vaadin-checkbox-group .value="${value}" @value-changed="${(e: CheckboxGroupValueChangedEvent) => { if (!Util.objectsEqual(e.detail.value, value, true)) { this.setRRuleValue(e.detail.value, part) } }}">
+                ${(options as [string, string][]).map(([value, label]) => html`<vaadin-checkbox value="${value}" label="${i18next.t(label).slice(0, 3)}"></vaadin-checkbox>`)}
+            </or-vaadin-checkbox-group>
+        ` : html`
+            <or-vaadin-multi-select-combo-box .label="${i18next.t(part)}" .items="${options}" @change="${(e: any) => this.setRRuleValue(e.target.value, part)}"></or-vaadin-multi-select-combo-box>
```
MartinaeyNL: I'd write a separate callback function for this `@change` event listener too.

ui/component/or-scheduler/src/index.ts
```diff
@@ -491,59 +570,63 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
+                        <or-vaadin-time-picker style="margin-top: auto" ?hidden=${this.isAllDay} .value="${moment(calendar.end).format("HH:mm")}"
+                            @change="${(e: any) => this.setRRuleValue(e.target.value, "end-time")}">
+                        </or-vaadin-time-picker>
                     </div>
                 </div>
-
-                <div class="layout horizontal">
-                    <or-mwc-input .value=${this.isAllDay} @or-mwc-input-changed="${(e: OrInputChangedEvent) => this.setRRuleValue(e.detail.value, "all-day")}"  .type="${InputType.CHECKBOX}" .label="${i18next.t("allDay")}"></or-mwc-input>
-                </div>
+                <or-vaadin-checkbox .checked=${this.isAllDay} @change="${(e: any) => this.setRRuleValue(e.target.checked, "all-day")}" .label="${i18next.t("allDay")}"></or-vaadin-checkbox>
```
MartinaeyNL: Can't we make use of `<or-translate>` inside the Vaadin checkbox?
In this case it wouldn't matter, because the scheduler is inside a dialog,
but if you would change languages, the text is stuck in the old language.

ui/component/or-scheduler/src/index.ts
```diff
@@ -553,25 +636,38 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
+                            </or-vaadin-number-field><or-translate style="margin-left: 10px"
+                                ?disabled="${this._ends !== "count"}" value="schedule.count" .options="${{ count: +this._count }}">
+                                <style>
+                                [disabled] {
+                                    color: var(--lumo-contrast-20pct);
+                                    user-select: revert;
+                                    -webkit-user-select: none;
+                                }
+                                </style>
+                            </or-translate>
```
MartinaeyNL: Can this be cleaned up? Or use `styleMap()` instead? This also feels like a hack 😂 
But I probably don't have the context here haha

ui/component/or-scheduler/stories/or-scheduler.stories.ts
```diff
@@ -0,0 +1,110 @@
...
+        ] as Frequency[]
+        // Unused rrule parts
+        const DISABLED_RRULE_PARTS = [
+            'bymonth',
+            'byweekno',
+            'byyearday',
+            'bymonthday',
+            'byhour',
+            'byminute',
+            'bysecond' // Partially broken
```
MartinaeyNL: I remember you mentioning this in the office. Shouldn't we remove this option as a whole then?

ui/component/or-scheduler/test/or-scheduler.test.ts
MartinaeyNL, line 1: Funny that the test file didn't need as much changes!
It also looks much simpler in some cases, which is a good sign haha

Great stuff

ui/component/or-vaadin-components/README.md
```diff
@@ -14,15 +14,32 @@ yarn add @openremote/or-vaadin-components
...
+<or-vaadin-checkbox-group>
+    <vaadin-checkbox></vaadin-checkbox>
+    <vaadin-checkbox></vaadin-checkbox>
+</or-vaadin-checkbox-group>
+
+<or-vaadin-checkbox-group>
+    <vaadin-radio-button></vaadin-radio-button>
+    <vaadin-radio-button></vaadin-radio-button>
+</or-vaadin-checkbox-group>
+```
```
MartinaeyNL, line 39: We should think about this as well...
Not the most important for now, but not ideal if we'd like to keep "overriding" logic.
Almost makes me want to fork the repository 😂

ui/component/or-icon/src/or-icon-set.ts
Ekhorn: yes it is from Figma

ui/component/or-scheduler/src/index.ts
Ekhorn: Well I think based on the RRule context https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10 it makes sense. I think that goes for anything in this component.

ui/component/or-scheduler/src/index.ts
Ekhorn: This is all related to the scheduler ATM.

> I would also consider to make this "custom checkbox group with texts" a separate component inside the /ui/component/or-vaadin-components directory.

Sure, will do!

ui/component/or-scheduler/src/index.ts
Ekhorn: Ye I tried... but we should iterate on mobile designs at a later point indeed.

ui/component/or-scheduler/src/index.ts
Ekhorn: Ye need to have a look again, but from initial testing it didn't work..

ui/component/or-scheduler/src/index.ts
Ekhorn: Oh ye, wanted to move that to the scheduler styles as CSS class for this piece of text. Ideally this would be the label of the field. I'll try the label approach with some custom styles first (might be specific theme we can add for fields to place the label on any side).

ui/component/or-scheduler/stories/or-scheduler.stories.ts
Ekhorn: Not sure, would rather not make a breaking change for something like this. Would rather just fix it at some point or leave it be. It works, but visually the time label shows some weird text.

ui/app/manager/src/index.ts
Ekhorn: Actually briefly tried this, but had no luck. Might need to take a deeper look.

ui/component/or-vaadin-components/src/or-vaadin-dialog.ts
```diff
@@ -17,19 +17,49 @@
...
+    DialogFooterRendererDirective,
+} from "@vaadin/dialog/lit.js";
+
+type LitJoin<T> = T & typeof LitElement;
 
 @customElement("or-vaadin-dialog")
 export class OrVaadinDialog extends Dialog implements OrVaadinComponent {
+    static get styles() {
+        return [
+            (Dialog as LitJoin<typeof Dialog>).styles,
```
Ekhorn:
```suggestion
type WithLit<T> = T & typeof LitElement;

@customElement("or-vaadin-dialog")
export class OrVaadinDialog extends Dialog implements OrVaadinComponent {
    static get styles() {
        return [
            (Dialog as WithLit<typeof Dialog>).styles,
```

ui/component/or-scheduler/src/index.ts
Ekhorn: The thing is that Remy also needs to maintain these, and I think it would only be overhead for him, unless he thinks otherwise. Best to discuss when we're all back in the office.

ui/component/or-scheduler/test/or-scheduler.test.ts
Ekhorn: Only the manual "Enter" actions are annoying

ui/component/theme/src/mixins/field-internal.css
```diff
@@ -0,0 +1,9 @@
+@media lumo_mixins_field-internal {
```
Ekhorn, line 23:
```suggestion
/* We intentionally remove paddings around fields so we only need to deal with
 * our own paddings, margins and gaps.
 */
@media lumo_mixins_field-internal {
```

ui/component/or-vaadin-components/README.md
Ekhorn: Yes..

Also maybe we should have a section on how to handle events for normal and group fields. Storybook would maybe cover this, but I think it's still nice to have.

ui/app/manager/src/index.ts
MartinaeyNL: Would not spend too much time on it honestly. Not worth spending hours to "prevent this single line being here".

ui/component/or-icon/src/or-icon-set.ts
MartinaeyNL: 👍

ui/component/or-vaadin-components/README.md
MartinaeyNL: I guess we should reference the Vaadin docs often as well, preventing to do double work.
Because I assume they have docs on handling events.
Something for later I guess.

I'll create an issue to investigate replacing child components with our own.
So, instead of using `<vaadin-radio-button>` using `<or-vaadin-radio-button>` as children for the group.

MartinaeyNL: https://github.com/openremote/openremote/issues/2536

ui/component/or-scheduler/src/index.ts
Ekhorn: Couldn't really find a nice way to catch this in the component itself.

ui/component/or-scheduler/src/index.ts
Ekhorn: Looking at the Vaadin number field HTML template I don't think it's even worth trying to make the label float to a specific side of the input https://github.com/vaadin/web-components/blob/e351211ebc7d82d7cadfbcb5743256279a32d8f6/packages/number-field/src/vaadin-number-field.js#L106.

I've instead just moved this to the style tag of the dialog component.

ui/component/or-scheduler/src/index.ts
Ekhorn: Made this a "button" theme for vaadin-checkbox/or-vaadin-checkbox-group.

ui/app/manager/src/index.ts
Ekhorn: Resolved in https://github.com/openremote/openremote/pull/2512/commits/43c5a4506e7c62a18bda07ef3e3ac121054bd9c9, please check if this is okay.

ui/component/or-scheduler/src/index.ts
```diff
@@ -313,133 +341,184 @@ export class OrScheduler extends translate(i18next)(LitElement) {
...
+    protected _onClose() {
+        this._dialog!.close();
+    }
+
+    protected _onDelete() {
+        this._dialog!.close();
+        this.dispatchEvent(new OrSchedulerRemovedEvent());
+    }
+
+    protected _onSave() {
```
Ekhorn, line 511: Changing to event type default seems to apply the timezone offset incorrectly.

ui/component/or-scheduler/src/index.ts
MartinaeyNL: Discussed with the 3 of us, and decided to use `or-` prefixed tokens that use lumo variables by default.
So, for example like this:
```
--lumo-space-1: 10px;
--or-space-1: var(--lumo-space-1);
```

PR #2514, opened 2026-02-10
Export SQL injection fix

ebariaux:
This currently fixes the SQL injection possibility but makes the code a bit more complex.
I'd like to take a look at PostgreSQL JDBC driver CopyManager class to further improve this code, this will be addressed in a separate issue.

### DIFF
--- a/manager/src/main/java/org/openremote/manager/syslog/SyslogService.java
+++ b/manager/src/main/java/org/openremote/manager/syslog/SyslogService.java
@@ -134,7 +134,7 @@ public class SyslogService extends Handler implements ContainerService {
                 "delete from SyslogEvent e " +
                     "where e.timestamp <= :date")
                     .setParameter("date",
-                            Date.from(Instant.now().minus(config.getStoredMaxAgeMinutes(), ChronoUnit.MINUTES)))
+                            Instant.now().minus(config.getStoredMaxAgeMinutes(), ChronoUnit.MINUTES))
                     .executeUpdate());
         } catch (Throwable e) {
             LOG.log(Level.WARNING, "Failed to purge syslog events", e);


test/src/test/groovy/org/openremote/test/assets/AssetDatapointExportTest.groovy
```diff
@@ -105,4 +110,104 @@ class AssetDatapointExportTest extends Specification implements ManagerContainer
         cleanup: "Remove the limit on datapoint querying"
         assetDatapointService.datapointExportLimit = assetDatapointService.OR_DATA_POINTS_EXPORT_LIMIT_DEFAULT;
     }
+
+    def "Export query is vulnerable to SQL injection via attributeRefs"() {
```
Ekhorn:
```suggestion
    def "Export query is not vulnerable to SQL injection via attributeRefs"() {
```

test/src/test/groovy/org/openremote/test/assets/AssetDatapointExportTest.groovy
```diff
@@ -105,4 +110,104 @@ class AssetDatapointExportTest extends Specification implements ManagerContainer
...
+        )
+
+        then: "the CSV export should not contain injected rows"
+        def csvExport = csvExportFuture.get()
+        assert csvExport != null
+        def csvExportLines = csvExport.readLines()
+        assert csvExportLines.size() == 1
+    }
+
+    def "Export query is vulnerable to SQL injection via REST API"() {
```
Ekhorn:
```suggestion
    def "Export query is not vulnerable to SQL injection via REST API"() {
```

Ekhorn (commented):
Nice work!

Tested manually with 3 different attributes on 2 different assets and ~1 million datapoints. Performance seems unaffected. This will conflict with https://github.com/openremote/openremote/pull/2020/.

Ekhorn:
I'm seeing some potential performance degradation at 10mil datapoints. I'll investigate again later today.

Ekhorn:
Did some more testing with 10mil datapoints. Single exports take (on my system) instead of ~15s now ~19s. With 3 active streams from ~16-17s to ~19-24s (though this varies quite a lot). The time before the first chunk is received by the client increased quite significantly, likely due to the temp table and added joins. There are possible optimizations here, but let's explore those separately we would need to increase the request timeout on the frontend anyways when going past a few million datapoints.

PR #2522, opened 2026-02-11
Add predicted datapoints event

ebariaux:
A new AssetPredictedDatapointEvent is now published in case predicted data points for an attribute are inserted, updated or deleted.


model/src/main/java/org/openremote/model/datapoint/AssetPredictedDatapointEvent.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Objects;
+
+/**
+ * This event is used when predicted data points are inserted, updated or deleted for an asset's {@link org.openremote.model.attribute.Attribute}.
+ * It does not include any of the data point values.
+ */
+public class AssetPredictedDatapointEvent extends SharedEvent {
+
+    public enum Cause {
+        UPSERT,
```
pankalog: I'm not sure if there is such a differentiation, but shouldn't the enum be `INSERT, UPDATE, DELETE`? Users could possibly have different behaviors when a predicted datapoint is created compared to it being created (potential use-case below):

Battery A receives an `INSERT` event, and inserts the predicted datapoint (could be setpoint) to its schedule (an Array of some sort for example). When it receives an `UPDATE` event, it knows that it currently exists, and it can run back to that timestamp (so access its local array using the same index), and it knows that it currently exists. 

With `UPSERT`, it has to first check if that index exists (probably O(logn) complexity since timestamps can be sorted), then create or update depending on if the index exists.

If this is not possible implementation wise (i.e. the predicted datapoint service does not provide this differentiation between update and delete), then feel free to disregard this comment.

ebariaux: @pankalog it is possible (in fact I had this initially) but makes the implementation more complex (and the runtime cost higher) because it requires an implicit check before emitting the event, it's not 2 cleanly separated code paths.
As I did not need it for my use case and thought it might not bring much anyway, I went for a simpler version.

ebariaux: Discussing this in the SU, we opted to remove the cause altogether, not opening the door for some confusion on why no UPDATE vs INSERT.

test/src/test/groovy/org/openremote/test/assets/AssetDatapointTest.groovy
```diff
@@ -381,7 +385,14 @@ class AssetDatapointTest extends Specification implements ManagerContainerTrait
             assert datapoints.isEmpty()
         }
 
-        when: "predicted data points are added"
+        when: "we subscribe to predicted data points events"
+        List<AssetPredictedDatapointEvent> predictedEvents = new CopyOnWriteArrayList<>()
+        Consumer<AssetPredictedDatapointEvent> predictedEventConsumer = { event ->
+            predictedEvents.add(event)
+        }
+        clientEventService.addSubscription(AssetPredictedDatapointEvent.class, null, predictedEventConsumer)
```
richturner, line 462: This is a privileged method call that doesn't hit the event authorisers, test should go through `authorizeEventSubscription` which is called when a subscription is made over websocket from a client

richturner (changes requested):

manager/src/main/java/org/openremote/manager/datapoint/AssetPredictedDatapointService.java
richturner, line 1: Will need an `EventSubscriptionAuthorizer` (see review comment)

PR #2526, opened 2026-02-12
Upgrade Redux from 1.8.1 to the latest 2.11.2

MartinaeyNL:
## Description
While I was looking through old JS dependencies in our codebase, I came across our old version of `@reduxjs/toolkit`.
We use version 1.8.1, which was released in March 2022. So I did the work and upgraded it to the latest version 2.11.2.

List of breaking changes:
- `@reduxjs/toolkit`: https://redux.js.org/usage/migrations/migrating-rtk-2#toolkit-only
- `reselect`: https://redux.js.org/usage/migrations/migrating-rtk-2#reselect

Funny enough, none of the breaking changes applied to us. 🥳 
This is also because we don't use any of the advanced functions that Redux provides.

The upgrade also helps resolving some vulnerabilities that dependabot reported internally. 👍 

## Changelog
- Upgrades `@reduxjs/toolkit` from version `1.8.1`, to the latest version `2.11.2`.
- Upgrades several Redux related dependencies like `reselect` to their latest versions.

## Checklist
- [ ] 1. Tests are written and all tests pass
- [ ] 2. Changes are manually tested by you and the reviewer

### DIFF
--- a/test/load2-eks/README.md
+++ b/test/load2-eks/README.md
@@ -6,8 +6,11 @@ and adapting the values files to configure the memory usage on the different pod
 For the manager, JVM parameters are used to make use of the extra memory available to the container.  
 There are different "profiles" available depending on the power required and the tests to be run:
 - large: minimal set-up useful to test memory leaks and pressure condition
-- xlarge: doubles memory allocation for manager compared to large profile (4Gi vs 2Gi), with room for further increases
-- 2xlarge: much bigger setup to test system limits and get a feel for performance level we can reach
+- xlarge-minimal: doubles memory allocation for manager compared to large profile (4Gi vs 2Gi), with room for further increases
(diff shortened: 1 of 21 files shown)


MartinaeyNL:
As the CI/CD tends to fail consistently, I have been investigating what's going on.
At first sight, I thought the amount of breaking changes in Redux were minor, but it turns out it's the opposite...

It takes a significant effort to replace all `createSelector()` statements with the new syntax.
What also impacts it, is the new behavior regarding equality checks using memoize.
These are mostly described here: https://redux.js.org/usage/migrations/migrating-rtk-2#reselect

This means it's quite a lot of work, with major impact across the Manager UI, which can cause trouble.
I'm closing this PR for now. Potentially we can pick this up later.

PR #2530, opened 2026-02-12
Add ApplyPredictedDatapointService to automatically apply predicted data points as the current value

ebariaux:


manager/src/main/java/org/openremote/manager/datapoint/ApplyPredictedDataPointsService.java
```diff
@@ -0,0 +1,279 @@
...
+
+    protected TimerService timerService;
+    protected AssetStorageService assetStorageService;
+    protected AssetPredictedDatapointService assetPredictedDatapointService;
+    protected AssetProcessingService assetProcessingService;
+    protected ClientEventService clientEventService;
+    protected ScheduledExecutorService scheduledExecutorService;
+
+    protected final Object scheduleLock = new Object();
+    protected final PriorityQueue<ScheduledEntry> scheduleQueue = new PriorityQueue<>(Comparator.comparingLong(e -> e.timestamp));
```
richturner, line 69: Would `PriorityBlockingQueue` be an option to avoid the synchronisation

manager/src/main/java/org/openremote/manager/datapoint/ApplyPredictedDataPointsService.java
```diff
@@ -0,0 +1,279 @@
...
+            new AttributePredicate().meta(new NameValuePredicate(APPLY_PREDICTED_DATA_POINTS, new BooleanPredicate(true)))
+        ));
+        group.operator = LogicGroup.Operator.AND;
+
+        List<Asset<?>> assets = assetStorageService.findAll(new AssetQuery().attributes(group));
+        assets.forEach(asset ->
+            asset.getAttributes()
+                .stream()
+                .filter(attr -> hasRequiredMeta(attr.getMeta())) // Required, need to ensure at attribute level that both MetaItem are present
+                .forEach(attr -> rescheduleAttribute(new AttributeRef(asset.getId(), attr.getName())))
```
richturner: Maybe consider an overload for `rescheduleAttribute(Attribute attribute)` as going from `Attribute -> AttributeRef -> Attribute` and the `hasRequiredMeta` is being performed twice

manager/src/main/java/org/openremote/manager/datapoint/ApplyPredictedDataPointsService.java
```diff
@@ -0,0 +1,279 @@
...
+                }
+            } else {
+                if (nextFuture == null || ts < nextFuture.getTimestamp()) {
+                    nextFuture = datapoint;
+                }
+            }
+        }
+
+        long currentTimestamp = attribute.getTimestamp().orElse(0L);
+        // We apply the most recent past prediction unless the attribute was update more recently
```
richturner:
```suggestion
        // We apply the most recent past prediction unless the attribute was updated more recently
```

manager/src/main/java/org/openremote/manager/datapoint/ApplyPredictedDataPointsService.java
ebariaux: You still want to atomically modify both the scheduleQueue and the scheduledEntries map at the same time.
I don't think there's a way to achieve that without the synchronization. Using `PriorityBlockingQueue` and `ConcurrentHashMap` would map the operations thread-safe without sync but would potentially allow the information they contain to get out of sync.

manager/src/main/java/org/openremote/manager/datapoint/ApplyPredictedDataPointsService.java
ebariaux: Makes sense, looking into it in more details, as there might even be a bug, not sure attr.getMeta() has been fetched given the initial query on assets.

ebariaux: Thanks, indeed that was totally possible, pushed a fix for that.

PR #2533, opened 2026-02-13
Optionally add locations to generated assets for load2

ebariaux:


model/src/main/java/org/openremote/model/geo/GeoJSONPoint.java
```diff
@@ -89,6 +90,54 @@ public boolean hasZ() {
...
+    /**
+     * Returns a new {@link GeoJSONPoint} offset from this point by the provided east/north distances in meters.
+     * Uses GeoTools {@link GeodeticCalculator} for geodetic accuracy.
+     *
+     * @param eastMeters  distance to move east in meters
+     * @param northMeters distance to move north in meters
+     * @return a new point offset from this point
+     */
+    public GeoJSONPoint offsetByMeters(double eastMeters, double northMeters) {
+        double distance = Math.hypot(eastMeters, northMeters);
```
pankalog, line 126: A more efficient way to do this is `if (eastMeters == 0d && northMeters = 0d)`. This removes the need for calculating the hypotenuse.

model/src/test/java/org/openremote/model/geo/GeoJSONPointTest.java
```diff
@@ -0,0 +1,37 @@
+package org.openremote.model.geo;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+public class GeoJSONPointTest {
```
pankalog, line 9: Why not groovy test?

model/src/main/java/org/openremote/model/geo/GeoJSONPoint.java
ebariaux: Yes but if the distance is not 0, you'll need to compute it anyway later in the algorithm.
You would optimize for that case. I don't mind changing the code but I feel it's slightly more readable this way.

model/src/test/java/org/openremote/model/geo/GeoJSONPointTest.java
ebariaux: It's a simple unit test, I believe JUnit is more appropriate than Spock for this. We have too many slow integration tests (in a separate module, meaning it's "far away" from the code it's testing) and not enough unit test covering IMHO.

PR #2559, opened 2026-02-26
Make Value Descriptor schema ETag weak

Ekhorn:
## Description

The proxy changes the Value Descriptor ETag to a weak ETag, because the schema gets compressed and isn't identical byte-for-byte anymore. This causes the cache to miss (for deployments with a proxy that modifies the body), so to resolve this we specify the ETag to be weak by default.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag

## Checklist


manager/src/main/java/org/openremote/manager/asset/AssetModelResourceImpl.java
Ekhorn, line 1: Nope doesn't fix the issue for deployments with a proxy.

Ekhorn: Apparently it does work on chromium (and WebKit), but not on Firefox.

It is broken on staging for chromium, WebKit and firefox.

On test1 it is only broken for Firefox.

Ekhorn: The problem is simply Firefox doesn't include the `If-None-Match
` header on a deployment. It seems that the `Cache-Control` header might explicitly need to be enabled for Firefox to send the header.

Ekhorn: Nevermind now it is including the header... and revalidating

Image

PR #2562, opened 2026-02-27
Use kubernetes version 1.35 when creating EKS clusters

ebariaux:


Ekhorn:
Looks like all versions have been updated to 1.35

Just checking, we're not using any deprecated/removed features listed in https://kubernetes.io/blog/2025/12/17/kubernetes-v1-35-release/#deprecations-removals-and-community-updates

- I see in the code base uses of an ingress with the `className: "nginx"` not sure that is the exact thing they describe in the release notes. If it is, maybe good to create an issue to migrate to the Gateway API (ref https://kubernetes.io/blog/2025/12/17/kubernetes-v1-35-release/#ingress-nginx-retirement). See https://github.com/openremote/openremote/blob/efb8b3682da8d3a35eb70e6afb895243e48c7636/kubernetes/manager/values.yaml#L80

PR #2579, opened 2026-03-04
Agent: Add write value filter functionality to agent links

Hackerberg43:
## Description
As discussed in https://github.com/openremote/vdl-es/issues/222 , the addition of writeValueFilter agentlink meta to apply filters to outbound values, in addition to the current valueFilter which only applies to inbound values. An application is bidirectional communication with an external attribute which has a different scaling, which has to work in both directions but inverted.
## Checklist

### DIFF
--- a/model/src/main/java/org/openremote/model/protocol/ProtocolUtil.java
+++ b/model/src/main/java/org/openremote/model/protocol/ProtocolUtil.java
@@ -138,7 +138,7 @@ public final class ProtocolUtil {
         valRef.set(ignoreAndConvertedValue.value);
 
         if (valRef.get() == null) {
-            return new Pair<>(true, null);
+            return new Pair<>(false, null);
         }
 
         // built in value conversion


model/src/main/java/org/openremote/model/protocol/ProtocolUtil.java
```diff
@@ -66,6 +66,18 @@ public static Pair<Boolean, Object> doOutboundValueProcessing(AttributeRef attri
...
+        value = agentLink.getWriteValueFilters().<Object>map(filters -> {
+            LOG.finest("Applying attribute write value filters to attribute: " + attributeRef);
+            return applyValueFilters(valueBeforeFilters, filters);
+        }).orElse(valueBeforeFilters);
+
+        if (value == null) {
+            LOG.finest("Write value filters produced a null value for attribute: " + attributeRef);
+            return new Pair<>(true, null);
+        }
+
```
richturner: Please copy and paste the `doInboundValueProcessing()` block and adjust logs statements and switch `getValueFilters()` for `getWriteValueFilters(). Log statemements should also align with wording in inbound value filtering.

Hackerberg43: That requires also altering the existing outbound value conversion to match with the inbound pattern. Right now ,the writevaluefilter is alligned with the existing outbound value conversion. Please let me know what you prefer.

richturner: Let's just get this in as code styling can be improved later

richturner (approved):

richturner:
Test keeps consistently failing, not checked if related to changes in this PR or general test stability issue, I suspect related to this PR

Hackerberg43:
I wil take a look right now

Hackerberg43:
So the tests did their job, because the introduced null check caused other tests to fail. Removed null check, passes now.

PR #2591, opened 2026-03-05
Set general indenting rules in .editorconfig

Ekhorn:
Setting indentation explicitly ensures that editors do not rely on users' default `.editorconfig` settings.

This was causing weird auto-formatting issues in (one of) the editors I use.

### DIFF
--- a/setup/src/load2/java/org/openremote/setup/load2/KeycloakSetup.java
+++ b/setup/src/load2/java/org/openremote/setup/load2/KeycloakSetup.java
@@ -89,10 +89,10 @@ public class KeycloakSetup extends AbstractKeycloakSetup {
 
         // Wait until all users created
         int waitCounter = 0;
-        int waitCounterLimit = accounts / 3;
+        int waitCounterLimit = Math.max(300, accounts / 3);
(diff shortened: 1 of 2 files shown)


MartinaeyNL:
Isn't this in conflict with #1441 ? Or would this change be an "for the mean time" solution?

Ekhorn:
> Isn't this in conflict with https://github.com/openremote/openremote/pull/1441 ?

In theory yes, Spotless doesn't itself seem to implement direct `.editorconfig` support looking at the issue tracker (you define per language how to format things). Though many tools it supports also infer settings from `.editorconfig`.

> Or would this change be an "for the mean time" solution?

Yes, for the mean time as historically we have delayed spotless integration. I can add a comment to #1441.

MartinaeyNL:
As discussed, I think it makes sense to close this issue, and stick to using it locally for personal use.
It's only a matter of time until we apply Spotless to all active branches, therefore that makes more sense.

Ekhorn:
I'll close this and use the following instead:

```sh
git update-index --skip-worktree .editorconfig
git update-index --no-skip-worktree .editorconfig
```

PR #2592, opened 2026-03-05
Include demo extension for OR_SETUP_TYPE=demo gradle run

Ekhorn:
This ensures that people who like to start the manager from the CLI (me :wink:) can include the demo setup as well.

This will only build the demo extension when the `OR_SETUP_TYPE` environment variable is set to `demo`.


MartinaeyNL:
I think it makes sense to let @wborn take a quick look 👍

manager/build.gradle
```diff
@@ -180,3 +180,19 @@ publish.mustRunAfter(setVersionProperties)
...
+dependencies {
+    demoExtension project(":extension:demo-setup")
+}
+
+tasks.named("run", JavaExec) {
+    def combinedClasspath = sourceSets.main.runtimeClasspath
+    if (System.getenv("OR_SETUP_TYPE") == "demo") {
+        combinedClasspath += configurations.demoExtension
+    }
+    classpath = combinedClasspath
```
wborn: Does something this also work (less code)?

```suggestion
    if (System.getenv("OR_SETUP_TYPE") == "demo") {
        classpath += configurations.demoExtension
    }
```

or else

```suggestion
    if (System.getenv("OR_SETUP_TYPE") == "demo") {
        classpath = sourceSets.main.runtimeClasspath
        classpath += configurations.demoExtension
    }
```

Issue #2521, opened 2026-02-11
Mismatch in timezones in 'Retrieve the predicted datapoints of an asset attribute' API


igorrutka:
**Bug**
The input date-times with timezone UTC mach the local date-times in the result and not the UTC date-times.

When using the '_Retrieve the predicted datapoints of an asset attribute_' API (https://docs.openremote.io/docs/rest-api/get-predicted-datapoints) the time range in the payload looks like this for 1 day:

Payload = {“fromTimestamp": 0, "toTimestamp": 0, "fromTime": "2026-02-12T00:00:00.000Z", "toTime": "2026-02-13T00:00:00.000Z", "type": "string"}

The first timestamp in the result is:

milliseconds since epoch = 1770850800000
UTC date-time = Wed Feb 11 2026 23:00:00.000
Local date-time = Thu Feb 12 2026 00:00:00.000

The Z part of the input date-time creates confusion. The Z indicates you are specifying the date-time in UTC time but if you look at the result it is actually local date-time that matches the input date-time.

**Solution**
If using a timezone in the input, the output should match the timezone

ebariaux:
Also check if anything specified in documentation and update as required.
Check if other APIs have similar behaviour/issues and create GH issues as required.

ebariaux:
The input does not support timezones.
Tje JSON payload is mapped to a `AssetDatapointQuery` object with `fromTime` and `toTime` properties of type `LocalDateTime`, that does not include a TZ.
The JSON parsing does not support specifying timezones, it either supports nothing or trailing Z, specifying a timezone returns a 400 error.
If trailing Z is present, it's basically ignored, it does not mean UTC.

ebariaux:
# Complete flow explanation

## Publishing a predicted data point

The `timestamp` column in `asset_predicted_datapoint` table is of type 'timestamp without time zone', which means it stores a date and a time without information about the timezone and not even knowing which timezone was used for storing the value.

```bash
curl -k -i -X PUT "https://localhost:443/api/master/asset/predicted/4cEkQkbke0bFz6n5En1hZy/SomeNumber" \
    -H "Authorization: Bearer $ACCESS_TOKEN" \
    -H "Content-Type: application/json" \
    -d '[
      { "timestamp": 1775047200000, "value": 123.45 }
    ]'
```
Performs the conversion of the given millis value to a date/time information using the default TZ of JVM (AMS for us, set as default in our Docker image), in this case it means 2026-04-01 14:40:00.
[...]

PR #2595, opened 2026-03-06
Support timezone in the get datapoints query payloads

ebariaux:
See issue #2521 for description


MartinaeyNL (approved):
I've tested this locally, and it all seems to work great! 👍 
Thanks!

PR #2610, opened 2026-03-13
Fix map marker flickering when receiving WebSocket events

### LINKED ISSUE #2290: Fix marker flickering when WebSocket updates happen
**Describe the bug**

The asset markers flicker whenever a WebSocket update is sent. The map clustering handles WebSocket updates naively by reloading all markers when an update is received https://github.com/openremote/openremote/pull/2284.

**To Reproduce**
1. Open the manager
2. Open a realm with asset markers
3. See them flicker on initial load (and when WebSocket updates come in)

**Expected behavior**

- [x] 1. When an asset event is received it should only update the markers that are affected.
- [x] 2. When an attribute event is received it should only update the markers that are affected.

**Additional context**

We currently call reload to update all map markers https://github.com/openremote/openremote/blob/a316707f3f9b21ade87a6519e720e72fc2795dac/ui/app/manager/src/pages/page-map.ts#L218

Ekhorn:
## Description

- Closes #2290 

## Changelog

- Rename `mapwidget.ts` to `base-map.ts`
- Separate asset related code from the base map to the new AssetMap class
  - The AssetMap class uses MapLibre source methods to update the map state
  - ~~The rendered features are now explicitly requested as rendered features~~
- Reworks the `page-map` state management, which resolves the map flickering issue
  - This is also now updates removed assets on the map through the WebSocket (before this only worked through attribute updates)
- ~~Adds setup for configurable map clustering~~

## Checklist

### DIFF
--- a/balena/docker-compose.yml
+++ b/balena/docker-compose.yml
@@ -1,3 +1,6 @@
+# Balena CLI requires this version, see: https://github.com/balena-io/balena-cli/issues/2314
+version: '2.4'
+
 volumes:
   proxy-data:
   postgresql-data:
--- a/balena/run_balena_build.sh
+++ b/balena/run_balena_build.sh
@@ -47,9 +47,9 @@ if [ ! -d "./manager-build" ]; then
     cp -r ../manager/build/install/manager ./manager-build
 fi
 
-# Push to balena with no cache or run docker-compose up if dry-run
+# Push to balena with no cache or run docker compose up if dry-run
 if [ "$DRY_RUN" = true ]; then
-    docker compose build --no-cache && docker-compose -f ./docker-compose.yml -p openremote up --force-recreate -d
+    docker compose -f ./docker-compose.yml -p openremote build --no-cache && docker compose -f ./docker-compose.yml -p openremote up --force-recreate -d
 else
     balena push "$BALENA_FLEET_SLUG" --nocache
 fi


ui/component/or-map/src/index.ts
```diff
@@ -377,18 +353,24 @@ export class OrMap extends LitElement {
...
-        this._map?.cleanUpAssetMarkers();
+    public removeAssets(ids: string[]) {
+        this._map?.removeAssets(ids);
     }
 
-    public async reload() {
-        await this._map?.load();
+    public clearAssets(): Promise<void> | void {
+        this._map?.clearAssets();
+    }
```
Ekhorn:
```suggestion
    public clearAssets(): void {
        this._map?.clearAssets();
    }
```

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,286 @@
...
+
+        this._map.on("data", this._onData);
+
+        this._mapContainer.dispatchEvent(new OrMapLoadedEvent());
+        this._loaded = true;
+    }
+
+    protected _updateMarkers() {
+        if (!this._map) return;
+        const features = this._map.queryRenderedFeatures({ layers: ["asset", "cluster"] });
```
Ekhorn: Rendered feature querying makes `_updateMarkers` take 6.9% of the sampled time compared to only 2.0% when querying all source features.

```suggestion
        const features = this._map.querySourceFeatures("assets");
```

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,286 @@
...
+        const newAssets: Record<string, AssetWithLocation> = {};
+
+        // TODO: check for duplicate features
+        for (const feature of features) {
+            if (!feature.properties.id) continue;
+            const id = feature.properties.id as string;
+            const [lon, lat] = (feature.geometry as Point).coordinates;
+            const marker = this._markersOR.get(id);
+            if (marker instanceof OrMapMarker) {
+                this._markersGL.get(marker)?.setLngLat([lon, lat]);
```
Ekhorn: This is redundant when setting the state on the marker.

```suggestion
```

ui/component/or-map/src/base-map.ts
```diff
@@ -744,17 +563,20 @@ export class MapWidget {
...
 
     protected _updateMarkerElement(marker: OrMapMarker, doAdd: boolean) {
-        let mGl = this._markers.get(marker);
+        let mGl = this._markersGL.get(marker);
         if (mGl) {
             marker._actualMarkerElement = undefined;
             this._removeMarkerClickHandler(marker, mGl.getElement());
             mGl.remove();
-            this._markers.delete(marker);
+            this._markersGL.delete(marker);
```
Ekhorn: It seems that the following map of markers is ever expanding. They're not properly cleaned.

```
            this._markersGL.delete(marker);
```

ui/app/manager/src/pages/page-map.ts
```diff
@@ -383,12 +414,24 @@ export class PageMap extends Page<MapStateKeyed> {
...
     }
 
+    shouldUpdate(_changedProperties: PropertyValues): boolean {
+        if (_changedProperties.has('_assets')) {
+            const newTypes = [];
+            for (const { type } of this._assets) {
+                if (!newTypes.includes(type)) newTypes.push(type);
+            }
+            this._assetTypes = newTypes;
+
```
Ekhorn:
```suggestion
```

ui/app/manager/src/index.ts
```diff
@@ -62,7 +62,9 @@ const rootReducer = combineReducers({
 type RootState = ReturnType<typeof rootReducer>;
 
 export const store = configureStore({
-    reducer: rootReducer
+    reducer: rootReducer,
+    // Disable for development to improve performance (usually always disabled in production)
+    middleware: (getDefaultMiddleware) => getDefaultMiddleware({ serializableCheck: false }),
```
Ekhorn: I probably should have deeper look into this as I've not really looked much into the documentation about this.

Ekhorn: The docs explain this is indeed configured by default with just `configureStore`, or `getDefaultMiddleware` (see https://redux-toolkit.js.org/api/serializabilityMiddleware). You can also read here https://redux-toolkit.js.org/api/getDefaultMiddleware#included-default-middleware that indeed in development the serializable middleware is included by default, but not in production.

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,300 @@
...
+            // On WebKit browsers the clicked marker may not be removed,
+            // if the camera zoom level is directly between 2 levels
+            if (e.marker instanceof OrClusterMarker) {
+                this._clustersOnScreen[e.marker._clusterId]?.remove();
+            }
+            this._updateMarkers();
+        });
+    };
+
+    protected _onData = (e: MapSourceDataEvent) => {
```
Miggets7, line 49: The _onData callback fires every time the `assets` GeoJSON source finishes loading (on every updateData() call). Each time, it re-registers this._onMove and this._onMoveEnd. MapLibre's on() does NOT deduplicate -- so after N source data events, _updateMarkers() fires N times per move event.

Either remove _onData listener after first fire, or add listeners once in load() instead

ui/app/manager/src/pages/page-map.ts
```diff
@@ -305,7 +329,22 @@ export class PageMap extends Page<MapStateKeyed> {
...
                 const attributeSubscriptionId = await manager.events.subscribeAttributeEvents(undefined, false, (event) => {
+                    const interested = attrsOfInterest.includes(event.ref.name);
+                    let asset: Asset;
+                    if (interested) {
+                        const assets = this._unlocatedAssetSelector(this.getState());
+                        asset = assets.find(asset => asset.id === event.ref.id);
+                        // Update the attribute if the asset already has a location
+                        // assuming the asset is in the located asset state
+                        if (!asset) {
+                            this._map.updateAttribute(event);
```
Miggets7: All other calls in the same context use this._map?
```suggestion
                            this._map?.updateAttribute(event);
```

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,300 @@
...
+        this._source?.updateData({ update: [{ id, newGeometry, addOrUpdateProperties: newProperties }] });
+    }
+
+    public removeAssets(ids: string[]) {
+        this._source?.updateData({ remove: ids });
+    }
+
+    public removeAllAssets() {
+        this._source?.updateData({ removeAll: true });
+    }
```
Miggets7, line 156: Does this also clean the `_assets ` dictionary?

If not, you may have stale entries in `_assets`
Don't forget to also check `_assetsOnScreen`, to see if it needs to be updated.

ui/app/manager/src/pages/page-map.ts
```diff
@@ -305,7 +329,22 @@ export class PageMap extends Page<MapStateKeyed> {
...
+                        }
+                    }
                     this._store.dispatch(attributeEventReceived([attrsOfInterest, event]));
```
Miggets7, line 343: Is is async?

Because if so, the `_assets` might not be fully populated yet

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,300 @@
...
+                type: "geojson",
+                cluster: this._clusterConfig?.cluster ?? true,
+                clusterRadius: this._clusterConfig?.clusterRadius ?? 180,
+                clusterMaxZoom: this._clusterConfig?.clusterMaxZoom ?? 17,
+                clusterProperties: this._getClusterProperties(),
+                promoteId: "id", // Promote the id property as feature id
+                data: { type: "FeatureCollection", features: [] }, // TODO: consider preloading data to avoid tile cache misses
+            })
+            .addLayer({ ...invisibleLayer, id: "asset", filter: ["has", "point_count"] })
+            .addLayer({ ...invisibleLayer, id: "cluster", filter: ["!", ["has", "point_count"]] });
```
Miggets7: Are these names swapped?

ui/component/or-dashboard-builder/src/widgets/map-widget.ts
Miggets7, line 95: Missing disconnect for `OrMapLoadedEvent`

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,300 @@
...
+import { Feature, Point } from "geojson";
+import { AssetWithLocation, ClusterConfig } from "./types";
+import { BaseMap } from "./base-map";
+import { OrClusterMarker, Slice } from "./markers/or-cluster-marker";
+import { getMarkerIconAndColorFromAssetType } from "./util";
+import { OrMapLoadedEvent, OrMapMarkersChangedEvent } from ".";
+import { Util } from "@openremote/core";
+import { AttributeEvent } from "@openremote/model";
+
+type MissingAsset = AssetWithLocation & { id: string; type: string };
```
Miggets7: Name suggests an asset that is absent, when it actually represents an asset with all required fields. Consider renaming to `RenderableAsset` or `MapAsset`

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,300 @@
...
+            this._assetTypeColors[asset.type] = getMarkerIconAndColorFromAssetType(asset.type)?.color as string; // TODO: test with marker config
+            this._source.workerOptions.clusterProperties = this._getClusterProperties();
+        }
+
+        this._assets[asset.id] = asset;
+        this._source.updateData({ add: [AssetMap._assetToFeature(asset)] });
+    }
+
+    /**
+     * Add multiple assets as (asset) features to the map. Optimized for adding mutliple assets at once.
```
Miggets7:
```suggestion
     * Add multiple assets as (asset) features to the map. Optimized for adding multiple assets at once.
```

ui/component/or-map/src/asset-map.ts
Ekhorn: I moved them out (down into the load method) next to the `on(data` event registration, but from very brief profiling I got the idea that it made things worse. I'll revisit this though MapLibre also does this in their example https://maplibre.org/maplibre-gl-js/docs/examples/display-html-clusters-with-custom-properties/

Ekhorn: I don't recall it registering multiple event listeners, but I'll check this. Looking at their code it seems that it would register the listener once in their listeners object (per event type) https://github.com/maplibre/maplibre-gl-js/blob/109fa99ea55c16a26cf8ce294d96bf4b0b93524a/src/util/evented.ts#L8 so it ends up resetting the listener each time.

Miggets7: I could be wrong, it just looked it might happen, maybe should have said that. I would say don't spend too much time on this if you see no strange stuff happening when profiling.

ui/app/manager/src/pages/page-map.ts
Ekhorn: I'm assuming you meant is the dispatch function async? Based on https://redux.js.org/api/store/#dispatchaction it seems synchronous.

ui/component/or-map/src/asset-map.ts
Ekhorn: ~~I assume you mean is the id overridden? -> Yes~~

~~I could maybe remove the type and id property from de-structured object to be more clear.~~

ui/component/or-map/src/asset-map.ts
Ekhorn: Changed to `IdentifiableAsset`

ui/component/or-map/src/asset-map.ts
Ekhorn: Yea, so just briefly looked at the map instance itself (which is where we register this) it extends the Camera class which extends the evented class with the event registration I mentioned (see https://github.com/maplibre/maplibre-gl-js/blob/c2e92ecc7f99d9bc8baa52379dabc6019bb6ea84/src/ui/map.ts#L1745).

Not looking further into it.

ui/component/or-map/src/asset-map.ts
Ekhorn: Resolved, see https://github.com/openremote/openremote/pull/2610/commits/78a6340b56c2859e225aeb0e431782b0e46efed7.

ui/app/manager/src/pages/page-map.ts
Ekhorn: This is what the jsdoc says about the dispatch, so should be fine.

```ts
/**
   * The base implementation only supports plain object actions. If you want
   * to dispatch a Promise, an Observable, a thunk, or something else, you
   * need to wrap your store creating function into the corresponding
   * middleware. For example, see the documentation for the `redux-thunk`
   * package. Even the middleware will eventually dispatch plain object
   * actions using this method.
   */
```

ui/component/or-map/src/asset-map.ts
Ekhorn: So, removing the stale assets adds a slight overhead to the `removeAssets` function (and not sure how garbage collection will behave). Addressed this in https://github.com/openremote/openremote/pull/2610/commits/866977f03bb1c1792b2cb1f63f60af6ebc5b0c2a.

There is no need to clearing `_assetsOnScreen` as this is always cleared when updating the markers and because these operations change the state of the map it will update the markers and thus clear the `_assetsOnScreen` state.

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,301 @@
...
+
+/**
+ * The main component that handles how assets are displayed on the map.
+ *
+ * Currently, the standard flow to display assets as markers on the map goes as follows:
+ *
+ * 1. The map page adds assets --> the {@link AssetMap}, which computes assets in view --> Then dispatches {@link OrMapMarkersChangedEvent}
+ * 2. The map page receives the event --> Renders the received assets as {@link OrMapMarkerAsset} by including them inside the map element's HTML body
+ * 3. The map component then receives either an event from the {@link OrMapMarkerAsset}
+ * or when the map component sees a slotted marker was added --> Which adds markers as maplibre marker to the maplibre-gl map.
```
Ekhorn:
```suggestion
 * 1. The map page adds assets to the {@link AssetMap}, which computes assets in view and then dispatches {@link OrMapMarkersChangedEvent}.
 * 2. The map page receives the event, renders the received assets as {@link OrMapMarkerAsset} by including them inside the map element's HTML body.
 * 3. The map component then receives either an event from the {@link OrMapMarkerAsset},
 * or when the map component sees a slotted {@link OrMapMarker} was added. From there the {@link OrMapMarker}s are passed to the {@link BaseMap} to be assigned to a MapLibre marker.
```

ui/component/or-map/src/or-map-legend.ts
```diff
@@ -77,22 +78,23 @@ export class OrMapLegend extends LitElement {
...
                                 ></or-icon>
                                 <span id="asset-label" style="flex: 1">${this._assetTypesInfo[assetType].label}</span>
                                 <or-mwc-input
                                     .type="${InputType.CHECKBOX}"
-                                    .value="${!this._excludedTypes.includes(assetType)}"
+                                    .value="${!this.excludedTypes.includes(assetType)}"
                                     @or-mwc-input-changed="${(ev: OrInputChangedEvent) => {
                                         if (ev.detail.value) {
-                                            this._excludedTypes.splice(this._excludedTypes.indexOf(assetType), 1);
+                                            this.excludedTypes.splice(this.excludedTypes.indexOf(assetType), 1);
```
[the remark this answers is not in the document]
Ekhorn: If asset type is not present something was miss configured, not going to deal with that.

ui/component/or-map/src/markers/or-cluster-marker.ts
```diff
@@ -33,15 +33,13 @@ export class OrClusterMarker extends LitElement {
...
      */
-    public hasTypes(types: string[]) {
-        for (const type of types) {
-            if (!this._slices.some(([t]) => t === type)) {
-                return false;
-            }
+    public slicesMatch(counts: Record<string, number>) {
+        for (const [type, , count] of this._slices) {
+            if (counts[type] !== count) return false;
         }
```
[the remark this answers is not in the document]
Ekhorn: When a new type appears that means a new feature matching a new cluster property. The clustering in MapLibre will recompute the cluster and give it a new id, thus invalidating the old one automatically. No need to make it more complicated.

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,301 @@
...
+    protected _source?: GeoJSONSource;
+
+    protected _onMove = () => this._updateMarkers();
+    protected _onMoveEnd = (e: any) => {
+        // Ensure marker updates happen after the frame
+        requestAnimationFrame(() => {
+            // On WebKit browsers the clicked marker may not be removed,
+            // if the camera zoom level is directly between 2 levels
+            if (e.marker instanceof OrClusterMarker) {
+                this._clustersOnScreen[e.marker._clusterId]?.remove();
```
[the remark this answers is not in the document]
Ekhorn: There's no need for this, as we're resetting the clusters on screen completely every update.

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,301 @@
...
+        // Create missing cluster markers
+        for (const feature of features) {
+            if (!feature.properties.cluster) continue;
+            const id: number = feature.properties.cluster_id;
+            let marker = this._cachedClusters[id] ?? this._getClusterMarker(feature).addTo(this._map!);
+
+            // Invalidate the cached cluster marker if the slices don't match
+            if (marker._element instanceof OrClusterMarker && !marker._element.slicesMatch(feature.properties)) {
+                marker.remove();
+                marker = this._getClusterMarker(feature).addTo(this._map!);
```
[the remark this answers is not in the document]
Ekhorn: Removed one redundant `addTo` call.

ui/app/manager/src/pages/page-map.ts
```diff
@@ -305,7 +329,22 @@ export class PageMap extends Page<MapStateKeyed> {
...
+                        // Update the attribute if the asset already has a location
+                        // assuming the asset is in the located asset state
+                        if (!asset) {
+                            this._map?.updateAttribute(event);
+                        }
+                    }
                     this._store.dispatch(attributeEventReceived([attrsOfInterest, event]));
+                    // Add the asset after map state has been updated
+                    if (interested && asset) {
+                        this._map?.addAsset(this._assets.find(asset => asset.id === event.ref.id));
```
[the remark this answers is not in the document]
Ekhorn: The `hasRequired` check in the AssetMap already deals with this no need to guard check every thing double. The types could be improved, but at the same time all properties on the asset type are nullable anyways so optional chaining deals with this.

wborn:
@codex review

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        if (!(asset.type in this._assetTypeColors)) {
+            this._assetTypeColors[asset.type] = getMarkerIconAndColorFromAssetType(asset.type)?.color as string;
+            this._source.workerOptions.clusterProperties = this._getClusterProperties();
+        }
+
+        this._assets[asset.id] = asset;
+        this._source.updateData({ add: [AssetMap._assetToFeature(asset)] });
+    }
```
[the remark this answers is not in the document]
Ekhorn: No, not worth my time.

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,301 @@
...
+                this._assetTypeColors[asset.type] = getMarkerIconAndColorFromAssetType(asset.type)?.color as string;
+            }
+            this._assets[asset.id] = asset;
+        }
+        this._source.workerOptions.clusterProperties = this._getClusterProperties();
+
+        this._source.updateData({
+            add: missing.map(AssetMap._assetToFeature),
+        });
+    }
```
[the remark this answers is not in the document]
Ekhorn: No, not worth my time.

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,301 @@
...
+        const newProperties = Object.entries(AssetMap._assetToFeature(asset as IdentifiableAsset).properties).map(
+            ([key, value]) => ({ key, value })
+        );
+        this._source?.updateData({ update: [{ id, newGeometry, addOrUpdateProperties: newProperties }] });
+    }
+
+    public removeAssets(ids: string[]) {
+        for (const id of ids) this._assets[id] = null;
+        this._source?.updateData({ remove: ids });
+    }
```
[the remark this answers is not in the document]
Ekhorn: no the `delete` operation is slow. I could consider a map, but I won't for now.

ui/component/or-dashboard-builder/src/widgets/map-widget.ts
```diff
@@ -153,6 +155,12 @@ export class MapWidget extends OrAssetWidget {
         `;
     }
 
+    protected _onMapLoaded(e: OrMapLoadedEvent) {
+        const assetType = this.widgetConfig.allOfType ? undefined : this.widgetConfig.assetType;
+        const assets = this.loadedAssets.filter(asset => !assetType || asset.type === assetType).filter(MapUtil.isAssetWithLocation);
+        this._map?.addAssets(assets);
```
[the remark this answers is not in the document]
Ekhorn: Haven't encountered this problem yet... if it even is a problem.

ui/component/or-map/src/asset-map.ts
```diff
@@ -0,0 +1,301 @@
...
+        const features = this._map!.querySourceFeatures("assets");
+
+        if (!this._hasRequired(asset) || !this._isMissing(asset, features)) {
+            return;
+        }
+
+        if (!(asset.type in this._assetTypeColors)) {
+            this._assetTypeColors[asset.type] = getMarkerIconAndColorFromAssetType(asset.type)?.color as string;
+            this._source.workerOptions.clusterProperties = this._getClusterProperties();
+        }
```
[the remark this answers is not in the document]
Ekhorn: If it breaks, we'll find out.

ui/app/manager/src/pages/page-map.ts
```diff
@@ -292,9 +307,18 @@ export class PageMap extends Page<MapStateKeyed> {
...
                 this._store.dispatch(setAssets(assets));
+                this._map?.addAssets(this._assets);
 
                 const assetSubscriptionId = await manager.events.subscribeAssetEvents(undefined, false, (event) => {
                     this._store.dispatch(assetEventReceived(event));
+                    switch (event.cause) {
+                        case "DELETE": this._map?.removeAssets([event.asset.id]); break;
+                        case "CREATE":
+                            if (MapUtil.isAssetWithLocation(event.asset)) {
+                                this._map?.addAsset(event.asset);
```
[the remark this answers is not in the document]
Ekhorn: Resolved in https://github.com/openremote/openremote/pull/2610/commits/102cb8690cb515a817a7ddb1ce5b921be7b96503

PR #2621, opened 2026-03-17
Bugfix for public asset permissions

richturner:
Closes https://github.com/openremote/team-private/issues/2

### DIFF
--- a/test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
+++ b/test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
@@ -304,13 +304,13 @@ class WebsocketClientTest extends Specification implements ManagerContainerTrait
             assert receivedMessages.size() == 2
             def attrSub = receivedMessages.find{it instanceof TriggeredEventSubscription && it.subscriptionId == "attributes"} as TriggeredEventSubscription<AttributeEvent>
             def assetSub = receivedMessages.find{it instanceof TriggeredEventSubscription && it.subscriptionId == "assets"} as TriggeredEventSubscription<AssetEvent>
-            assert attrSub != null
             assert assetSub != null
-            assert attrSub.events.size() == 1
             assert assetSub.events.size() == 1
             assert assetSub.events.get(0).id == managerTestSetup.apartment1LivingroomId
             assert assetSub.events.get(0).attributeNames.size() == 8
             assert !assetSub.events.get(0).attributeNames.contains("testAttribute")
+            assert attrSub != null
+            assert attrSub.events.size() == 1
             assert attrSub.events.get(0).id == managerTestSetup.apartment1LivingroomId
             assert attrSub.events.get(0).name == Asset.NOTES.name
         }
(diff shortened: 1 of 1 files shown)


test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
```diff
@@ -129,11 +134,25 @@ class WebsocketClientTest extends Specification implements ManagerContainerTrait
...
+                    new StringEncoder(CharsetUtil.UTF_8),
+                    new StringDecoder(CharsetUtil.UTF_8),
+                    new AbstractNettyIOClient.MessageToMessageDecoder<String>(String.class, client3)
+            ].toArray(new ChannelHandler[0])
+        })
+
         and: "we add callback consumers to the clients"
         def connectionStatus = client.getConnectionStatus()
         def connectionStatus2 = client2.getConnectionStatus()
+        def connectionStatus3 = client2.getConnectionStatus()
```
ebariaux:
```suggestion
        def connectionStatus3 = client3.getConnectionStatus()
```

test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
```diff
@@ -239,25 +286,101 @@ class WebsocketClientTest extends Specification implements ManagerContainerTrait
...
+            def attrSub = receivedMessages.find{it instanceof TriggeredEventSubscription && it.subscriptionId == "attributes"} as TriggeredEventSubscription<AttributeEvent>
+            def assetSub = receivedMessages.find{it instanceof TriggeredEventSubscription && it.subscriptionId == "assets"} as TriggeredEventSubscription<AssetEvent>
+            assert attrSub != null
+            assert assetSub != null
+            assert attrSub.events.size() == 1
+            assert assetSub.events.size() == 1
+            assert assetSub.events.get(0).id == managerTestSetup.apartment1LivingroomId
+            assert assetSub.events.get(0).attributeNames.size() == 8
+            assert !assetSub.events.get(0).attributeNames.contains("testAttribute")
+            assert attrSub.events.get(0).id == managerTestSetup.apartment1LivingroomId
```
ebariaux, line 314: Why those 2 asserts not just after the attrSub size assert ?

test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
```diff
@@ -239,25 +286,101 @@ class WebsocketClientTest extends Specification implements ManagerContainerTrait
...
         def apartment1Livingroom = assetStorageService.find(managerTestSetup.apartment1LivingroomId)
         apartment1Livingroom.addAttributes(
-                new Attribute<>("testAttribute", ValueType.BOOLEAN)
+                new Attribute<>("testAttribute", ValueType.BOOLEAN).addOrReplaceMeta(
+                        new MetaItem<>(MetaItemType.ACCESS_RESTRICTED_READ, false),
+                        new MetaItem<>(MetaItemType.ACCESS_PUBLIC_READ, false)
+                )
+        )
+        apartment1Livingroom.getAttribute(Asset.NOTES).get().addOrReplaceMeta(
+                new MetaItem<>(MetaItemType.ACCESS_PUBLIC_READ)
```
ebariaux, line 298: This is adding a MetaItem not providing a value, which means value is null.
For a boolean, shouldn't this mean false by default ?
If I understand code below, this attribute becomes readable by anonymous user.
I haven't traced the code path that makes this happens but it does not feel right to me.

Ekhorn:
Unrelated, but it seems there is 1 other instance of this where we incorrectly handle boolean meta items.

https://github.com/openremote/openremote/blob/29cce9419a5c7415d3f56886fdced10bb0a242e3/extension/energy/src/main/java/org/openremote/extension/energy/manager/EnergyOptimisationService.java#L805

richturner:
> Unrelated, but it seems there is 1 other instance of this where we incorrectly handle boolean meta items.
> 
> https://github.com/openremote/openremote/blob/29cce9419a5c7415d3f56886fdced10bb0a242e3/extension/energy/src/main/java/org/openremote/extension/energy/manager/EnergyOptimisationService.java#L805

Seems to be a few usages like this including in the `ApplyPredictedDataPointsService` via the `hasRequiredMeta` check

test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
richturner: See MetaItem:L74

test/src/test/groovy/org/openremote/test/protocol/websocket/WebsocketClientTest.groovy
richturner: I have cleaned up the assertion ordering

richturner:
> Unrelated, but it seems there is 1 other instance of this where we incorrectly handle boolean meta items.
> 
> https://github.com/openremote/openremote/blob/29cce9419a5c7415d3f56886fdced10bb0a242e3/extension/energy/src/main/java/org/openremote/extension/energy/manager/EnergyOptimisationService.java#L805

Now created the following issue:

https://github.com/openremote/openremote/issues/2622

PR #2646, opened 2026-03-24
Fix deadlock in AbstractNettyIOClient

richturner:
During testing I encountered a deadlock where the test runner thread called the `AbstractNetttyIOClient.disconnect()` method and a deadlock occurred. This is a result of the Failsafe library being used also using locks and interfering with the synchronisation logic in `AbstractNettyIOClient`, this PR fixes that issue.


wborn:
@codex review

PR #2650, opened 2026-03-25
Replace obsolete Keycloak adapter with Nimbus based JWT token verifier

### LINKED ISSUE #2011: Replace keycloak adapter with jjwt
The `keycloak-adapter` is no longer maintained and has several vulnerabilities, we should move to something like `jjwt` and wire it up to `JAX-RS` resteasy and undertow.

richturner:
## Pull request overview
This PR migrates the server/container security integration away from the obsolete Keycloak Undertow adapter toward a Nimbus-based JWT verification approach, while also consolidating outbound JAX-RS client usage around a shared WebTargetBuilder client. It updates affected APIs and tests to match the new authentication/token handling and resource method return types.

Changes:

- Introduce JWT-based servlet authentication plumbing (token verifier, key resolver, JWT auth filter) and remove Undertow Keycloak adapter integration classes.
- Refactor multiple services/protocols/tests to use WebTargetBuilder.getClient() and to avoid leaking JAX-RS Response objects.
- Adjust model/manager APIs (e.g., resource methods returning domain objects instead of Response) and update tests accordingly.
- `AuthContext.hasResourceRoleOrIsSuperUser()` removed - use `AuthContext.isSuperUser() || AuthContext.hasResourceRole()` instead
- Various method refactors in `WebService`:
  - `getStandardProviders` removed realmIndex param
  - `configureDeployment` replaced with `configureServlet`
- KeycloakSecurityContext rename to AuthContext on the MQTTHandler methods: checkCanSubscribe, checkCanPublish, canSubscribe, canPublish
- KeycloakIdentityProvider getTokenUri() now protected to prevent Identity provider implementation leakage

- Closes #2011 
- Depends on #2656


container/src/main/java/org/openremote/container/security/TokenVerifierImpl.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        JWTProcessor<SecurityContext> jwtProcessor = getJwtProcessor(realm);
+        JWTClaimsSet claimsSet;
+        TokenPrincipal principal;
+
+        // Process the token. This verifies the signature and validates the claims
+        // We also perform super user cross realm checks
+        try {
+            claimsSet = jwtProcessor.process(token, null);
```
Ekhorn, line 129: The UI app tests are failing, because the token that's passed to the verify method isn't a "compact-encoded URL-safe string" i.e. the token shouldn't be prefixed with `Bearer `. We could do something like the following to fix this, but I'm not sure how we would want to handle this.

```java
    private static final Pattern BEARER_PATTERN = Pattern.compile("(?i)^Bearer\\s+");
```

```suggestion
            String compactJwt = BEARER_PATTERN.matcher(token.trim()).replaceFirst("");
            claimsSet = jwtProcessor.process(compactJwt, null);
```

I ran all UI apps tests with these changes locally and they now pass, besides 1. Could there be a change to Keycloak theming? Maybe it's just local issues or a flaky test..

richturner: Thanks for investigating @Ekhorn. There was a breaking change introduced into Websocket auth handling that has now been resolved I think that explains what you were seeing and no changes are needed in the `TokenVerifierImpl`

wborn:
@codex review

wborn:
@codex review

container/src/main/java/org/openremote/container/security/basic/BasicIdentityProvider.java
```diff
@@ -148,34 +67,8 @@ public void stop(Container container) {
...
-            }
-
-            @Override
-            public Account verify(Credential credential) {
-                return null;
-            }
-        });
+    public FilterRegistration.Dynamic secureDeployment(ServletContext servletContext) {
+       throw new UnsupportedOperationException("Basic authentication is not supported for servlet deployments");
     }
```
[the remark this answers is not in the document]
richturner: Deprecated an unused so ignoring

richturner:
@Ekhorn all copilot feedback taken on board and PR updated

agent/build.gradle
```diff
@@ -65,6 +65,8 @@ dependencies {
     testImplementation "org.mockito:mockito-core:$mockitoVersion"
     testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitVersion"
     testRuntimeOnly "org.junit.platform:junit-platform-launcher"
+
+    implementation "org.bouncycastle:bcprov-jdk18on:$bouncyCastleVersion"
```
Ekhorn, line 69: Apparently there is a 5.5 CVE in this current version. Resolved in 1.84 (ref https://nvd.nist.gov/vuln/detail/CVE-2026-0636)

container/src/main/java/org/openremote/container/security/basic/BasicIdentityProvider.java
Ekhorn, line 1: Don't forget to include the breaking changes for the release notes. Maybe good to put them in the PR description.

container/src/main/java/org/openremote/container/security/KeyResolver.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class KeyResolver {
+
+    private String keycloakBaseUrl;
```
Ekhorn, line 34:
```suggestion
    private final String keycloakBaseUrl;
```

container/src/main/java/org/openremote/container/security/TokenPrincipal.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public List<String> getRealmRoles() {
+        return realmRoles;
+    }
+
+    public List<String> getResourceRoles(String resource) {
+        return resourceRoles.get(resource);
+    }
+
+    // TODO: Decide the role lookup logic
```
Ekhorn, line 86: Can't this be removed? Or did you have the intention for a follow up?

```suggestion
```

container/src/main/java/org/openremote/container/security/TokenVerifierImpl.java
```diff
@@ -0,0 +1,147 @@
...
+
+        processor.setJWTClaimsSetVerifier(verifier);
+        return processor;
+    }
+
+    @Override
+    public TokenPrincipal verify(String realm, String token) throws AuthenticationException {
+
+        if (realm == null || realm.isBlank()) {
+            throw new AuthenticationException("Invalid token");
```
Ekhorn, line 116:
```suggestion
            throw new AuthenticationException("Invalid realm");
```

container/src/main/java/org/openremote/container/security/TokenVerifierImpl.java
```diff
@@ -0,0 +1,147 @@
...
+        return processorCache.computeIfAbsent(realm, this::initialiseProcessor);
+    }
+
+    protected JWTProcessor<SecurityContext> initialiseProcessor(String realm) {
+        DefaultJWTProcessor<SecurityContext> processor = new DefaultJWTProcessor<>();
+
+        // 1. Get the JWK source for the realm from our service
+        JWKSource<SecurityContext> keySource = keyResolverService.getJwkSource(realm);
+
+        // 3. Configure the processor with a key selector for the appropriate JWS algorithm (e.g., RS256)
```
Ekhorn, line 66:
```suggestion
        // 1. Get the JWK source for the realm from our service
        JWKSource<SecurityContext> keySource = keyResolverService.getJwkSource(realm);

        // 2. Configure the processor with a key selector for the appropriate JWS algorithm (e.g., RS256)
```

container/src/main/java/org/openremote/container/security/TokenVerifierImpl.java
```diff
@@ -0,0 +1,147 @@
...
+
+        final String expectedIssuer = keycloakPublicUrl + "/realms/" + realm;
+        final String expectedClientId = Constants.KEYCLOAK_CLIENT_ID;
+        final JWTClaimsSetVerifier<SecurityContext> defaultClaimsVerifier = new DefaultJWTClaimsVerifier<>(null, null);
+
+        JWTClaimsSetVerifier<SecurityContext> verifier = (claims, context) -> {
+            // Preserve Nimbus built-in checks for exp / nbf validation.
+            defaultClaimsVerifier.verify(claims, context);
+
+            // 1) issuer check (keep/remove depending on your policy)
```
Ekhorn, line 79:
```suggestion
            // issuer check (keep/remove depending on your policy)
```

container/src/main/java/org/openremote/container/security/TokenVerifierImpl.java
Ekhorn, line 1: None of the exceptions thrown here are covered by the tests according to the JaCoCo report. I think it is important we prevent regressions for this security related code.

container/src/main/java/org/openremote/container/security/JWTAuthenticationFilter.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        if (!(request instanceof HttpServletRequest httpRequest) || !(response instanceof HttpServletResponse httpResponse)) {
+            chain.doFilter(request, response);
+            return;
+        }
+
+        String realm = httpRequest.getHeader(Constants.REALM_PARAM_NAME);
+        if (realm == null) {
+            httpResponse.sendError(HttpServletResponse.SC_BAD_REQUEST, Constants.REALM_PARAM_NAME + " header is missing");
```
Ekhorn, line 57: This branch is missed in the tests according to JaCoCo.

container/src/main/java/org/openremote/container/security/JWTAuthenticationFilter.java
```diff
@@ -0,0 +1,97 @@
...
+                @Override
+                public String getAuthType() {
+                    return AUTH_TYPE;
+                }
+            };
+
+            chain.doFilter(authenticatedRequest, response);
+
+        } catch (AuthenticationException e) {
+            throw new NotAuthorizedException("Authentication has failed: " + e.getMessage(), Response.status(Response.Status.UNAUTHORIZED).build(), e);
```
Ekhorn, line 94: This branch is missed in the tests according to JaCoCo.

container/src/main/java/org/openremote/container/security/basic/BasicIdentityProvider.java
Ekhorn, line 1: Apparently none of the basic auth code is covered by tests (according to JaCoCo), but as this is deprecated not sure it is worth the effort to test this. Would be good to at least check the `verifyAccount` flow here.

container/src/main/java/org/openremote/container/security/keycloak/KeycloakIdentityProvider.java
```diff
@@ -431,18 +351,63 @@ public static String getSubjectNameAndRealm(Subject subject) {
             }).orElse(null);
     }
 
-    public static String getSubjectId(Subject subject) {
-        if (subject == null || subject.getPrincipals() == null) {
-            return null;
-        }
+    public static boolean isSuperUser(KeycloakSecurityContext securityContext) {
+        return securityContext != null && Constants.MASTER_REALM.equals(securityContext.getRealm()) && securityContext.getToken().getRealmAccess().isUserInRole(Constants.SUPER_USER_REALM_ROLE);
```
Ekhorn, line 354: According to JaCoCo `isSuperUser` isn't covered by tests.

container/src/main/java/org/openremote/container/security/AuthContext.java
Ekhorn, line 63: Unused, could be removed. `AssetPredictedDatapointService` now explicitly writes out this check rather than using this method.

```suggestion
```

Ekhorn (commented):
Nice refactor @richturner! I think it would be nice to improve some of the test coverage. I'll be doing some manual testing.

agent/build.gradle
richturner: I'll update I just took the same version as was being used as an indirect dependency previously, will see where this is used elsewhere and force the newer version

container/src/main/java/org/openremote/container/security/basic/BasicIdentityProvider.java
richturner: It's not used anywhere so I'm not bothered about this

container/src/main/java/org/openremote/container/security/TokenVerifierImpl.java
richturner: @Ekhorn The coverage in tests is at the integration level where authentication/authorization is tested and the HTTP responses verified, unless I'm missing something, this sounds more like unit testing which we don't do.

container/src/main/java/org/openremote/container/security/JWTAuthenticationFilter.java
richturner: This is very hard to cover as the `RealmPathExtractorFilter` assumes the part of the path after `/api/` is the realm name we when using a JAX-RS client proxy the path of the resource becomes the realm (e.g. `AssetResource -> asset`) and authentication will fail but on Token verification. I could formulate in invalid request with no resource in the path but this seems redundant.

container/src/main/java/org/openremote/container/security/JWTAuthenticationFilter.java
richturner: Not sure how this branch can be missed as it is the main path through this Filter and executes on every auth request

container/src/main/java/org/openremote/container/security/keycloak/KeycloakIdentityProvider.java
richturner: Removed as AuthContext method is used instead

richturner:
Hey @Ekhorn thanks for the review, comments taken onboard where I could, please see comments on items I don't think were applicable

wborn:
@codex review

richturner:
@Ekhorn feedback comments now incorporated

Ekhorn (approved):
Went through everything to see if there are anymore breaking changes we maybe missed. Many custom projects will need to migrate the following things:

- `KeycloakSecurityContext` -> `AuthContext` on the MQTTHandler methods: `checkCanSubscribe`, `checkCanPublish`, `canSubscribe`, `canPublish` 
- `WebService.getStandardProviders` -> `realmIndex` removal (like you mention in the description)

Nice work @richturner! I'll let the final button press to you.

PR #2655, opened 2026-03-27
Websocket connection initialisation threading issue fix

richturner:
The camel route used for Websocket sessions incorrectly uses the `threads` DSL on the session events which meant that `ONOPEN` can be handled after incoming messages, this PR fixes this


wborn:
@codex review

richturner:
The review has been helpful and I believe I have now resolved the highlighted issue

ebariaux:
Fix made sense to me but Codex found another potential issue

• The route refactor introduces regressions around websocket lifecycle handling: disconnect work now runs on Undertow I/O threads, and gateway disconnects can overtake already-queued gateway messages. Those issues make the patch unsafe under realistic subscription and disconnect scenarios.

  Full review comments:

  - [P2] Keep close/error cleanup off the Undertow I/O thread — /Users/ebariaux/Development/OpenRemote/openremote-dependency-update/manager/src/main/java/org/openremote/manager/event/ClientEventService.java:171-174
    If a websocket has many active subscriptions, this change now executes the ONCLOSE/ONERROR branches on Undertow's I/O thread. Those branches synchronously iterate the session's subscriptions and call removeSubscription() for each consumer, and removeSubscription() does a
    CopyOnWriteArraySet.removeIf(...), so disconnecting a heavily subscribed client can spend a long time copying the global subscription set and stall unrelated websocket traffic. The previous route-level threads() hop avoided doing this expensive teardown on the I/O thread.
  - [P2] Preserve gateway message ordering relative to disconnects — /Users/ebariaux/Development/OpenRemote/openremote-dependency-update/manager/src/main/java/org/openremote/manager/event/ClientEventService.java:262-263
    Because only the otherwise() branch is now dispatched to executorService, a gateway socket's ONCLOSE can run before earlier SharedEvent frames from that same connection have been drained from the worker queue. In that case GatewayService.processGatewayDisconnected() clears the connector
    session first, and processGatewayMessage() later drops the queued event as belonging to an obsolete session. A gateway that sends a final event and then closes can therefore lose that last event.



So I asked it for a proposed fix given the initial reason for the code change

 I’d fix this by enforcing FIFO per websocket session, not by special-casing ONOPEN. The original bug happened because the route-level async hop let ONOPEN and later frames run in parallel; the current change at ClientEventService.java (#/Users/ebariaux/Development/OpenRemote/openremote-
  dependency-update/manager/src/main/java/org/openremote/manager/event/ClientEventService.java#L171) fixes that only by making lifecycle events inline, which then breaks close/message ordering and puts teardown back on Undertow I/O threads.

  My proposal:

  - Restore a single async handoff for this route in ClientEventService.java (#/Users/ebariaux/Development/OpenRemote/openremote-dependency-update/manager/src/main/java/org/openremote/manager/event/ClientEventService.java#L171), but replace Camel threads() with a per-session serial dispatcher
    keyed by UndertowConstants.CONNECTION_KEY.
  - Send ONOPEN, normal frames, ONCLOSE, and ONERROR through that same queue. That preserves the ordering assumptions in GatewayService.java (#/Users/ebariaux/Development/OpenRemote/openremote-dependency-update/manager/src/main/java/org/openremote/manager/gateway/GatewayService.java#L340) and
    GatewayService.java (#/Users/ebariaux/Development/OpenRemote/openremote-dependency-update/manager/src/main/java/org/openremote/manager/gateway/GatewayService.java#L697).
  - Do session cleanup only in the queued close/error task. Once a terminal event runs, mark the session closed and drop anything that arrives after it.

  In other words: the invariant should be “all work for one websocket runs off the I/O thread, in arrival order.” That fixes both races: frames can’t overtake ONOPEN, and disconnects can’t overtake already-received frames.

  A simple implementation is a keyed tail-future/serial-executor pattern:

  sessionTails.compute(sessionKey, (key, tail) ->
      (tail == null ? CompletableFuture.completedFuture(null) : tail)
          .handle((r, ex) -> null)
          .thenRunAsync(() -> processInbound(exchangeCopy), executorService)
  );

  I would not keep threads() inside otherwise(). That just trades the original ONOPEN race for a ONCLOSE/gateway-ordering race. If you want, I can turn this into a concrete patch and add regression tests for “subscribe then immediate disconnect” and “final gateway event then close.”

Think that makes sense, multi-threading per web socket but serialized for 1 connection.

richturner:
Seems like I opened up a can of worms on this issue.
Taken all feedback on board and made the adjustment proposed

ebariaux:
That looked nice to me but Codex seems to find more potential issues, although I might consider they're improvements and could be part of another PR as I believe current situation is already better than before.

The refactor improves ordering, but it also introduces an unbounded in-memory queue per websocket session and moves shared-event fan-out onto that same serialized path. Those regressions can break availability and disconnect handling under realistic slow-client or slow-subscriber conditions.

  Full review comments:

  - [P1] Restore backpressure for per-session websocket work — /Users/ebariaux/Development/OpenRemote/openremote-dependency-update/manager/src/main/java/org/openremote/manager/event/ClientEventService.java:179-190
    When one client sends frames faster than processInbound can drain them, this sessionTails chain appends a new stage that retains exchangeCopy for every frame. Because those stages are only scheduled after the previous one completes, the route now accepts an unbounded per-session backlog in
    heap instead of applying the immediate executor backpressure that threads().executorService(...) provided before. A noisy websocket or gateway client can therefore grow memory linearly with pending frames and eventually OOM the manager.
  - [P2] Keep inbound shared events on the publish queue — /Users/ebariaux/Development/OpenRemote/openremote-dependency-update/manager/src/main/java/org/openremote/manager/event/ClientEventService.java:410-410
    sendToSubscribers(event) now runs every shared-event consumer inline on the session tail. For websocket requests such as ReadAssetEvent, ReadAssetsEvent, or ReadAssetTreeEvent, that means the storage work in AssetStorageService happens before the next frame from that socket can be handled, so
    UNSUBSCRIBE/ONCLOSE and gateway disconnect cleanup are delayed until the read finishes. The previous .to(PUBLISH_QUEUE) handoff kept that fan-out off the per-session processing path.
    
> What plan would you propose to tackle those 2 issues based on the current state of the code ?

Based on the current code, I’d keep the refactor’s structure in manager/src/main/java/org/openremote/manager/event/ClientEventService.java:209, and only change two things: how inbound websocket work is queued at manager/src/main/java/org/openremote/manager/event/ClientEventService.java:175, and
  how non-attribute SharedEvents are handed off at manager/src/main/java/org/openremote/manager/event/ClientEventService.java:390. I would not touch manager/src/main/java/org/openremote/manager/gateway/GatewayService.java:340 unless tests show a remaining race there.

  1. Replace sessionTails with a bounded per-session dispatcher.
     The current CompletableFuture chain at manager/src/main/java/org/openremote/manager/event/ClientEventService.java:179 gives ordering, but it is also an unbounded heap queue.
     I’d switch to an explicit SessionDispatchState per sessionKey: a small ArrayDeque, a draining flag, and a closed/closing flag.
     Only one drain task per session gets submitted to executorService; that task loops and processes queued exchanges in order.
     The queue must be bounded. On overflow, I’d close the websocket and drop further data rather than keep buffering indefinitely.
  2. Keep using the current container executor for actual work, but stop using heap as the queue.
     container/src/main/java/org/openremote/container/Container.java:118 shows container.getExecutor() is a ThreadPoolExecutor with SynchronousQueue and CallerRunsPolicy.
     That means the system already has “no hidden executor queue” semantics. The fix should preserve that by having at most one runnable per session submitted to the executor, not one CompletableFuture stage per frame.
  3. Restore the publish handoff for non-attribute SharedEvents.
     At manager/src/main/java/org/openremote/manager/event/ClientEventService.java:410, replace inline sendToSubscribers(event) with a handoff to the existing publish route at manager/src/main/java/org/openremote/manager/event/ClientEventService.java:195.
     The per-session dispatcher should only do ordering-sensitive work: parse, auth, subscription bookkeeping, gateway interception, and deciding where the event goes.
     Fan-out belongs on PUBLISH_QUEUE, not on the serialized session path.
  4. Add one defensive closed-session check while touching this code.
     In the subscription-add path at manager/src/main/java/org/openremote/manager/event/ClientEventService.java:344, I’d re-add a “session still alive” guard before creating a consumer.
     With bounded queues and forced closes on overload, that is useful defense-in-depth against late work recreating session state.
  5. Add focused regression tests before or alongside the patch.
     One test for ONOPEN -> frame -> ONCLOSE ordering.
     One test for a gateway’s final SharedEvent not being lost before disconnect cleanup.
     One test that a blocked session dispatcher hits the queue cap and closes instead of growing unbounded state.
     One test that a slow ReadAssetsEvent fan-out does not block UNSUBSCRIBE or ONCLOSE once the event has been handed to PUBLISH_QUEUE.

  If you want the smallest safe patch, I’d do steps 1 and 3 first, then add the closed-session guard and tests in the same change.

richturner:
I have taken the latest feedback on board and made another refactor which should address all of the points

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -203,11 +231,54 @@ public void configure() throws Exception {
...
+           }
+       }
+   }
+
+    /**
+     * Handles all inbound websocket exchanges
+     */
+    protected void processInbound(Exchange exchange) {
       UndertowConstants.EventType eventType = exchange.getIn().getHeader(UndertowConstants.EVENT_TYPE_ENUM, UndertowConstants.EventType.class);
+      boolean isTerminalEvent = (eventType == UndertowConstants.EventType.ONCLOSE || eventType == UndertowConstants.EventType.ONERROR);
```
ebariaux: This variable is not used in this method
```suggest
```

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -174,21 +183,40 @@ public void configure() throws Exception {
...
+               boolean isTerminalEvent = (eventType == UndertowConstants.EventType.ONCLOSE || eventType == UndertowConstants.EventType.ONERROR);
+
+               synchronized (state) {
+                   // 1. Drop normal frames if closed, but ALWAYS let terminal events through
+                   if (state.closed && !isTerminalEvent) {
+                       exchange.setRouteStop(true);
+                       return;
+                   }
+
+                   // 2. Enforce Backpressure (only for non-terminal events)
```
ebariaux: Would change the comment because it's not back pressure, it's overload protection that kills the connection.
It also clears the queue, so last MAX_QUEUE_SIZE messages are just lost.

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -174,21 +183,40 @@ public void configure() throws Exception {
...
+                       exchange.setRouteStop(true);
+                       return;
+                   }
+
+                   // 3. Queue the frame and trigger the drainer if dormant
+                   // Create a safe copy of the exchange for async processing
+                   state.queue.add(exchange.copy());
+                   if (!state.draining) {
+                       state.draining = true;
+                       executorService.submit(() -> drainSession(sessionKey, state));
```
ebariaux: From Codex
  - [P2] Clear the draining flag when submit() is rejected — /Users/ebariaux/Development/OpenRemote/openremote-version-check/manager/src/main/java/org/openremote/manager/event/ClientEventService.java:214-216
    If the shared executorService rejects this submit call (for example while the pool is saturated or shutting down), state.draining stays true and the copied frame remains queued. From that point on the session never schedules another drainer, so subsequent websocket frames
    for that session just accumulate until MAX_QUEUE_SIZE is hit and then get dropped.


In theory indeed submit() can throw exceptions, including `RejectedExecutionException`.
That would eventually be handled in `AssetProcessingService`, line 209 just logging the issue.
So comment seems legit to me, maybe just handling exceptions to turn draining back to false and then re-throwing is the fix.

richturner: Thanks for the review. Without having access to the code right now I believe executorservice uses a Caller Executes rejection policy and the only time the executor shuts down is when the container is stopping.

Catching the exception and setting draining back to false still wouldn't mean the remaining queued messages would be processed unless another message arrives would it?

richturner: I've finished a slight refactor taking on board the points highlighted in the reviews

PR #2664, opened 2026-03-31
Make UI app tests wait for the manager to be healthy

### LINKED ISSUE #2323: Make UI App tests wait for healthy containers locally
## User Story
As a developer I want to to run `gradle npmTest`, so that I can run them locally without the tests immediately failing while containers start.

## Description

Playwright UI app tests should wait for the required app backend to be ready.

The pipeline currently handles this through the following script https://github.com/openremote/openremote/blob/a803cc68ab97a18ba7ed9a2f928fc15b049c0748/.github/workflows/ci_cd.yml#L807

https://github.com/openremote/openremote/blob/a803cc68ab97a18ba7ed9a2f928fc15b049c0748/.ci_cd/host_init/healthy.sh#L3

## Acceptance Criteria
- [ ] 1. All required containers must be started before tests run:
  - [ ] postgres
  - [ ] keycloak
  - [ ] manager 

## Implementation Notes

There are options for the web server but its not clear whether any of these will https://playwright.dev/docs/test-webserver#configuring-a-web-server make it wait for the server.

Preferably we can just wait for https://docs.openremote.io/docs/rest-api/get-health-status.

We can always increase the timeout for `setup.test.ts` and check for healthy containers (either through and endpoint or nodejs script).

Ekhorn:
## Description

This ensures that the tests wait for the manager to be healthy. It retries for a minute until the manager is healthy using the following interval: 

1. Wait 5 seconds
2. If not healthy try 1 second later
3. Still not try 2 seconds later
4. Otherwise continue to try every 5 seconds

We'll also be needing this in the upcoming pipeline changes. 

- Closes #2323 

## Checklist

### DIFF
--- a/ui/app/manager/test/test.setup.ts
+++ b/ui/app/manager/test/test.setup.ts
@@ -9,25 +9,28 @@ import permissions from "./fixtures/data/permissions";
 
 const { admin, smartcity } = users; 
 
+// Must always run tests in setup sequentially to ensure the "Wait for manager to be healthy"
+// setup test runs first
 setup.describe.configure({ mode: "serial" });
 
-setup(`Login as "admin" user`, async ({ page, manager, context }) => {
-    const { username, password } = admin;
+setup(`Wait for the manager to be healthy`, async ({ manager }) => {
     await expect.poll(async () => {
         const noop = () => null;
-        const token = await manager.getAccessToken("master", username, password).catch(noop);
+        const token = await manager.getAccessToken("master", "admin", "secret").catch(noop);
         const response = await manager.api.StatusResource.getHealthStatus({
-            headers: { "Authorization": "Bearer " + token },
+            headers: { "Authorization": "Bearer " + token }
         }).catch(noop);
         return response?.status;
     }, {
         message: "Manager to become healthy",
         intervals: [5_000, 1_000, 2_000, 5_000],
-        timeout: 60_000,
+        timeout: 60_000
     }).toBe(200);
+});
 
+setup(`Login as "admin" user`, async ({ page, manager, context }) => {
     await manager.goToRealmStartPage("master");
-    await manager.login("admin");
+    await manager.login(admin.username);
     await page.waitForURL("**/manager/**");
     await page.waitForTimeout(2000);
     await context.storageState({ path: adminStatePath });


ui/app/manager/test/test.setup.ts
```diff
@@ -1,58 +1,81 @@
...
+    adminStatePath,
+    test as setup,
+    userStatePath,
+} from "./fixtures/manager";
 import { users } from "./fixtures/data/users";
 import permissions from "./fixtures/data/permissions";
 
+const { admin, smartcity } = users; 
+
+setup.describe.configure({ mode: "serial" });
```
Ekhorn, line 14: Might be able to remove this.

```suggestion

// Prevent setup tests to be run in parallel
setup.describe.configure({ mode: "serial" });
```

Ekhorn: This is indeed not needed, we explicitly do not run the UI app tests in parallel. But it cannot hurt to include this either..

ui/app/manager/test/test.setup.ts
```diff
@@ -1,58 +1,81 @@
...
+        intervals: [5_000, 1_000, 2_000, 5_000],
+        timeout: 60_000,
+    }).toBe(200);
+
+    await manager.goToRealmStartPage("master");
+    await manager.login("admin");
+    await page.waitForURL("**/manager/**");
+    await page.waitForTimeout(2000);
+    await context.storageState({ path: adminStatePath });
 });
```
Ekhorn, line 37:
```suggestion
// The wait for manager setup test must always be run first
setup.describe.configure({ mode: "serial" });

setup(`Wait for the manager to be healthy`, async ({ manager }) => {
    await expect.poll(async () => {
        const noop = () => null;
        const token = await manager.getAccessToken("master", "admin", "secret").catch(noop);
        const response = await manager.api.StatusResource.getHealthStatus({
            headers: { "Authorization": "Bearer " + token }
        }).catch(noop);
        return response?.status;
    }, {
        message: "Manager to become healthy",
        intervals: [5_000, 1_000, 2_000, 5_000],
        timeout: 60_000
    }).toBe(200);
});

setup(`Login as "admin" user`, async ({ page, manager, context }) => {
    await manager.goToRealmStartPage("master");
    await manager.login(admin.username);
    await page.waitForURL("**/manager/**");
    await page.waitForTimeout(2000);
    await context.storageState({ path: adminStatePath });
});
```

PR #2673, opened 2026-04-07
Feature: Notification management

### LINKED ISSUE #1601: Notifications: UI/UX improvements

### LINKED ISSUE #2742: UI: migrate notifications page to Vaadin

### LINKED ISSUE #2775: Notifications: Add read/write notifications roles
## Description

There are a few things we want to add on top of the current notifications management implementation.

## Acceptance Criteria
- [x] Split logic per realm
- [x] Role based access control (read/write for notifications)
	- [x] Super admin users can see all notifications
	- [x] Users should only see their own notifications (with `read:notifications`)
		- [x] user/asset details may only be retrieved if the user has the `read:assets` / `read:users` role
	- [x] Restricted users should not have the user/realm recipient option
	- [x] The "users recipient option" may only be used when the user has the `read:users` role

### LINKED ISSUE #2776: Notifications: add pagination
## Description

Currently all notifications are fetched, this list can grow very large. The table that shows them already shows them as if pagination is in use.

Ekhorn:
## Description

This PR introduces the new notification management page, allowing users to manually send notifications and read notifications that have been sent.

- Closes #1601
- Closes #2742
- Closes #2775
- Closes #2776
 
Also please see:
- #1615 and #1572

## Migration

Shows 10 million datapoints being migrated and the services/notifications roles being applied.

https://github.com/user-attachments/assets/d1158ffb-6624-43c4-8f27-7a7ebd6f20ea

## Changelog

  - Add Notifications management page (list, filter, view details, and send notifications from the UI)
  - Add notification count endpoint (`getNotificationsCount`)
  - Add server-side sorting and pagination for the notifications list (`sort`, `descending`, `offset`, `limit` query params)
  - Add source and date-range filtering to the notifications list (defaults to the current day)
  - Add `read:notifications` / `write:notifications` client roles, included in the `read`/`write` composite roles
  - Add a `realm` column to sent notifications (also returned in API responses), with migrations to backfill existing rows and create the new roles
  - Breaking: `POST /notification/alert` (send notification) now requires `write:notifications` (or `write:admin`); previously any authenticated user could call it and authorization was purely target-based - grant the role to any service accounts or custom-role users that send notifications
  - Breaking: `GET /notification` is now realm-scoped for non-superusers - results are limited to a realm the caller can access, defaulting to the caller's own realm when `realmId` is omitted; previously any `read:admin` caller could query notifications of any realm. Superusers are unchanged (any realm, or all realms when `realmId` is omitted). The endpoint also now accepts `read:notifications` as an alternative to `read:admin`
  - Breaking: the `realmId` query parameter of `GET /notification` now filters on the notification's `realm` column (all target types); previously it only matched REALM-targeted notifications
  - Restricted users can now read notifications (via `read:notifications`), limited to notifications they sent, that target them, or that target their realm

## Checklist


wborn:
@codex review

ui/component/or-asset-tree/src/or-add-asset-dialog.ts
Ekhorn, line 1: Check why this was done.

ui/component/or-asset-tree/src/style.ts
Ekhorn, line 1: Check why this was done.

ui/component/or-asset-tree/src/index.ts
Ekhorn, line 1: Check why this was done.

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,699 @@
...
+            </div>
+        `;
+    }
+
+    protected _renderHeader(writeNotifications: boolean) {
+        return html`
+            <div id="title">
+                <div style="display: flex; align-items: center;">
+                    <or-icon icon="message-outline" style="padding: 0 10px 0 4px;"></or-icon>
+                    <span><or-translate value="${i18next.t("notifications.notifications")}"/></span>
```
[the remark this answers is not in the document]
Ekhorn: You're wrong. The icon was in the main header of the application next to all the other menu items, but has since been moved to the dropdown menu. The issue calls for the removal of that menu item (not the icon itself). It's also already checked off.

ui/app/manager/src/components/notifications/notification-form.ts
```diff
@@ -0,0 +1,591 @@
...
+        name: '',
+        title: '',
+        body: '',
+        priority: 'Normal',
+        targetType: 'Asset',
+        targets: [''],
+        actionUrl: '',
+        openButtonText: '',
+        closeButtonText: ''
+    };
```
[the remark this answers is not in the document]
Ekhorn: Resolved in cbf91a12337c12616b601120fb9be4c701ff4bbc

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,699 @@
...
+                buttons: buttons.length > 0 ? buttons : undefined
+            };
+
+            const notification: Notification = {
+                name: formData.title,
+                message: message,
+                targets: formData.targets.map(id => ({
+                    type: formData.targetType as NotificationTargetType,
+                    id: id
+                }))
```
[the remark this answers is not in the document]
Ekhorn: Addressed in 3bf9f87ca3beb8b61b48fbea2ca4c3a87900377a

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,699 @@
...
+    public stateChanged(state: AppStateKeyed): void {
+        if (state.app.page === "Notifications") {
+            if (this.realm === undefined || this.realm === state.app.realm) {
+                this._loadData();
+            }
+            this.realm = state.app.realm;
+            this.requestUpdate('realm');
+
+            this._loadData();
+        }
```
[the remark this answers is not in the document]
Ekhorn: Addressed in c492e4a199c293fa397e04242ac0069a5c8150c0

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,699 @@
...
+                }))
+            };
+
+            const response = await this.notificationService.sendNotification(notification, this.realm);
+
+            showSnackbar(undefined, i18next.t("notifications.successfullySentNotification"));
+            dialog.close();
+        } catch (error) {
+            console.error("Error creating notification:", error);
+            showSnackbar(undefined, i18next.t("notifications.failedToCreateNotification", error));
```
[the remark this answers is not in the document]
Ekhorn: Key mismatch is resolved, will look at error messages later.

ui/app/manager/src/components/notifications/notification-form.ts
```diff
@@ -0,0 +1,591 @@
...
+            </div>
+        `;
+
+    }
+
+    protected _renderActionButtonContainer(inputDisabled: boolean) {
+        return html`
+            <div class="actionButtonContainer">
+                <h5>${i18next.t("Actions")}</h5>
+                <or-mwc-input
```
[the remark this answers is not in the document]
Ekhorn: This has been addressed.

ui/app/manager/src/components/notifications/or-notifications-table.ts
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        return super.willUpdate(changedProps);
+    }
+
+    protected getTableRows(notifications: SentNotification[]): TableRow[] | undefined {
+        notifications.forEach(notification => {
+            this.resolveTargetDetails(notification);
+        })
+
```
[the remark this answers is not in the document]
Ekhorn: This has been addressed

Ekhorn:
Found a bug when switching realms it remembers the assets/users from the previous realm. Maybe since dialog change?

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,699 @@
...
+            }
+            if (formData.closeButtonText) {
+                buttons.push({title: formData.closeButtonText});
+            }
+
+            const message: PushNotificationMessage = {
+                type: "push" as const,
+                title: formData.title,
+                body: formData.body,
+                data: {},
```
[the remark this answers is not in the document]
Ekhorn: This has been resolved since.

ui/component/or-asset-tree/src/or-add-asset-dialog.ts
Ekhorn: Image

Image

Ekhorn: I'll revert this unrelated to this PR.

ui/component/or-asset-tree/src/style.ts
Ekhorn: Prevents items from being no longer aligned

Image

ui/component/or-asset-tree/src/index.ts
Ekhorn: See https://github.com/openremote/openremote/pull/2673/changes#r3499785552

model/src/main/java/org/openremote/model/notification/NotificationResource.java
```diff
@@ -34,25 +34,28 @@
...
                                         @QueryParam("type") String type,
                                         @QueryParam("from") Long fromTimestamp,
                                         @QueryParam("to") Long toTimestamp,
                                         @QueryParam("realmId") String realmId,
                                         @QueryParam("userId") String userId,
-                                        @QueryParam("assetId") String assetId);
+                                        @QueryParam("assetId") String assetId,
+                                        @QueryParam("source") Notification.Source source,
+                                        @QueryParam("sort") SentNotification.SortField sort,
+                                        @QueryParam("descending") Boolean descending,
```
Ekhorn, line 56:
```suggestion
                                        @QueryParam("sort") SentNotification.SortField sort,
```

manager/src/main/resources/org/openremote/manager/setup/database/V20260701_1232__AddNotificationRealmAndRoles.sql
Ekhorn, line 1: rename file

ui/app/manager/src/components/notifications/or-notifications-table.ts
```diff
@@ -0,0 +1,461 @@
...
+            const event = new NotificationTableClickEvent(item.data.notification.id)
+            this.dispatchEvent(event);
+        }
+    }
+
+    async getRowCount(): Promise<number> {
+        return this.totalCount ?? await super.getRowCount();
+    }
+
+    async getPaginationControls(): Promise<TemplateResult> {
```
Ekhorn, line 455: Duplication, may be refactored?

wborn:
@codex review

manager/src/main/java/org/openremote/manager/notification/NotificationService.java
```diff
@@ -217,6 +247,7 @@ public void configure() throws Exception {
                                     .setTarget(target.getType())
                                     .setTargetId(target.getId())
                                     .setMessage(notification.getMessage())
+                                    .setRealm(resolveTargetRealm(target, notificationRealm.get()))
```
MartinaeyNL, line 254: If the `source` is `GLOBAL_RULESET`, then the `notificationRealm` will be null.
Wouldn't that mean, in the `resolveTargetRealm()` function, it will reset to `master` ?

Not sure what practice there is for GLOBAL rulesets, but I would assume it doesn't have a realm at all.

manager/src/main/java/org/openremote/manager/notification/NotificationService.java
```diff
@@ -321,21 +379,115 @@ public SentNotification getSentNotification(Long notificationId) {
...
+            parameters.add(assetIds);
+        } else if (userIds != null && !userIds.isEmpty()) {
+            builder.append(" AND n.target = ?").append(idx++).append(" AND n.targetId IN ?").append(idx++);
+            parameters.add(Notification.TargetType.USER);
+            parameters.add(userIds);
+        }
+
+        // Only restricted users are limited to notifications they sent, that target them, or that target their realm;
+        // any other caller with access (read:notifications / read:admin / superuser) sees all of them
+        if (authContext != null && identityService.getIdentityProvider().isRestrictedUser(authContext)) {
```
MartinaeyNL: Just thinking outside the box here; is this the file and/or place where "restricted user handling" should take place?
In case the authorization logic changes, we would have to replace this as well.
Maybe `restricted` as a boolean function parameter makes more sense?
(once again; just thinking outside of the box)

manager/src/main/java/org/openremote/manager/setup/database/V20260701_1232__AddNotificationRealmAndRoles.java
MartinaeyNL, line 1: What happens to the performance is hundreds or thousands of notifications are already present in the system?

manager/src/main/java/org/openremote/manager/setup/AbstractKeycloakSetup.java
MartinaeyNL, line 1: Maybe an odd question in this review phase; but do we have to include the new `READ_NOTIFICATIONS` and `WRITE_NOTIFICATIONS` in this PR? Can't we exclude them for now, and rely on `READ_ADMIN` and `WRITE_ADMIN `or superusers for the time being? Because I imagine this migration having a larger impact than we'd like.
(kind of thinking outside the box)

model/src/main/java/org/openremote/model/notification/NotificationResource.java
```diff
@@ -34,25 +34,28 @@
...
+     * pagination (offset/limit). Restricted users only ever see notifications they sent or that target them or their
+     * realm; non-superusers are limited to a realm they can access (their own when none is specified).
      */
     @GET
     @Produces(APPLICATION_JSON)
-    @RolesAllowed({Constants.READ_ADMIN_ROLE})
-    @Operation(operationId = "getNotifications", summary = "Retrieve all sent notifications by targets")
+    @RolesAllowed({Constants.READ_ADMIN_ROLE, Constants.READ_NOTIFICATIONS_ROLE})
+    @Operation(operationId = "getNotifications", summary = "Retrieve sent notifications matching the supplied criteria")
     SentNotification[] getNotifications(@BeanParam RequestParams requestParams,
```
MartinaeyNL, line 46: Btw, other than the breaking change in what roles are allowed;
Isn't the response of this HTTP endpoint any different after this PR is merged?
For example, does it return notifications from all realms or only the user' realm?
And is the behavior any different for superusers?

If it is, please list the breaking / behavior changes in the PR description.

model/src/main/java/org/openremote/model/notification/NotificationResource.java
MartinaeyNL: Once integrated, I think you can remove this ancient commented code

model/src/main/java/org/openremote/model/notification/NotificationResource.java
```diff
@@ -145,4 +149,22 @@ void notificationAcknowledged(@BeanParam RequestParams requestParams,
...
+    @RolesAllowed({Constants.READ_ADMIN_ROLE, Constants.READ_NOTIFICATIONS_ROLE})
+    @Operation(operationId = "getNotificationsCount", summary = "Count sent notifications matching the supplied criteria")
+    long getNotificationsCount(@BeanParam RequestParams requestParams,
+                               @QueryParam("type") String type,
+                               @QueryParam("from") Long fromTimestamp,
+                               @QueryParam("to") Long toTimestamp,
+                               @QueryParam("realmId") String realmId,
+                               @QueryParam("userId") String userId,
+                               @QueryParam("assetId") String assetId,
+                               @QueryParam("source") Notification.Source source);
```
MartinaeyNL, line 149: What's the exact use case of this?
I believe if we have pagination on the `GET notification` endpoint, that's enough right?

wborn:
@codex review

setup/src/notification/java/org/openremote/setup/notification/KeycloakNotificationSetup.java
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright 2025, OpenRemote Inc.
```
MartinaeyNL: You can upgrade these to 2026 😉

MartinaeyNL (commented):
Initial review of the API related changes. I'd like to get this out of the way first.
More to come, like a frontend review, but I'll do that later because it takes a lot of my time 😉

manager/src/main/java/org/openremote/manager/setup/AbstractKeycloakSetup.java
Ekhorn: That would have an impact on quite a few conditions throughout the backend and UI (including tests). I would rather we settle the discussion on this first then make changes.

model/src/main/java/org/openremote/model/notification/NotificationResource.java
Ekhorn: In theory we could get away with just the get endpoint, but the `or-mwc-table` component displays the max (count) value.

image

manager/src/main/java/org/openremote/manager/setup/database/V20260701_1232__AddNotificationRealmAndRoles.java
Ekhorn: The following is with 10million notifications :)

https://github.com/user-attachments/assets/61a0b16b-99cf-4d71-964c-672d26741b3d

manager/src/main/java/org/openremote/manager/notification/NotificationService.java
Ekhorn: You're right — `GLOBAL_RULESET` was the only source that left `notificationRealm` null. Made it explicit in 2c9d4a65:
`GLOBAL_RULESET` now sets `notificationRealm` to `Constants.MASTER_REALM`, matching what `INTERNAL` already did.

I did consider making `realm` nullable since a global ruleset genuinely has no realm, but kept it `NOT NULL` for a few reasons:

- `master` is already where OpenRemote puts everything system-level (superusers, internal sources), so mapping global rulesets there is consistent with the rest of the system
- A nullable authorization column is fragile — every query consumer has to decide what `NULL` means, and getting it wrong silently hides or exposes rows
- In practice very few rows land in `master` anyway — REALM/USER/ASSET targets always resolve their realm from the target itself, only truly unresolvable targets (CUSTOM, deleted asset/user) fall back

manager/src/main/java/org/openremote/manager/notification/NotificationService.java
Ekhorn: Addressed in 437b91812 — agreed, the query builder shouldn't be making authorization decisions. `getNotifications`, `getNotificationsCount` and `buildNotificationCriteria` now take a `boolean restricted` parameter; `NotificationResourceImpl` computes it via `isRestrictedUser(authContext)` at both call sites.

model/src/main/java/org/openremote/model/notification/NotificationResource.java
Ekhorn: Updated the PR description.

manager/src/main/java/org/openremote/manager/notification/NotificationService.java
```diff
@@ -321,21 +379,115 @@ public SentNotification getSentNotification(Long notificationId) {
...
+            builder.append(" AND (n.sourceId = ?").append(idx++)
+                .append(" OR (n.target = ?").append(idx++).append(" AND n.targetId = ?").append(idx++).append(")")
+                .append(" OR (n.target = ?").append(idx++).append(" AND n.targetId = ?").append(idx++).append("))");
+            parameters.add(authContext.getUserId());
+            parameters.add(Notification.TargetType.USER);
+            parameters.add(authContext.getUserId());
+            parameters.add(Notification.TargetType.REALM);
+            parameters.add(authContext.getAuthenticatedRealmName());
+        }
```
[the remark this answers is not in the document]
Ekhorn: The code is intentional — a REALM-targeted notification is a broadcast that was delivered to every user in that realm, including the restricted user, so showing them the sent record exposes nothing they didn't already receive.

The issue was in the PR description which understated the scope. Updated the changelog to accurately reflect the
behaviour: restricted users can see notifications they sent, that target them, or that target their realm.

wborn:
@codex review

manager/src/main/java/org/openremote/manager/notification/NotificationResourceImpl.java
```diff
@@ -175,7 +190,7 @@ protected void verifyAccess(SentNotification sentNotification, String targetId)
                 LOG.fine("DENIED: Anonymous request to update a notification sent to an asset that doesn't exist or isn't public");
                 throw new WebApplicationException("Anonymous request can only update public assets not linked to a user", FORBIDDEN);
             }
-
+            // TODO : What is meant by the message below?
```
MartinaeyNL: Seems that the code below is already present in the codebase above, see line 190-191.
I think you can simply remove it, or leave it commented; whatever you like.

setup/src/notification/java/org/openremote/setup/notification/ManagerNotificationSetup.java
```diff
@@ -0,0 +1,275 @@
...
+            .setConsoleProvider(PushNotificationMessage.TYPE, new ConsoleProvider(
+                "fcm", true, true, true, true, false, Map.of("token", "dev-fcm-token-" + user.getUsername())
+            ));
+        console.setRealm(realm);
+        console = assetStorageService.merge(console);
+
+        assetStorageService.storeUserAssetLinks(List.of(new UserAssetLink(realm, user.getId(), console.getId())));
+    }
+
+    private void persistNotification(String title, String body, String userId, String assetId, String realm, Instant sentOn) {
```
MartinaeyNL: Rename function to `persistPushNotification(...)`

setup/src/notification/java/org/openremote/setup/notification/ManagerNotificationSetup.java
```diff
@@ -0,0 +1,275 @@
...
+
+    private void persistEmailNotification(String subject, String html, String userId, String assetId, String realm, Instant sentOn) {
+        EmailNotificationMessage message = new EmailNotificationMessage()
+            .setSubject(subject)
+            .setHtml(html);
+
+        persist(subject, EmailNotificationMessage.TYPE, message, userId, assetId, realm, sentOn);
+    }
+
+    private void persist(String name, String type, AbstractNotificationMessage message, String userId, String assetId, String realm, Instant sentOn) {
```
MartinaeyNL: Rename function to `persistNotification(...)`

ui/component/theme/src/index.js
```diff
@@ -19,3 +19,17 @@
...
+ */
+export const globals = `
+    h4 {
+        margin: 0;
+        padding-bottom: var(--lumo-space-m);
+        font-size: var(--lumo-font-size-l);
+        font-weight: 600;
+        line-height: 125.303%;
+    }
+`;
```
MartinaeyNL: I'd like to vouch for using a different solution than a `globals` variable.
Because this means we have to put `unsafeCSS(globals)` in every individual file.
Especially when creating components we need to "remember" that this has to be added.

I do agree that we need some sort of solution for this.
It's a larger discussion point though.

ui/app/manager/src/components/notifications/or-notification-form.ts
```diff
@@ -0,0 +1,667 @@
...
+                    ${when(this._targetType === NotificationTargetType.ASSET,
+                            () => html`
+                                <or-asset-tree
+                                        id="asset-selector"
+                                        .selectedIds="${this._selectedAssetIds}"
+                                        .showSortBtn="${false}"
+                                        expandNodes
+                                        checkboxes
+                                        @or-asset-tree-selection="${(e: OrAssetTreeSelectionEvent) => this._onAssetSelectionChanged(e)}"
+                                ></or-asset-tree>
```
MartinaeyNL, line 552: Shouldn't the list of assets be limited to "console assets only" ?

ui/app/manager/src/components/notifications/notification-form.ts
```diff
@@ -0,0 +1,667 @@
...
+        :host([readonly]) .form-container {
+            opacity: 0.9;
+        }
+    `;
+
+    @state()
+    protected _message: NotificationMessage = {type: "push"};
+
+    @state()
+    protected _targetType: NotificationTargetType = NotificationTargetType.ASSET;
```
MartinaeyNL: Not sure if `NotificationTargetType.ASSET` is the most logical default?
If not done yet, please double check with Pierre or Remy.

ui/app/manager/src/pages/page-notifications.ts
MartinaeyNL, line 1: On switching realm, the list of assets in the notification dialog stays the same. Reproduce:
- Log in in Master realm with a superuser, navigate to Notifications, and click "Send new"
- The list of assets on the left hand side is correct
- Close the dialog, switch to Smart City and click "Send new" again.
- List of assets is still from Master

MartinaeyNL, line 1: I get `HTTP 400` when I try to send a push notification with the required fields.
However it's not reflected to the UI other than a snackbar.
I assume it's because I selected a "web browser" console asset, but that is not clear.

(email also doesn't work, probably due to no email address being configured)

Image

ui/app/manager/src/components/notifications/or-notification-form.ts
MartinaeyNL, line 1: I can add and delete assets from the asset tree in this window.

Image

ui/app/manager/src/pages/page-notifications.ts
MartinaeyNL: Although, this also fails with HTTP 400...

Image

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,670 @@
...
+                    id="createDialog"
+                    width="1024px"
+                    header-title="${i18next.t("notifications.createNotification")}"
+                    ?opened="${this._createDialogOpen}"
+                    @opened-changed="${(ev: CustomEvent) => {
+                        if (!ev.detail.value && this._createDialogOpen) {
+                            this._createDialogOpen = false;
+                            this._loadData(); // reload after the dialog closes (cancel or create)
+                        }
+                    }}"
```
MartinaeyNL, line 481: I don't know what Remy his take on this is, but I would use `noCloseOnOutsideClick` to prevent closing this,
and maybe `noCloseOnEsc` as well. (HTML attribute name might be different) It prevents users closing the dialog by accident, and losing changes in the process.

ui/app/manager/src/components/notifications/notification-form.ts
```diff
@@ -0,0 +1,667 @@
...
+
+    constructor() {
+        super(NotificationFormChangedEvent.NAME, {
+            bubbles: true,
+            composed: true
+        });
+    }
+}
+
+@customElement("notification-form")
```
MartinaeyNL: I'd rename this to `or-notifications-form` for consistency.

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,670 @@
...
+
+            showSnackbar(undefined, i18next.t("notifications.successfullySentNotification"));
+            this._createDialogOpen = false;
+            // Reload explicitly: setting _createDialogOpen=false above means the @opened-changed handler's
+            // guard is already false, so it won't reload for us. Without this the new notification only
+            // appears after the next refetch (e.g. a filter change).
+            await this._loadData();
+        } catch (_error) {
+            showSnackbar(undefined, i18next.t("notifications.failedToCreateNotification"));
+            await this._loadData();
```
MartinaeyNL: Why is this necessary? I noticed, when retrieving an `HTTP 400` it still updates the list of notifications and the count. (so, two additional requests)

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,670 @@
...
+                ${!manager.authenticated
+                        ? html`
+                            <or-translate value="notAuthenticated"/>`
+                        : html`
+                            ${this._renderHeader(writeNotifications, hasRecipientType)}
+                            <div id="table-container">
+                                ${this._renderNotificationsTable()}
+                            </div>
+                        `
+                }
```
MartinaeyNL: No need to use an `manager.authenticated` wrapper, we don't do this anywhere.

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,670 @@
...
+            ${this._renderCreateDialog()}
+            ${this._renderDetailsDialog()}
+        `;
+    }
+
+    protected _renderCreateDialog() {
+        return html`
+            <or-vaadin-dialog
+                    id="createDialog"
+                    width="1024px"
```
MartinaeyNL, line 469: Any reason to hardcode 1024px, instead of "growing with the controls" ?
I understand it has to be large in width, but this would cause trouble on tablets for example.

ui/app/manager/src/pages/page-notifications.ts
```diff
@@ -0,0 +1,670 @@
...
+            </notification-form>
+        `;
+    }
+
+    protected _renderDetailsForm(notification?: SentNotification) {
+        return html`
+            <notification-form
+                    .realm=${this.realm}
+                    .notification=${notification}
+                    ?readonly=${true}>
```
MartinaeyNL: Replace this with `readonly` instead of `?readonly=${true}`

model/src/main/java/org/openremote/model/notification/NotificationResource.java
MartinaeyNL, line 1: I get really weird behavior when I logged into `smartcity` once, creating a console asset, and using that for notifications. Although I get a 400 from the server, it is still created in the database? (refreshing still shows it)
Or is this error handling in the frontend?

Image

MartinaeyNL (commented):
I'm having plenty of issues while testing the UI, so I wrote those down.
Along with that, I got some code review comments which you can see below.
Definitely worth a second look, and a bit of cleanup.

ui/app/manager/src/pages/page-notifications.ts
Ekhorn: It does shrink for smaller screens, it just makes sure on larger screens it doesn't extend past 1024px.

model/src/main/java/org/openremote/model/notification/NotificationResource.java
Ekhorn: Let's discuss this, because I also see your local branch is probably not up-to-date. I see that the status of those provisioned notifications is "pending", this changes to "sent".

ui/app/manager/src/pages/page-notifications.ts
Ekhorn: You seem to have tested on older changes. I can't reproduce your errors when I fill in the same things, and unless your logged in with the admin user you shouldn't see the realm recipient option.

Ekhorn: Also see https://github.com/openremote/openremote/pull/2673#discussion_r3636422174

Ekhorn: Good find!

ui/app/manager/src/components/notifications/or-notification-form.ts
Ekhorn: Any asset can be linked to a user right? It's just that console assets are linked to a user by default.

ui/app/manager/src/components/notifications/notification-form.ts
Ekhorn: Discussed made it "user".

Ekhorn:
@MartinaeyNL currently re-deploying https://github.com/openremote/openremote/actions/runs/30015918896

ui/app/manager/src/components/notifications/or-notification-form.ts
```diff
@@ -0,0 +1,688 @@
...
+    }
+
+    private async _populateFromNotification() {
+        if (!this.notification) return;
+
+        this._message = (this.notification.message || {type: "push"}) as NotificationMessage;
+        this._targetType = this.notification.target || NotificationTargetType.ASSET;
+        this._targets = this.notification.targetId ? [this.notification.targetId] : [];
+
+        await this.requestUpdate();
```
MartinaeyNL: Not sure whether this `await` works. Normally you'd wait for `this.updateComplete` to make sure everything is OK. Please double check.

ui/app/manager/src/components/notifications/or-notification-form.ts
MartinaeyNL: Oh yes of course. I wasn't sure whether this was handled correctly, but I think you've covered that in tests?

MartinaeyNL (commented):
Looks great!
Only a few minor comments to either address or resolve; up to you 👍 
Let me know when a final review is needed 😉

MartinaeyNL (approved):
🥳

PR #2674, opened 2026-04-07
Expose `queryDatapoints` in the `AbstractDatapointService` for ease of use

pankalog:
I need to easily retrieve the predicted datapoints I have previously set in the Protocol of an Agent. 

This PR is for exposing that API in particular: 
```java
org.openremote.manager.datapoint.AbstractDatapointService#queryDatapoints(java.lang.String, java.lang.String, org.openremote.model.datapoint.query.AssetDatapointQuery)
```


richturner (commented):
Don't see the implementation of this new interface method?

pankalog:
> Don't see the implementation of this new interface method?

@richturner there is no new implementation of the method, we're just exposing the `queryDatapoints` function in `AbstractDatapointService.java:180` to `ProtocolPredictedDatapointService`.

richturner:
> > Don't see the implementation of this new interface method?
> 
> @richturner there is no new implementation of the method, we're just exposing the `queryDatapoints` function in `AbstractDatapointService.java:180` to `ProtocolPredictedDatapointService`.

I think we need to look at how we expose services to protocols as unlike the `ProtocolAssetService` this is not realm scoped; will create a separate issue for this.

richturner:
See https://github.com/openremote/openremote/issues/2675

richturner (approved):

PR #2678, opened 2026-04-08
RuleService: preserve AttributeInfo ordering

### LINKED ISSUE #2676: AttributeInfo get pushed out of order in RulesEngine

ebariaux:


kubernetes/manager/templates/_helpers.tpl
wborn, line 1: I think all these Kubernetes changes got inadvertently added to this PR?

ebariaux: Thanks, fixed

wborn:
@codex review

manager/src/main/java/org/openremote/manager/rules/RulesService.java
```diff
@@ -724,11 +737,24 @@ protected void insertOrUpdateAttributeInfo(AttributeEvent attributeEvent) {
             return;
         }
 
-        LOG.log(FINEST, () -> "Inserting attribute event: " + attributeEvent);
+        final boolean inserted;
+        synchronized (attributeEventsLock) {
```
wborn: Do you have an idea what will be the impact on performance with all the added synchronization? Is it by design to not use `attributeEventsLock` for synchronizing all operations on `attributeEvents` and `attributeEventsByRef`? E.g. `deployGlobalRuleset(..)` and `deployRealmRuleset(..)` don't use it. As well as any call to `getAssetStatesInScope(..)`.

manager/src/main/java/org/openremote/manager/rules/RulesService.java
```diff
@@ -724,11 +737,24 @@ protected void insertOrUpdateAttributeInfo(AttributeEvent attributeEvent) {
...
+            AttributeEvent currentEvent = attributeEventsByRef.get(attributeEvent.getRef());
+            if (isOlderThanCurrent(attributeEvent, currentEvent)) {
+                AttributeEvent finalCurrentEvent = currentEvent;
+                LOG.log(FINEST, () -> "Ignoring attribute event older than current rule state: incoming=" + attributeEvent + ", current=" + finalCurrentEvent);
```
wborn: The `finalCurrentEvent` variable is redundant:

```suggestion
                LOG.log(FINEST, () -> "Ignoring attribute event older than current rule state: incoming=" + attributeEvent + ", current=" + currentEvent);
```

manager/src/main/java/org/openremote/manager/rules/RulesService.java
```diff
@@ -761,6 +792,37 @@ protected List<AttributeInfo> getAssetStatesInScope(String assetId) {
             .collect(Collectors.toList());
     }
 
+    protected void bufferPreInitAttributeEvent(AttributeEvent attributeEvent) {
+        synchronized (preInitAttributeEvents) {
+            AttributeEvent currentEvent = preInitAttributeEvents.get(attributeEvent.getRef());
+            if (!shouldReplaceBufferedPreInitEvent(currentEvent, attributeEvent)) {
+                AttributeEvent finalCurrentEvent = currentEvent;
+                LOG.log(FINEST, () -> "Ignoring buffered pre-init attribute event older than current buffered state: incoming=" + attributeEvent + ", current=" + finalCurrentEvent);
```
wborn: The `finalCurrentEvent` variable is redundant:

```suggestion
                LOG.log(FINEST, () -> "Ignoring buffered pre-init attribute event older than current buffered state: incoming=" + attributeEvent + ", current=" + currentEvent);
```

test/src/test/groovy/org/openremote/test/rules/RulesServiceStateOrderingTest.groovy
```diff
@@ -0,0 +1,258 @@
+package org.openremote.test.rules
```
wborn, line 20:
```suggestion
/*
 * Copyright 2026, OpenRemote Inc.
 *
 * See the CONTRIBUTORS.txt file in the distribution for a
 * full listing of individual contributors.
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Affero General Public License as
 * published by the Free Software Foundation, either version 3 of the
 * License, or (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 * GNU Affero General Public License for more details.
 *
 * You should have received a copy of the GNU Affero General Public License
 * along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
package org.openremote.test.rules
```

manager/src/main/java/org/openremote/manager/rules/RulesService.java
ebariaux: I don't have an idea on the perf impact.
For deployGlobalRuleset, deployRealmRuleset and getAssetStatesInScope, it's only using attributeEvents, that's a ConcurrentHashMap, so that should not be a problem.
The lock is required in the other places to maintain attributeEvents and attributeEventsByRef in sync.

But taking this into account, I now refactored the code to only use attributeEventsByRef so the lock can be removed.
There's now a copy of the values being taken in the above referenced methods but those should be used less often so less of an issue.

wborn:
@codex review

manager/src/main/java/org/openremote/manager/rules/RulesService.java
```diff
@@ -755,12 +773,46 @@ protected void retractAttributeInfo(AttributeEvent attributeEvent) {
...
     protected List<AttributeInfo> getAssetStatesInScope(String assetId) {
-        return attributeEvents
+        return getAttributeEvents()
             .stream()
             .filter(assetState -> Arrays.asList(assetState.getPath()).contains(assetId))
             .collect(Collectors.toList());
     }
 
+    protected Collection<AttributeEvent> getAttributeEvents() {
+        return new ArrayList<>(attributeEventsByRef.values());
```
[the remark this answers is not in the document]
wborn: It probably only makes sense to see how much of an issue it really is after some profiling on a bigger instance.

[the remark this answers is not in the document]
ebariaux: Applied the proposed fix, keeping the getter to keep tests calling this API stable.

wborn (approved):
Nice fix to make tests and production more stable! :+1:

PR #2703, opened 2026-04-21
Fix: skip kc token refresh for unauthenticated (public) users

Hackerberg43:
## Description
On public dashboards, users were redirected to keycloak after a while. This was caused by the reconnect function in UI core, trying to refresh the kc token ,even for unauthenticated users, who do not have one.
This PR makes the reconnect function in ui core skip the keycloak token refresh for unauthenticated users and reconnects to the event provider directly.

## Checklist


richturner:
@MartinaeyNL I have assigned you to this as you were looking at auth/websocket issues recently

MartinaeyNL (commented):
I ran some manual tests locally, figuring out whether there are edge cases where `_authenticated = false`.
Because that would break the whole reconnection logic. But I couldn't find any 😉 
Other tests I did, in regards to "offline / reconnect behavior", also don't cause any problems.
So I'm happy with the changes here 👍 

Quick question on the end-to-end tests I saw in the commit history;
I assume they were removed on purpose? Because I would be really interested if we can find a way to simulate Manager UI disconnects, and prevent bugs like these in the future.

Hackerberg43:
Hey Martin, thanks for taking a look!
Well I tried some things but couldn't directly replicate the correct disconnect behavior in playwright. And it has quite a large impact on test time because you'll have to wait for several timeouts to pass. Would be happy to look into it again if there are more use cases to justify the time increase

Hackerberg43:
Just waiting for your opinion about the relevant test :)

MartinaeyNL:
Oh yeah, sorry for not responding. 😅 
I'm happy to merge this, and to create a separate issue for picking up the "complex E2E tests" later.
They're a larger topic to address anyway across all of the frontend UI.

Thanks!

MartinaeyNL (approved):

MartinaeyNL:
See #2712 for the follow-up issue on the end to end tests 👍

MartinaeyNL:
This PR introduced an issue, which I'm fixing in #2729.

PR #2721, opened 2026-04-30
Fix stale attribute panel

### LINKED ISSUE #2609: Insight dashboard fail to render if used asset attribute has been deleted
With manager v1.19.0

I had a thing asset with 2 attributes, with predicted data points.
Created a dashboard with a line chart based on those 2 attributes.
I go back to the asset and delete one of the attribute.

Go back to the dashboard and the line chart stays blank.
Exception in the web console
```
bundle.3b7ac65872d99e96.js:9689 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'name')
    at bundle.3b7ac65872d99e96.js:9689:736681
    at Array.map (<anonymous>)
    at n.value (bundle.3b7ac65872d99e96.js:9689:736444)
    at n.value (bundle.3b7ac65872d99e96.js:5861:245517)
    at n.value (bundle.3b7ac65872d99e96.js:5861:239623)
    at n.value (bundle.3b7ac65872d99e96.js:5861:238685)
    at n.<anonymous> (bundle.3b7ac65872d99e96.js:5861:238506)
    at u (bundle.3b7ac65872d99e96.js:5861:234129)
    at Object.<anonymous> (bundle.3b7ac65872d99e96.js:5861:233416)
    at S (bundle.3b7ac65872d99e96.js:5861:229758)
```

Editing the dashboard, there's only the remaining attribute linked to the line chart and it's not possible to remove the one that does not exist anymore.

Logout/login did not change anything.

### LINKED ISSUE #2719: Dashboard settings attribute-panel unable to deal with stale attributes
**Describe the bug**
You can break a widget by invalidating its attribute config. An invalid config can be created by linking an attribute with the attribute-panel, and then removing that attribute/asset. The result is an empty list with only the option to change the attribute. However, it's not directly clear that the there is still a config active, because of the empty list

Image

Image

**To Reproduce**
1. Create an attribute widget
2. Link it to a custom attribute
3. Delete the custom attribute on the asset
4. Result: widget will not render, config is broken but not fully clear why.

**Expected behavior**
The list should report "broken reference" instead of showing an empty list, also adding back the delete button.
Or, the config should be cleaned, like it seems to be the case for line chart widget with multiple attributes.

Hackerberg43:
Fixes https://github.com/openremote/openremote/issues/2719 ,  fixes https://github.com/openremote/openremote/issues/2609

## Description
Makes attribute-panel in or-dashboard-builder show configured attributerefs that do not exist anymore as broken reference and enable the possibility to delete them.

afbeelding

## Checklist


ui/component/or-dashboard-builder/src/panels/attributes-panel.ts
```diff
@@ -296,9 +318,23 @@ export class AttributesPanel extends LitElement {
...
                                 }
+                                // Reference points to an attribute that no longer resolves.
+                                if (!this._assetsLoading && attributeRef) {
+                                    return html`
+                                        <div class="attribute-list-item broken">
+                                            <div class="attribute-list-item-icon">
+                                                <or-icon icon="link-variant-off"></or-icon>
+                                            </div>
+                                            <div class="attribute-list-item-label">
+                                                <span class="broken-label">${i18next.t("brokenReference")}</span>
```
MartinaeyNL: I think you can simply replace this with `<or-translate>`

```suggestion
                                                <or-translate class="broken-label" value="brokenReference"></or-translate>
```

ui/component/or-dashboard-builder/src/panels/attributes-panel.ts
MartinaeyNL, line 1: While I was testing out locally, I noticed that the colors can be inconsistent once assets/attributes are removed.
I did the following:
- Open the smartcity demo
- Delete the Erasmusbrug Parking asset
- Open the default Smartcity Parking Dashboard that is already present
- Select the chart widget in edit mode, you'll notice the colors in the settings are different from the ones shown.

This is probably because the default colors (1 Blue, 2 Orange, and 3 Green) are assigned to the 3 attributes of the widget. The Erasmusbrug asset, which was Blue, is still present in the `widgetConfig`, and therefore not visible.
I guess it would be solved by moving the "unknown assets" to the end of the list in the array.
Could you take a look at this?

This was already an issue before this PR, but I think it's an easy one to solve.
If not, feel free to let me know.

MartinaeyNL (commented):
Thanks for opening the PR!

Code looks OK to me! Please take a look at the two comments I have written down.
Let me know what you think, and if you need additional context. 👍

ui/component/or-dashboard-builder/src/panels/attributes-panel.ts
Hackerberg43: Sure, will take a look :)

Hackerberg43: Took the opportunity to make this even more clear: attributeRefs not resolving will not be filtered out in the chart/barcharts, so now you can even see from the insights viewer that something is wrong with the chart. Consequentially also solves your issue :) 
image

Hackerberg43: Following thought came to mind as additional action on broken reference: allowing a user to swap a broken reference for another attribute. This makes templating dashboards a real option? Load in a dashboard config from another instance/realm, switch the attributes and the chart layouts would stay intact. Would this actually be used?

MartinaeyNL:
Oh, sorry for not picking up this PR. It got lost on my part 😅 
I'll take one more look at this on Monday.

MartinaeyNL (commented):
I'm happy with these changes!

Last thing is to add basic coverage with a few simple UI tests, preventing the bug to reoccur.
Can you take a look at that?

Hackerberg43:
So one more thing I added is a replace button, so you wont lose the attribute specific settings when replacing the broken reference. Also added it to non broken attributes now as that was literally a few lines. 
image

This PR depends on https://github.com/openremote/openremote/pull/2725 now, that needs to be merged first. Will implement tests once that one is merged.

Hackerberg43:
@MartinaeyNL Tests are in

ui/app/shared/locales/nl/or.json
```diff
@@ -926,7 +926,9 @@
     "fill": "Vullen",
     "extendData": "Verleng laatste datapunt",
     "lineColor": "Lijnkleur",
-    "barColor": "Kleur van de balk"
+    "barColor": "Kleur van de balk",
+    "replaceAttribute": "Vervang atribuut",
+    "attributeAlreadyAdded": "Dit atribuut is al toegevoegd."
```
MartinaeyNL, line 932: You've added `brokenReference` as a key to the English translations, but not to the Dutch one.

ui/app/manager/test/insights.test.ts
Ekhorn, line 1: For the sake of clarity it would be good to also resize the widgets in the new test so that you can better see what happens 😉 .

```ts
await insightsPage.resizeWidgetTo(chartWidget, [<number>, <number>]);
```

Image

MartinaeyNL (commented):
Happy with these changes! Works great!
I've only one small note on the translations. I think it's ready to merge after that 😉

ui/app/manager/test/insights.test.ts
Hackerberg43: added. I did need to alter the existing fixtures however to make this work, but al the tests pass now. have a look

ui/app/shared/locales/nl/or.json
Hackerberg43: fixed

ui/app/manager/test/insights.test.ts
Ekhorn: @Hackerberg43 made a few changes in https://github.com/Hackerberg43/openremote/pull/11, will approve after it's been addressed.

Ekhorn (approved):

Ekhorn:
Thanks!

PR #2733, opened 2026-05-06
Fix RULE_RESET_IMMEDIATE not working as expected

### LINKED ISSUE #2066: `RULE_RESET_IMMEDIATE` meta item doesn't work as expected in when then rule
Copied from: https://forum.openremote.io/t/use-of-reset-rule-inmediate-meta-item/3712

`RULE_RESET_IMMEDIATE` meta item on an attribute doesn't cause when then rule to re-trigger on next attribute update that still matches the condition.

**To Reproduce**
We have defined an asset called "Sensor de Suelo" (Soil Sensor) with several attributes, including the "SoilHum" (Humedad) attribute. This attribute has the following meta items:

Image

This sensor reads the soil attributes every 60 minutes.

We have also created a rule based on this attribute, if the reading is below 30% it should send an email to a user called "administrador" (in the future we will make it so that with the activation of this rule we can turn on the irrigation system in the area).

Image

The rule is effectively activated when the reading is below 30% but, if the next reading is below 30% again it is not activated. We know that this is by design but within the meta items we find "Rule reset immediate" which we understand is used when the rule state is required to be reset with the change of the value in the attribute. We added this meta item but the behavior remains the same.

Is our understanding of the use of the metaitem "Rule reset immediate" correct?
If so, why is it not working for us?
If not, what is its use? Is there a way to make the rule trigger with every reading below 30% even if they are continuous?

**Expected behavior**
The rule should fire on each attribute update that still matches the when condition of the rule.

ebariaux:
# What was wrong

  JSON rule conditions track previouslyMatchedAssetStates to avoid firing repeatedly while an attribute remains matched. AttributeEvent.equals(...) only compares asset id and attribute name, not value or timestamp, so a newer event for the same attribute is considered the same set
  entry.

  Before the fix, every update did this:

  previouslyMatchedAssetStates.remove(event.assetState);
  previouslyMatchedAssetStates.add(event.assetState);

  That refreshed the previous-match set with the new timestamp before rule evaluation. Later, the RULE_RESET_IMMEDIATE logic tried to detect a newer matching event:

  resetImmediately && matchedAssetState.getTimestamp() > previousAssetState.getTimestamp()

  But both sides now referred to the new timestamp, so the comparison became effectively:

  newTimestamp > newTimestamp

  That is false, so the previous match was not cleared, and the rule did not fire again.

# What the fix does

  For normal attributes, behavior stays the same: the previous matched state is refreshed on update, so a rule does not retrigger just because an already-matching value is updated.

  For attributes with RULE_RESET_IMMEDIATE, we do not eagerly replace the previous matched state. That leaves the old timestamp in previouslyMatchedAssetStates until evaluation. Then the existing reset logic can correctly compare:

  newTimestamp > oldTimestamp

  When true, the previous match is removed. After that, the current matching state is no longer filtered out as “already matched,” so the rule becomes eligible to fire again.

### DIFF
--- a/manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
+++ b/manager/src/main/java/org/openremote/manager/rules/JsonRulesBuilder.java
@@ -223,7 +223,8 @@ public class JsonRulesBuilder extends RulesBuilder {
                 // only need up to date values in the previously matched asset states previously unmatched asset states is only
                 // used to compare asset ID and attribute name.
                 if (event != null) {
-                    if (previouslyMatchedAssetStates.remove(event.assetState)) {
+                    boolean resetImmediately = event.assetState.getMeta().getValue(MetaItemType.RULE_RESET_IMMEDIATE).orElse(false);
+                    if (!resetImmediately && previouslyMatchedAssetStates.remove(event.assetState)) {
                         previouslyMatchedAssetStates.add(event.assetState);
                     }
                 }
--- a/test/src/test/groovy/org/openremote/test/rules/RuleResetImmediate.groovy
+++ b/test/src/test/groovy/org/openremote/test/rules/RuleResetImmediate.groovy
@@ -1,0 +1,134 @@
+package org.openremote.test.rules
+
+import org.openremote.manager.notification.EmailNotificationHandler
+import org.openremote.manager.notification.NotificationService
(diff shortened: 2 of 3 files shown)


richturner (approved):
Nice find for this issue and good clear test for it

PR #2734, opened 2026-05-06
Add helper methods to get latitude and longitude from a GeoJSONPoint

ebariaux:


model/src/main/java/org/openremote/model/geo/GeoJSONPoint.java
Ekhorn, line 1: There are already getters (`getX` and `getY`) for these properties.

ebariaux: I know, I should have put more context, I added a comment in the code.
The getX() and getY() are not intuitive and actually lead me to create a bug in MyGrid because methods working with coordinates took latitude, longitude as param and I wrote loc.getX(), loc.getY() as params because it was the "logical" order.
I did not realise that latitude is actually Y (until I saw the bug and had to think about it).

Ekhorn: Clear.

Ekhorn (approved):

PR #2744, opened 2026-05-11
Add nixos dependencies through a flake to run the project

o-x-e-y:
adds a `flake.nix`, `.envrc` to load the flake automatically through direnv and adds `.direnv` to the gitignore.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,37 @@
+{
+  description = "A Bun devshell";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs =
+    {
+      self,
+      nixpkgs,
+      flake-utils,
+      ...
+    }:
+    flake-utils.lib.eachDefaultSystem (
+      system:
+      let
+        pkgs = import nixpkgs {
+          inherit system;
+        };
+      in
+      {
+        devShells.default =
+          with pkgs;
+          mkShell {
+            buildInputs = [
+              nodejs
+              docker
+              yarn
+              jdk
+              pkg-config
+            ];
+          };
+      }
+    );
+}
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ deployment/manager/fcm.json
 .idea/
 .sts4-cache/
 .local/
+.direnv/
 .DS_*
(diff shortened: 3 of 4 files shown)


Ekhorn:
Will discuss with the team (I would personally advocate for Nix integration).

flake.nix
Ekhorn, line 1: Please consider the following (instead of introducing flake-utils)

```nix
      # Using nixpkgs.lib.systems.flakeExposed systems,
      # see https://github.com/NixOS/nixpkgs/blob/5d65a618c663db71662a434a3d5887f2ee7f0a1f/lib/systems/flake-systems.nix
      forAllSystems = f: builtins.mapAttrs f nixpkgs.legacyPackages;
```

flake.nix
```diff
@@ -0,0 +1,37 @@
...
+      in
+      {
+        devShells.default =
+          with pkgs;
+          mkShell {
+            buildInputs = [
+              nodejs
+              docker
+              yarn
+              jdk
```
Ekhorn, line 31:
```suggestion
              jdk21_headless
```

flake.nix
```diff
@@ -0,0 +1,37 @@
...
+              pkg-config
```
Ekhorn, line 32: I don't use this myself, will check whether it is relevant to include.

```suggestion
```

o-x-e-y:
I made this by mistake, I'm part of the Fontys team and I meant to PR this to our own fork. But feel free to merge this either way obviously!

flake.nix
Ekhorn: See the following thread: https://discourse.nixos.org/t/cross-system-flakes-whats-your-favorite-forallsystems/68924/8

Ekhorn: You can then use it like so:

```nix
      devShells = forAllSystems (
        system: pkgs: {
          default = mkShell {
            buildInputs = with pkgs; [];
          };
        }
      );
```

Ekhorn (commented):
Not sure at this point we'll be integrating this, but do what you want with the review 😉

flake.nix
Ekhorn: Generally specifying the specific versions we document on (https://docs.openremote.io/docs/developer-guide/preparing-the-environment/) to be compatible with OpenRemote should be installed.

Ekhorn:
We decided to not merge this as we are not sure at this time it will be popular enough to justify maintaining this and would like stay away from tooling that isn't fully cross-platform (without WSL). I may create a dedicated flake for OpenRemote that you can use to install the required dependencies in a separate repository and we may point to it in our documentation, so you can install the required dependencies with a single command.

PR #2773, opened 2026-05-18
Add Asset table fillfactor migration script and document postgres container memory settings

### LINKED ISSUE #2611: DB change asset table fillfactor to 80 to allow HOT updates
The asset table is constantly being updated (whenever an attribute event is written to the attributes JSONB column) this causes excessive autovacuum executions but is also a significant potential bottleneck for large asset tables due to the way row updates are handled by postgreSQL. Some info: 

> Heap-Only Tuples (HOT) updates.
> 
> Normally, when PostgreSQL inserts that new version of the row, it also has to update every single index on the table to point to the new physical location of the row. This is incredibly expensive (this is what causes "index bloat").
> 
> A HOT update bypasses the indexes entirely. It places the new version of the row on the exact same disk page as the old version, and just links them together. It is vastly faster and prevents index bloat.
> 
> For a HOT update to happen, two rules must be met:
> 
> You cannot be updating a column that has an index on it. (If you have a GIN index on that entire jsonb column, HOT updates are disabled, and you will suffer massive bloat).
> 
> There must be empty space on the current disk page to fit the new row.

To facilitate this we have to reduce the `fillfactor` on the asset table and ensure no indexes are in place on that column (which they aren't), choosing the fillfactor:

> The Math Behind the Space
> PostgreSQL stores data in 8KB (8,192 bytes) pages.
> 
> A fillfactor of 80 leaves about 1,638 bytes of free space per page.
> 
> A fillfactor of 90 leaves about 819 bytes of free space per page.
> 
> For a HOT update to succeed, the entire new version of the row must fit into that free space on the exact same page.
> 
> If your average row in the asset table is relatively small (for example, 150 bytes), leaving 10% free space (819 bytes) gives PostgreSQL enough room to perform 5 consecutive HOT updates to a row before that page completely fills up and requires pruning.
> 
> How to Choose Between 90 and 80
> Go with fillfactor = 90 (10% free space) if:
> 
> Your JSONB updates are relatively small and the overall row size is small (under a few hundred bytes).
> 
> The updates are somewhat spread out, giving PostgreSQL's opportunistic page pruning a chance to clean up the dead HOT tuples between updates.
> 
> You want to minimize the storage and RAM penalties we talked about earlier.
> 
> Stick with fillfactor = 80 (20% free space) if:
> 
> Your rows are physically large (e.g., storing massive JSON documents where a single row takes up 1,000+ bytes). If a row is 1,000 bytes, 10% free space (819 bytes) literally isn't big enough to hold a single HOT update!
> 
> A single row gets battered with back-to-back-to-back updates in a matter of seconds, meaning you need room for multiple dead copies to pile up before pruning can catch up.

### The Golden Rule of HOT Updates
Remember, free space is only half of the equation. Even if you set the fillfactor to 50 and leave half the page empty, a HOT update will instantly fail and revert to a standard, bloat-causing update if you modify a column that has an index on it. As long as your frequent JSONB updates are not touching indexed keys, dropping the fillfactor to 90 is a fantastic, highly efficient starting point.

The fillfactor may need to be tweaked on an instance basis depending on the number of attributes in each asset and this should be covered in the docs.

How to apply this change is very simple and should be added as a one off migration script:

```sql
ALTER TABLE asset SET (fillfactor = 80);
```

### LINKED ISSUE #2612: DB ensure maintenance_work_mem aligns with postgres container shm_size
By default docker allocates 64MB as the shared memory limit which is tiny and postgres will set a much higher value for `maintenance_work_mem` this can result in errors like the following when vacuum operations run:

```
ERROR:  could not resize shared memory segment "/PostgreSQL.1846622834" to 348151776 bytes: No space left on device
```

Make sure the default postgres shared memory limit is more appropriate for postgres and that `maintenance_work_mem` is set to a sensible default

richturner:
Closes #2611, Closes #2612


manager/src/main/resources/org/openremote/manager/setup/database/V20260518_01__AssetFillFactor.sql
```diff
@@ -0,0 +1,2 @@
+-- SET ASSET FILL FACTOR TO 90%
+ALTER TABLE ${schemaName}.asset SET (fillfactor = 90);
```
denniskuijs, line 2: The issue (#2611) describes setting the fillfactor to 80%, but it’s currently configured to 90%. Is this intended?

richturner: 90% should be sufficient and avoids too much disk wastage, will update issue

denniskuijs (approved):

Issue #2684, opened 2026-04-14
Consider creating clear distinctions between integration and unit tests


Ekhorn:
## User Story
As a developer I want to have unit- and integration tests clearly separated, so that it's easier to divide when and how their run (e.g. for parallelism/sharding in the pipeline).

## Description

I've attempted to use the [JVM test suite plugin](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#jvm_test_suite_plugin) to make clear distinctions between our unit- and integration tests. Because we have a mix of unit- and integration tests inside the `:test` project.

Besides the backend test distinction we may also be able to make a clearer distinction for the ui app/component tests.

However this has major structural impact on where the test files are stored, how the test dependencies are prefixed and the defined Gradle tasks. We can't just make a change like this for only the main repo as this would impact the shared pipeline for custom projects, nor is it desirable to different structures.

## Definitions

- Unit test: Any backend test (JUnit / Spock) that doesn't implement a facade/fixture to integrate with an external service (e.g. `ManagerContainerTrait`) and doesn't require a dependency on multiple core projects which would result in a circular dependency chain.
- Integration test: A backend test under the `:test` project which implements a facade/fixture to integrate with an external service (e.g. `ManagerContainerTrait`) or if the test covers functionality from multiple core projects that could cause a circular dependency chain.

## Acceptance Criteria
- [ ] 1. Unit tests must be possible to execute separately to allow better control over parallelism i.e. they're found under `test/src/test/**` for each project.
- [ ] 2. Integration tests must be identified by the `test/src/integration/` path (and their dependencies are prefixed with `integrationTest` to be compliant with the `jvm_test_suite_plugin`)
- [ ] 3. Custom projects should be brought up-to-date with the new structure that the pipeline assumes, including test directories, dependencies and Gradle tasks.

## Implementation Notes

The [jvm_test_suite_plugin](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#jvm_test_suite_plugin) plugin integrates with the [test_report_aggregation_plugin](https://docs.gradle.org/current/userguide/test_report_aggregation_plugin.html).

ebariaux:
We standardize on using Spock for all tests and not JUnit.
Unit tests live in the `test` folder of the module they apply to.
Integration tests, that require a running stack (DB + keycloak), live under the `test` module.

Ekhorn:
[...]

PR #2774, opened 2026-05-19
Fix WebSocket unsubscribe without event type or subscriptionId

wborn:
## Description

The frontend usually cancels WebSocket subscriptions by sending only the subscriptionId, without the event type. This worked before the subscription handling was refactored in PR #1432. Afterwards those unsubscribe requests no longer matched the stored subscription keys.

Because subscriptions are not removed, every page switch between views such as Map and Assets creates additional subscriptions for the same session. This causes duplicate events to be sent to the frontend and becomes problematic on large installations.

Fall back to matching the subscriptionId suffix when no event type is provided, remove the matching consumer, and clean up the session entry when it has no remaining subscriptions.

## Checklist

### DIFF
--- a/manager/src/main/java/org/openremote/manager/event/ClientEventService.java
+++ b/manager/src/main/java/org/openremote/manager/event/ClientEventService.java
@@ -127,14 +127,25 @@ public class ClientEventService extends RouteBuilder implements ContainerService
     protected static final String WEBSOCKET_URI = "undertow://ws://0.0.0.0/websocket/events?fireWebSocketChannelEvents=true&sendTimeout=15000"; // Host is not used as existing undertow instance is utilised
     protected static final System.Logger LOG = System.getLogger(ClientEventService.class.getName());
     protected static final String PUBLISH_QUEUE = "direct://ClientPublishQueue";
+
+    /**
+     * Delimiter used when composing the internal subscription key from {@code eventType} and {@code subscriptionId}.
+     *
+     * <p>The key format is: {@code <eventType>::<subscriptionId>}
+     *
+     * <p>This allows unsubscribe requests to match subscriptions even when the
+     * client only provides one part of the original key, which is required for
+     * websocket clients that unsubscribe using only the subscription id or only the event type.
+     */
     protected static final String SUBSCRIPTION_KEY_DELIMITER = "::";
 
-    final protected Collection<EventSubscriptionAuthorizer> eventSubscriptionAuthorizers = new CopyOnWriteArraySet<>();
-    final protected Collection<EventAuthorizer> eventAuthorizers = new CopyOnWriteArraySet<>();
-    final protected Set<Pair<EventSubscription<? extends Event>, Consumer<? extends Event>>> eventSubscriptions = new CopyOnWriteArraySet<>();
-    final protected Map<String, WebSocketChannel> sessionChannels = new ConcurrentHashMap<>();
-    final protected Map<String, Map<String, Consumer<? extends Event>>> websocketSessionSubscriptionConsumers = new ConcurrentHashMap<>();
-    final protected Map<String, SessionDispatchState> sessionStates = new ConcurrentHashMap<>();
+    protected final Collection<EventSubscriptionAuthorizer> eventSubscriptionAuthorizers = new CopyOnWriteArraySet<>();
+    protected final Collection<EventAuthorizer> eventAuthorizers = new CopyOnWriteArraySet<>();
+    protected final Set<Pair<EventSubscription<? extends Event>, Consumer<? extends Event>>> eventSubscriptions = new CopyOnWriteArraySet<>();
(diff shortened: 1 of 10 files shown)


manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -485,8 +487,21 @@ void procesInboundEvent(Exchange exchange) {
...
          LOG.log(TRACE, () -> "Cancelling subscription for session '" + sessionKey + "': " + cancelEventSubscription);
          websocketSessionSubscriptionConsumers.computeIfPresent(sessionKey, (s, subscriptionConsumers) -> {
-            String subscriptionKey = cancelEventSubscription.getEventType() + cancelEventSubscription.getSubscriptionId();
-            Consumer<? extends Event> consumer = subscriptionConsumers.remove(subscriptionKey);
+            Consumer<? extends Event> consumer = null;
+            if (!cancelEventSubscription.getEventType().isEmpty()) {
+               String subscriptionKey = cancelEventSubscription.getEventType() + cancelEventSubscription.getSubscriptionId();
+               consumer = subscriptionConsumers.remove(subscriptionKey);
+            } else if (!cancelEventSubscription.getSubscriptionId().isEmpty()) {
+               Pattern pattern = Pattern.compile("(\\d+)$");
```
richturner: This could be statically initialised depending how often we end up here?

richturner: Also wouldn't `String.endsWith(cancelEventSubscription.getSubscriptionId())` be more efficient here

wborn: I extracted it into a constant. I thought about using `endsWith` but it is less deterministic e.g. when the frontend unsubscribes from `1`  when there is for some reason an `asset1` and `asset11` subscription.

Ekhorn:
Manually tested this, was able to reproduce the issue and can see it is resolved with these changes.

This could be covered in the tests as there is currently no code coverage for the `CancelEventSubscription` branch according to the JaCoCo report.

richturner:
OK I have looked into this a bit more and based on the javadoc in `CancelEventSubscription` this doesn't fully fix the issue and also I think we should avoid doing regex matching for performance reasons; I propose the following steps:

1. As @Ekhorn highlighted; add integration tests for `CancelEventSubscription` logic  for cases where only eventType is provided, only subscriptionID is provided or both are provided
1. `ClientEventService:L470` Change subscriptionKey to include a delimiter (e.g. `:` -> `eventType::subscriptionId`)
1. Fail any subscription with a subscription ID prefixed with the delimiter and include this in EventSubscription javadoc cover this in test
1. Change the new logic to:
   1. Match whole subscription key if eventType & subscription ID provided
   1. startsWith match with delimiter if only eventType provided
   1. endsWith match with delimiter prefix if only subscription ID provided

manager/test/src/groovy/org/openremote/manager/event/ClientEventServiceSubscriptionTest.groovy
Ekhorn, line 1: Maybe you can move this test under a test directory in the manager project, as discussed in this issue https://github.com/openremote/openremote/issues/2684#issuecomment-4458007639.

test/build.gradle
```diff
@@ -28,6 +28,7 @@ dependencies {
     }
 
     testImplementation "com.google.firebase:firebase-admin:$firebaseAdminVersion"
+    testImplementation "org.apache.camel:camel-undertow:$camelVersion"
```
wborn:
```suggestion
```

manager/src/main/java/org/openremote/manager/event/ClientEventService.java
```diff
@@ -485,11 +513,27 @@ void procesInboundEvent(Exchange exchange) {
...
+               if (predicate != null) {
+                  subscriptionConsumers.keySet().stream().filter(predicate).forEach(subscriptionKeys::add);
+               }
             }
+            subscriptionKeys.forEach(subscriptionKey -> {
+               Consumer<? extends Event> consumer = subscriptionConsumers.remove(subscriptionKey);
+               if (consumer != null) {
+                  removeSubscription(consumer);
+               }
+            });
```
richturner: Don't get the need to filter -> subscriptionKeys -> remove why not a single predicate and then `subscriptionCnsumers.keySet().removeIf()`

wborn: How would you then call `removeSubscription(consumer)`?

richturner: Call it within the removeIf predicate before returning true

wborn: OK I rewrote it using all predicates and `removeIf`

richturner (approved):

PR #2787, opened 2026-05-21
Use ubi10 based OpenJDK base image and pin version to 1.24-5

### LINKED ISSUE #2784: Review base image used for manager Docker image
We're currently building our Dockerimage from ubi9/openjdk-21-runtime.
UBI10 images are available, that should be the first step.

When looking at all possible images, in the UBI family, there are 4 different flavours, 3 potentially interesting: ubi10, ubi10-minimal and ubi10-micro.
[[(Re)Introducing the Red Hat Universal Base Image](https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image)](https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image)

There are also pre-made images with OpenJDK (what we're using so far), so that would now be ubi10/openjdk-21-runtime:1.24-5
We should probably update to that for now, including the specific version number so builds are reproducible.

This version seems to be built on top of ubi-minimal. If we want to go smaller, we could build our own image starting from ubi-micro, but this has constraint.

Another, maybe preferred option, is to use [[Red Hat Hardened Images](https://www.redhat.com/en/products/hardened-images)](https://www.redhat.com/en/products/hardened-images) and specifically hi/openjdk:21.0.11
But this has important constraint, e.g. there's no shell or typical command that could help in debugging issues on running containers.
That should probably be considered a future option if required by our security level or by specific projects.

ebariaux:


Ekhorn:
Nice reduction (and the Grype scan results).

```sh
docker images
registry.access.redhat.com/ubi10/openjdk-21-runtime   1.24-5          8dbafd8739f5   8 days ago      329MB
registry.access.redhat.com/ubi9/openjdk-21-runtime    latest          8d09296c7dd4   10 days ago     396MB
```

Ekhorn (approved):

PR #2791, opened 2026-05-26
Dont fallback to OR_DEV_MODE for OR_SETUP_RUN_ON_RESTART

Ekhorn:
## Description

As discussed in the retrospective this behavior made sense for tests, but is a bad default for development flows that require some things to be setup for testing features from the manager UI.

However this will likely break pipelines for custom projects as I assume they rely on this same fallback when running the tests.

## Checklist

### DIFF
--- a/project.gradle
+++ b/project.gradle
@@ -133,13 +133,15 @@ if (project == rootProject) {
                 "Demo Setup"(Application) {
                     moduleName = getProject().idea.module.name + ".extension.demo-setup.main"
                     envs = [
-                            OR_SETUP_TYPE: "demo"
+                        OR_SETUP_RUN_ON_RESTART: "true",
+                        OR_SETUP_TYPE: "demo"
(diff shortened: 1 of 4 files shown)


project.gradle
Ekhorn, line 1: On the topic of changing the default profiles I think adding `OR_APP_DOCROOT=manager/build/install/manager/web` to all of them would be nice for those that don't want to run `npm run serve(NoModelBuild)` every time.

However I don't know if that would impact anything. From testing it shouldn't interfere with `npm run serve`.

Ekhorn:
I would like to see if I could somehow build the manager with these changes locally and then use those artifacts in the custom projects to see if the tests still pass.

Ekhorn:
Also I think it would make sense to have `OR_APP_DOCROOT=manager/build/install/manager/web` enabled by default in devmode.

Issue #1936, opened 2025-06-13
Sending push notification with external website link opens app


pierrekil:
## User Story
I want to send a push notification which links to an external website

## Description
When I add an external website link to a push notification it still opens the app, and not the external website in the browser.

## Acceptance Criteria
- [ ] 1. [Criteria]

## Implementation Notes
[Relevant technical notes]

MartinaeyNL:
Did a quick code analysis, and I think this might be bug on the consoles.
Haven't tested it myself, so I'll try to reproduce this behavior first.

MartinaeyNL:
Cannot reproduce this. Tested both on iOS and Android, with several cases.
I was using `https://openremote.io` as website URL.

### Test server with OpenRemote Android app

- "Open in Browser" with app unopened: **Works**
- "Open in Browser" with app in background: **Works**
- "Open in Browser" with app running: **Works**
- "Open in Browser" using OK button, with app unopened: **Works**
- "Open in Browser" using OK button, with app in background: **Works**
- "Open in Browser" using OK button, with app running: **Works**
 
### Test server with OpenRemote iOS app:

- "Open in Browser" with app unopened: **Works**
- "Open in Browser" with app in background: **Works**
- "Open in Browser" with app running: **Works**
- "Open in Browser" using OK button, with app unopened: **Works**
- "Open in Browser" using OK button, with app in background: **Works**
- "Open in Browser" using OK button, with app running: **Works**

MartinaeyNL:
After having checked this with @pierrekil, the issue was occurring on custom projects.
His case was to send a notification to users saying "Hey open this link", which should open in the browser.
He was using an a production deployment of the O***** custom project.

Having done a quick test, this seems to only occur on iOS, not on Android.
I will do further testing with different custom project, different use cases, and both on Android and iOS to validate.

MartinaeyNL:
Tested this on a development deployment of the K**** custom project / app as well.
Results in the same behavior on both on Android and iOS.
See here;

___

### dev.k******.com with their custom Android app:

- "Open in Browser" with app unopened: **Works**
- "Open in Browser" with app in background: **Works**
- "Open in Browser" with app running: **Works**

### dev.k******.com with their custom iOS app:

- "Open in Browser" with app unopened: **No notification (might be expected behavior)**
- "Open in Browser" with app in background: **Opens the app, not the browser**
- "Open in Browser" with app running: **Opens the app, not the browser**

[...]

Issue #1467, opened 2024-10-29
Notifications: Browser does not open when opening push notification


DonWillems:
**Reproduce:**
1. Create When-Then rule with push notification action. Set e.g. "openremote.io" as URL and enable "open in external browser"
2. Trigger the rule
3. Receive push notification
4. Open the push notification by clicking it or by pressing the action button
5. Nothing happens
6. Do this a few more times and on Android you get the warning "[App name] keeps closing"

**Expected:**
5. Website opens in default browser

This is an issue in both Android and iOS

Tested on Android 13

Miggets7:
I've noticed that when creating a rule, `openInBrowser` appears to be on, but the notification sent has it set to `false`. Only after turning the toggle off, saving, turning it back on and saving it, makes the notification to have `openInBrowser` set to `true`.

Seems like some UI bug

Also, it's important to have `http(s)://` in the url, otherwise the app can't determine what activity should be opened.

I could though, add some code that checks if `openInBrowser` is set to `true`, to then check if `https://` should be appended.

@DonWillems should I add that?

Miggets7:
I've checked the code in `or-rule-form-push-notification.ts` and it hasn't been modified for over 4 years, so not sure why it's broken now. Maybe something for @MartinaeyNL to check?

PR #2794, opened 2026-05-27
Set action on push notification confirm button

### LINKED ISSUE #2605: Push notification confirm button missing action — shows destructive button on iOS
## Bug Description

The confirm button in push notification rules only sets a `title` on the `PushNotificationButton`, but does not set an `action` (`PushNotificationAction`). This causes iOS to render the button as a destructive action, and the button does not execute any action when tapped.

## Location

**File:** `ui/component/or-rules/src/json-viewer/forms/or-rule-form-push-notification.ts` (lines 82–87, 136–142)

The `_onButtonTitleChange` handler creates a `PushNotificationButton` with only a `title`:

```ts
protected _onButtonTitleChange(ev: OrInputChangedEvent, key: number, message: PushNotificationMessage) {
    message.buttons = message.buttons || [];
    message.buttons[key] = {
        title: ev.detail.value
    } as PushNotificationButton;
    this.dispatchEvent(new OrRulesJsonRuleChangedEvent());
}
```

The `PushNotificationButton` model (`model/src/main/java/.../PushNotificationButton.java`) supports both a `title` and an `action` (`PushNotificationAction`), but the rule form never sets the `action` field on buttons.

## Expected Behavior

The confirm button (index 0) should have its `action` set — for example, using the notification's own `message.action` or a dedicated `confirmAction`. This way, when the user taps "Confirm" on iOS, the app executes the intended action (e.g., writing an attribute value) and renders the button as a normal (non-destructive) action.

## Actual Behavior

- The confirm button has no `action`, only a `title`
- On iOS, this results in the button being shown as a **destructive** button
- Tapping the button does **not** execute the notification's action

## Suggested Fix

In `_onButtonTitleChange` (or via a separate mechanism), the confirm button (index 0) should also have its `action` populated. For example, the notification message's `action` could be copied to the confirm button:

```ts
message.buttons[key] = {
    title: ev.detail.value,
    action: key === 0 ? message.action : undefined
} as PushNotificationButton;
```

Alternatively, a dedicated action field/input could be added for button actions in the rule form UI.

Miggets7:
## Summary
- Populates the `action` field on the confirm button (index 0) created by `_onButtonTitleChange` in `or-rule-form-push-notification.ts`, copying the notification's own `message.action`.
- On iOS the button is no longer rendered as a destructive action, and tapping it now executes the notification's action.
- Applies the fix exactly as suggested by the reporter; the rules UI is intentionally left unchanged per maintainer guidance on the issue.

## Test plan
- [x] Open a rule with a push notification action in the manager UI.
- [x] Set an action URL on the notification and a confirm button title.
- [x] Verify the persisted rule JSON contains both `title` and `action` on the confirm button, and only `title` on the decline button.
- [x] On an iOS device, trigger the rule and confirm the button renders as a normal (non-destructive) action and executes the notification's action when tapped.

Fixes #2605


Ekhorn (commented):
It works! Reproduced the issue and it is fixed. I did find another unrelated existing issue (I'll make a separate issue).

Also tried this on Android, but that returns "FCM token not found for ..." as error. Would like to test Android before approving.

Miggets7:
@Ekhorn I believe the Android app should be updated. The `google-services.json` needs to be updated. I'll create an issue for this.

Miggets7:
@Ekhorn sorry not true, the App Signing SHA1 wasn't added in Google API Console, it should already work now

Ekhorn (approved):
Great @Miggets7! It still seems to work like before on Android. I did find that clicking the action doesn't open the specified page on Android. I see there are 2 closed issues about this https://github.com/openremote/openremote/issues/1467 and https://github.com/openremote/openremote/issues/1936.

PR #2820, opened 2026-06-03
Align skip conditions with branch protection checks

Ekhorn:
## Description

There are still branch protection check issues when jobs are skipped, because jobs are skipped at the wrong level. They should be skipped not where they are called (reused), but inside the reusable workflow itself.

I had to add an aggregation job for the sharded UI App tests job, because matrix jobs are incompatible when a required status check is expected for each shard while being skipped.

## Checklist

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -36,62 +36,82 @@ jobs:
   ui-check-dependencies:
     name: Lint
     needs: detect-changes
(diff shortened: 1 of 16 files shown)


.github/workflows/review.yml
Ekhorn, line 1: The changes in this file could also be applied to the other environment workflows, but there is currently no reason for it other than consistency.

wborn (approved):

PR #2829, opened 2026-06-06
Await code

tejas-ae:
The code around line 627 in ui/app/manager/src/pages/page-alarms.ts looked like it might have an issue. The promise result gets used as if it were already resolved. this patch adds the missing await so the async path is actually sequenced.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


MartinaeyNL (commented):
As Copilot mentioned, the `until()` Lit directive is used in rendering. (see docs [here](https://lit.dev/docs/templates/directives/#until))
It accepts a `Promise<?>` object, and awaits rendering until the function is finished.
So the `await` keyword is unnecessary here.

MartinaeyNL:
Closing this PR. Feel free to reopen whenever you want 👍

PR #2835, opened 2026-06-09
Upgrade to gradle 9.5.1

ebariaux:


Ekhorn (commented):
When I run the following `./gradlew wrapper --gradle-version 9.5.1` more files end up changed than in this PR. Not sure it is ok, to just modify the version number only here.

```
git status
        modified:   gradle/wrapper/gradle-wrapper.jar
        modified:   gradle/wrapper/gradle-wrapper.properties
        modified:   gradlew
        modified:   gradlew.bat
```

In this [commit](https://github.com/openremote/openremote/commit/7be7db21c22284b4ce2a30fad012309b1a523e8f) you also changed a line that points to a commit hash, which also happens when you run the `./gradlew wrapper` command.

Ekhorn (approved):

PR #2868, opened 2026-06-12
Entrypoint changes affecting heap dump storage and `OR_JAVA_OPTS`

richturner:
Store heap dumps in /storage dir and add support for multiple and warn on heap dump presence 

This allows heap dumps to persist between manager container recreation for analysis purposes.

The entrypoint script also outputs a warning message if heap dumps are found to encourage removal.

Replace `OR_JAVA_OPTS` with more common options `JAVA_OPTS` and `JAVA_OPTS_APPEND`


richturner:
@wborn Can you trigger another review please?

richturner:
@openremote/core-developers  Ready for review

balena/heapdump-rename.sh
ebariaux, line 1: A bit of a shame that we duplicate the file from manager folder but not sure the extra work to handle a centralized file is worth it.
We just need to be careful to not have the files diverge for the wrong reasons

ui/test/manager.cjs
```diff
@@ -26,16 +26,17 @@ console.log("Starting OpenRemote Manager...");
...
-    "-Xms500m",
-    "-Xmx2g",
+// We split JAVA_OPTS by space if it exists, or use the default JVM options defined in our Dockerfile
+const javaOpts = process.env.JAVA_OPTS ? process.env.JAVA_OPTS.split(" ") : [
+    "-XX:InitialRAMPercentage=75.0",
+    "-XX:MaxRAMPercentage=75.0",
     "-XX:NativeMemoryTracking=summary",
     "-Xlog:all=warning:stdout:uptime,level,tags",
     "-XX:+HeapDumpOnOutOfMemoryError",
     "-XX:HeapDumpPath=./dump.hprof",
```
ebariaux, line 36: This is only for ephemeral containers while running UI tests but do we want to align with what's done for elsewhere and use /storage as location

balena/heapdump-rename.sh
richturner: Agreed I didn't want to duplicate but the way docker build uses context path makes it tricky, should have this balena build within the manager module I think

ui/test/manager.cjs
richturner: I wasn't sure how/when this was being used and I saw a relative path so thought to leave as it is

wborn (commented):
We probably also needdocumentation updates for this as there are some pages about using JAVA_OPTS and heap dumps:

* https://docs.openremote.io/search?q=JAVA_OPTS
* https://docs.openremote.io/search?q=heapdump

richturner:
Doc PR also created openremote/documentation#162

richturner:
Heap dump references don't need replacing as they relate to ad-hoc heap dump generation

ebariaux (approved):

PR #2869, opened 2026-06-12
Improved asset datapoints export access control

ebariaux:


MartinaeyNL (approved):

MartinaeyNL:
A reminder for the user adding this PR to the next release; Add a very simple note on this "breaking change in the HTTP API".
Because maybe people were abusing this bug on purpose😂 

Adding the `Breaking Change` label would be too much.

PR #2871, opened 2026-06-14
Improve access control on asset import / discovery for agents

ebariaux:
Branch mentions Velbus but applies to all asset import/discovery


MartinaeyNL (approved):

MartinaeyNL:
A reminder for the user adding this PR to the next release; Add a very simple note on this "breaking change in the HTTP API".
Because maybe people were abusing this bug on purpose😂 

Adding the `Breaking Change` label would be too much.

PR #2377, opened 2025-12-19
New Vaadin UI integration

MartinaeyNL:
## Description
This PR replaces almost all of the `<or-mwc-input>` components with the new `<or-vaadin-*>` design system.
It's a very large one, with a lot of changes, that I'll try to summarize in the changelog.
Feel free to test out this branch, and see the changes for yourself.

### Migrating Material UI to Vaadin
- Migrated most of the input fields from Material UI to [Vaadin Web Components](https://vaadin.com/docs/latest/components).
This is done by hand, where all Material components are changed to the new `<or-vaadin-*>` format.
Most of the HTML attributes are similar, or exactly the same, as Vaadin follows the common Web API specification.
It's also common to use slotted elements now, for example displaying labels with `<or-translate>`.
In my personal experience, the syntax is more professional and easier to understand.
Fixes #2593 
- Deprecated the old `<or-mwc-*>` components, as they'll be unmaintained after merging this.
- Bump the Vaadin components dependency to the latest version, `v25.1.3`
### Visual changes
- Replaced the default Manager UI font from system defaults (like Segoe UI) to use [Inter](https://fonts.google.com/specimen/Inter) for every operating system.
The typefont files are now bundled in `@openremote/theme` using RSPack.
For the Manager UI, it acts similar to how we handle and bundle images.
### Bugfixes
- Improved input validity checks across several input fields, to prevent malicious input.
- Improved Rules editor when the user only has READ access.
Fixes #2682
- Bugfix in Chart and Bar Chart widgets, where time presets could be selected while a custom timeframe was set.
- Data export page is now only accessible for users with `READ_ASSETS` role.
### Other
- Minor adjustments of app and component tests, removing hardcoded `<or-mwc-*>`, and simplifying the syntax.
- Added experimental RSBuild tasks in the `package.json`, to allow testing with RSBuild tooling.
We already use it for Storybook, but might for all apps in the foreseeable future. Unstable as of now.
- Cleanup of unused UI code.

## Checklist
- [ ] 1. Acceptance criteria of the linked issue(s) are met
- [ ] 2. Tests are written and all tests pass
- [ ] 3. Changes are manually tested by you and the reviewer
- [ ] 4. Documentation is written or updated

PR #2875, opened 2026-06-15
UI bugfix for asset pagination in Rules UI not working anymore

### LINKED ISSUE #2874: Rules UI asset selector only shows first 200 assets after Vaadin input migration
## Summary

After the Vaadin UI migration in #2377, the asset selector in the When-Then JSON Rules UI no longer uses the scalable remote-search behaviour that was introduced in #2271.

When there are 200 or more assets with rule state, the new Vaadin-based selector only lists a maximum of 200 assets. Assets outside that initial result set cannot be found/selected through the selector.

This is a regression from the behaviour implemented in #2271, where the Rules UI used a searchable input that queried assets remotely while typing, so larger asset sets remained usable.

## Steps to reproduce

1. Use an environment with at least 200 assets that have rule state / are available for selection in the When-Then JSON Rules UI.
2. Open the Rules UI and create or edit a When-Then JSON rule.
3. Open an asset selector, for example in a When or Then clause.
4. Try to find/select an asset that is not included in the first 200 listed assets.

## Actual behaviour

The new Vaadin selector lists at most 200 assets.

Searching/filtering does not expose assets outside that initial 200-item result set, so some assets cannot be selected.

## Expected behaviour

The selector should keep the scalable behaviour from #2271:

- for larger asset sets, assets should be queried remotely while typing;
- the user should be able to search for and select any matching asset, not only assets from the first page/result set;
- the UI should remain usable with large numbers of assets.

## Regression

Likely introduced by #2377, which migrated the input components to the new Vaadin-based components.

This appears to have overwritten or bypassed the remote-search asset selector behaviour added in #2271.

## Suggested fix

Restore the remote asset search behaviour for the Vaadin-based asset selector in the When-Then JSON Rules UI.

The selector should not rely only on the initially loaded/paginated asset list when the number of assets exceeds the remote-search threshold.

## Acceptance criteria

- With 200+ selectable assets, assets outside the first 200 can be found by typing in the selector.
- The selector uses remote querying/searching for large asset sets as implemented in #2271.
- Existing tests for the large-asset Rules UI scenario are restored or extended to cover the Vaadin selector.

MartinaeyNL:
## Description
Fixes #2874 

## Changelog
- Add `assetDataProvider` that dynamically fetches assets based on user search,
like it was before #2377 was introduced.
- Bugfix where the asset search would only look for exact matches, versus "it containing parts of the string"

## Checklist


MartinaeyNL:
Resolved all review comments. Ready for review again 👍

wborn (approved):

PR #2882, opened 2026-06-16
Fix timezone handling for datapoints export

### LINKED ISSUE #2881: CSV datapoint export misses rows when PostgreSQL session timezone differs from JVM timezone
**Description**
The asset datapoint CSV export can return only the header row when the PostgreSQL session timezone differs from the JVM timezone.

This was observed locally with PostgreSQL using Europe/Amsterdam while the test JVM uses UTC via the pseudo clock. CI likely passes because PostgreSQL runs with a UTC session timezone, masking the issue.

**Observed behavior**
AssetDatapointExportTest writes five datapoints within the requested export window, but the default CSV export returns:

timestamp,name,attribute_name,value

The test fails with:

  csvExport1Lines.size() == 6
  csvExport1Lines.size() == 1

**Expected behavior**
The CSV export should return the header plus all five datapoints, independent of the PostgreSQL session timezone.

**Likely cause**
asset_datapoint.timestamp is a timestamp without time zone, but the export SQL filters with PostgreSQL to_timestamp(...), which returns timestamp with time zone.

That means comparisons like this are timezone-sensitive:

  ad.timestamp >= to_timestamp(...)

Other datapoint query paths already use LocalDateTime parameters, so they do not hit this mismatch in the same way.

**Proposed fix**
Update the export path to compare timestamp-without-time-zone values consistently. For example:

  - Convert fromTimestamp / toTimestamp to LocalDateTime using the JVM default zone, matching existing datapoint query behavior.
  - Avoid to_timestamp(...) in the export SQL filter.
  - For the COPY TO STDOUT query, use a temporary bounds table populated via PreparedStatement, then filter with:

  ad.timestamp >= bounds.from_timestamp
  and ad.timestamp <= bounds.to_timestamp

**Test coverage**
Add or update a regression test so CI reproduces the issue by explicitly setting the database session timezone to Europe/Amsterdam while the JVM timezone remains UTC/non-Amsterdam. This should fail before the fix and pass after the export SQL is corrected.

ebariaux:


Ekhorn (commented):
Reproduced the issue locally on master and it seems indeed resolved.

Tested with a 5 million datapoint export before and after. Will do 1 with 100million datapoints.

Ekhorn (commented):
The regular export query seems to be ~0.5-1.4% faster now for 100million datapoints (but likely within the margin of error).

Also crosstab seems to be about the same, though at 100million datapoints the regular crosstab query seems to silently fail, but that is a separate issue (which was already present).

test/src/test/groovy/org/openremote/test/assets/AssetDatapointExportTest.groovy
```diff
@@ -379,4 +399,53 @@ class AssetDatapointExportTest extends Specification implements ManagerContainer
...
+                    resultSet.close()
+                }
+            } finally {
+                statement.close()
+            }
+        }
+    }
+
+    private static <T> T withConnection(PersistenceService persistenceService, Closure<T> closure) {
+        def dataSource = persistenceService.persistenceUnitProperties.get(AvailableSettings.DATASOURCE) as DataSource
```
richturner: AvailableSettings.DATASOURCE is deprecated

test/src/test/groovy/org/openremote/test/assets/AssetDatapointExportTest.groovy
```diff
@@ -379,4 +399,53 @@ class AssetDatapointExportTest extends Specification implements ManagerContainer
         }
     }
 
+    private static void setDatabaseSessionTimeZone(PersistenceService persistenceService, String timeZone, int connectionCount) {
+        def dataSource = persistenceService.persistenceUnitProperties.get(AvailableSettings.DATASOURCE) as DataSource
```
richturner: AvailableSettings.DATASOURCE is deprecated

richturner (commented):
Although this approach solves the situation where JVM and DB timezones are different it is still fragile in that data points are being commited in non UTC time which was an oversight on how LocalDateTime conversion happens within JDBC. My wrong understanding was that JDBC would convert the LocalDateTime back into UTC for timestamp without timezone columns but it seems to be the complete opposite. If an instance suddenly changes timezone then data overwrites or gaps can appear.

manager/src/main/java/org/openremote/manager/datapoint/AssetDatapointService.java
```diff
@@ -253,13 +254,17 @@ protected PipedInputStream doExportDatapoints(AttributeRef[] attributeRefs,
...
 
-                    // Create temp table and insert attribute refs safely via PreparedStatement
-                    // to prevent SQL injection (instead of string-interpolating IDs into the query)
+                    // Create temp tables and insert parameters safely via PreparedStatement
+                    // to prevent SQL injection (instead of string-interpolating IDs into the query).
                     try (Statement statement = connection.createStatement()) {
                         statement.execute(
                             "create temp table " + tempTableName + " (entity_id text, attribute_name text) on commit drop"
                         );
+                        statement.execute(
```
richturner, line 267: Couldn't a single temp table rather than two be used for all parameterised values?

ebariaux: Both are possible but I still preferred this option, as we have only 1 period for all attributes so 1 row in 2nd table but multiple rows in first table, one for each exported attribute.
Putting everything in 1 table would mean repeating the same period values for all attributes.
I added a comment to explain the reason.

richturner (approved):

Issue #2326, opened 2025-12-10
UdpClientProtocolTest unstable


wborn:
This test often fails for me when building locally.

Logging:

```
Condition not satisfied after 10.00 seconds and 51 attempts
Condition not satisfied after 10.00 seconds and 51 attempts
	at spock.util.concurrent.PollingConditions.within(PollingConditions.java:206)
	at spock.util.concurrent.PollingConditions.eventually(PollingConditions.java:158)
	at org.openremote.test.protocol.udp.UdpClientProtocolTest.Check UDP client protocol and linked attribute deployment(UdpClientProtocolTest.groovy:324)
Caused by: Condition not satisfied:

lastBytes != null
|         |
null      false

	at org.openremote.test.protocol.udp.UdpClientProtocolTest.Check UDP client protocol and linked attribute deployment_closure18(UdpClientProtocolTest.groovy:325)
	at org.openremote.test.protocol.udp.UdpClientProtocolTest.Check UDP client protocol and linked attribute deployment_closure18(UdpClientProtocolTest.groovy)
	at spock.util.concurrent.PollingConditions.within(PollingConditions.java:186)
	... 2 more


Condition not satisfied:

lastBytes != null
|         |
null      false

Condition not satisfied:

lastBytes != null
|         |
null      false

	at org.openremote.test.protocol.udp.UdpClientProtocolTest.Check UDP client protocol and linked attribute deployment_closure18(UdpClientProtocolTest.groovy:325)
	at org.openremote.test.protocol.udp.UdpClientProtocolTest.Check UDP client protocol and linked attribute deployment_closure18(UdpClientProtocolTest.groovy)
	at spock.util.concurrent.PollingConditions.within(PollingConditions.java:186)
	at spock.util.concurrent.PollingConditions.eventually(PollingConditions.java:158)
	at org.openremote.test.protocol.udp.UdpClientProtocolTest.Check UDP client protocol and linked attribute deployment(UdpClientProtocolTest.groovy:324)
```

Based on the logging there seems to be a resource problem:

```
2025-12-10 09:28:07.073  FINE    [ContainerExecutor-11          ] otocol.io.AbstractNettyIOClient.PROTOCOL : Connection attempt '1' for: udp://127.0.0.1:34419 (bindPort: 46841)
...
[...]

PR #2893, opened 2026-06-17
Stop using SO_REUSEADDR to avoid false connection reporting

richturner:
## Description
This change removes the usage of the `SO_REUSEADDR` channel option as it caused false connection reporting, particularly on Windows. This modification aims to improve connection reliability and eliminate misleading behavior.


wborn (commented):
I think this change will make the `UdpClientProtocolTest` unstable again as it was introduced to fix that, see: https://github.com/openremote/openremote/issues/2326

richturner:
Hmmm. So on Windows the previous fix has caused the original problem seen on Linux.

With the reuse address option enabled Windows silently allows a reconnect to an already binded socket and then the test tries to send a message and it goes to the old stale socket connection.

Removing the flag caused the address already in use error but the reconnect logic does reconnect after a couple of attempts. There were various improvements made to the IO client reconnect logic since the #2326 `fix` was made maybe that was the real reason it was not succeeding.

@wborn are you able to test this branch on your Linux machine in the repeating loop like you did previously to see if the test still fails after ~10 attempts please?

wborn:
I gave it a test but it failed again with these changes:

```
> Task :test:test

org.openremote.test.protocol.udp.UdpClientProtocolTest > Check UDP client protocol and linked attribute deployment took: 23379ms
org.openremote.test.protocol.udp.UdpClientProtocolTest

  Test Check UDP client protocol and linked attribute deployment FAILED (23.3s)

  Condition not satisfied after 10.00 seconds and 51 attempts
      at app//spock.util.concurrent.PollingConditions.within(PollingConditions.java:205)
      at app//spock.util.concurrent.PollingConditions.eventually(PollingConditions.java:157)
      at org.openremote.test.protocol.udp.UdpClientProtocolTest.Check UDP client protocol and linked attribute deployment(UdpClientProtocolTest.groovy:324)
  Caused by: Condition not satisfied:
  
  lastBytes != null
  |         |
  null      false
      at org.openremote.test.protocol.udp.UdpClientProtocolTest.Check UDP client protocol and linked attribute deployment_closure18(UdpClientProtocolTest.groovy:325)
      at org.openremote.test.protocol.udp.UdpClientProtocolTest.Check UDP client protocol and linked attribute deployment_closure18(UdpClientProtocolTest.groovy)
      at app//spock.util.concurrent.PollingConditions.within(PollingConditions.java:185)
      ... 2 more
```

wborn (changes requested):
I think we should handle the test side explicitly here: either disable the affected test on platforms where this behavior is unreliable, or update the test so it deterministically releases the port/resources before asserting connection state.

Right now the production change removes the false-positive behavior, but if the test still depends on timing or OS-specific socket reuse behavior, it may remain flaky or start failing for the wrong reason.

wborn:
AI suggestions for fixing this:

---

I don't think removing `SO_REUSEADDR` alone properly fixes this issue.

It resolves the Windows behaviour where a new socket can bind successfully while the old socket still owns the port, but it also restores the original Linux failure from #2326. I reproduced that failure on this branch: after the protocol is recreated, the first message can still be lost because the replacement UDP client has not successfully bound yet.

## Likely root cause

The underlying problem appears to be a race in the `AbstractNettyIOClient` lifecycle:

1. A connection attempt starts an asynchronous `bootstrap.bind(...)`.
2. The client is disconnected or replaced while that bind is still pending.
3. `disconnect()` cleans up the currently known channel and cancels the Failsafe retry future.
4. The Netty bind operation can still finish afterward.
5. `initChannel()` then stores this late channel in the shared `channel` field.
6. That stale channel can keep the UDP port bound or interfere with a newer connection attempt.

The current `CompletableFuture` wrapper does not propagate cancellation back to the underlying Netty `ChannelFuture`; it only observes its eventual result. Also, the current concurrency test only covers disconnecting before the retry task starts, not disconnecting while `bind()` or `connect()` is already in progress.

`SO_REUSEADDR` hid this problem on Linux by allowing overlapping binds. On Windows, that results in datagrams potentially continuing to go to the old socket, which explains the false `CONNECTED` status observed here.

## Suggested fix

I think this should be fixed in `AbstractNettyIOClient`, rather than only changing `UDPIOClient`.

### Track the channel belonging to each connection attempt

`startChannel()` should expose the channel immediately, together with the future indicating when it is ready. For example:

```java
protected record ChannelStart(
    Channel channel,
    CompletableFuture<Void> readyFuture
) {
}
```

For UDP:

```java
@Override
protected ChannelStart startChannel() {
    ChannelFuture bindFuture = bootstrap.bind("0.0.0.0", bindPort);
    return new ChannelStart(
        bindFuture.channel(),
        toCompletableFuture(bindFuture)
    );
}
```

This makes it possible to close a pending channel even before the bind has completed and before `initChannel()` has assigned it to the shared `channel` field.

### Invalidate stale connection attempts

Each connection attempt should have a generation or attempt ID. When `disconnect()` is called, that generation should be incremented so that any bind or connect completing afterward can detect that it is stale and immediately close its channel.

Conceptually:

```java
long generation;

synchronized (this) {
    generation = ++connectionGeneration;
}

ChannelStart start = startChannel();

try {
    start.readyFuture().join();

    synchronized (this) {
        if (generation != connectionGeneration
            || connectionStatus != ConnectionStatus.CONNECTING
            || !start.channel().isActive()) {

            start.channel().close();
            throw new CancellationException("Connection attempt is stale");
        }

        channel = start.channel();
    }
} catch (Throwable e) {
    start.channel().close();
    throw e;
}
```

The channel belonging to a failed, timed-out or cancelled attempt must be closed before Failsafe starts another attempt.

### Close both active and pending channels during disconnect

`disconnect()` should invalidate the current generation and capture both the active and pending channels:

```java
synchronized (this) {
    connectionGeneration++;

    retry = connectRetry;
    connectRetry = null;

    connecting = connectFuture;
    connectFuture = null;

    activeChannel = channel;
    channel = null;

    connectingChannel = pendingChannel;
    pendingChannel = null;
}
```

The futures can then be cancelled and both channels closed outside the synchronized block.

### Make callbacks channel-specific

The close listener currently acts based on the global connection status. A late close callback from an obsolete channel could therefore affect a newer channel.

The callback should verify that it belongs to the currently active channel:

```java
channel.closeFuture().addListener(closedFuture -> {
    boolean reconnect = false;

    synchronized (this) {
        if (AbstractNettyIOClient.this.channel != channel) {
            return;
        }

        AbstractNettyIOClient.this.channel = null;

        if (connectionStatus == ConnectionStatus.CONNECTED) {
            onConnectionStatusChanged(ConnectionStatus.CONNECTING);
            reconnect = true;
        }
    }

    if (reconnect) {
        scheduleDoConnect(5000);
    }
});
```

The exception handler should similarly complete the future belonging to that specific connection attempt, rather than the shared `connectFuture`, which may already refer to another retry.

## Tests

It would be useful to add deterministic lifecycle tests to `NettyIOClientTest`:

1. Start a UDP bind and pause its completion.
2. Disconnect the client while the bind is pending.
3. Allow the old bind to complete.
4. Verify that its channel is closed.
5. Verify that another client can immediately bind the same port without `SO_REUSEADDR`.

Additional useful cases would be:

* attempt 1 completes after attempt 2 has already connected;
* a stale channel's close callback must not reconnect or disconnect the active channel;
* a timed-out bind must close its channel before the retry begins.

The existing `UdpClientProtocolTest` should also remain as an integration regression test and be run repeatedly on both Linux and Windows where possible.

Once the lifecycle race is fixed, removing `SO_REUSEADDR` is appropriate. The temporary logging changes in `UdpClientTest` can then also be removed.

PR #2916, opened 2026-06-22
Asset soft delete and background purge task

### LINKED ISSUE #2972: Asynchronous Asset and datapoint deletion
## User Story
Deleted assets can have a lot of compressed data points associated with them which can take a long time to fully purge so rather than synchronously purging everything a soft asset delete mechanism should be implemented and the actual deletion task broken down into smaller chunks to minimise load on the DB.

## Acceptance Criteria
- [ ] Clients and services should still see the soft delete as a regular deletion
- [ ] Trying to merge or reparent an asset onto an asset that is marked as delete pending should throw a conflict
- [ ] A PersistenceEvent should be raised when the asset(s) is finally deleted.

richturner:
## Description
This PR changes asset hard deletion to a soft deletion without making breaking changes to existing APIs.

The reason for this change is to handle situations where assets have millions of data points which must also be purged, this data point purging is now handled in batches in a background process, if the batched data point deletion fails for some reason the existing FK cascade will take over on the asset deletion (as the mechanism was previously), there is retry capability on the failed asset deletions and resume on startup.

A new DELETE_FINISHED persistence event is emitted when the asset is finally deleted but until then an asset merge with the same ID will fail. 

This PR contains #2996 which should be reviewed and merged first.

## Checklist

- Test updated to cover new code


wborn:
@codex review

richturner:
@wborn can we get another codex review please 😃

wborn:
@codex review

manager/src/main/java/org/openremote/manager/asset/AssetStorageService.java
```diff
@@ -907,6 +974,225 @@ public boolean delete(List<String> assetIds, boolean skipGatewayCheck) {
...
+        long start = System.currentTimeMillis();
+
+        if (datapointCount > assetDeleteDatapointBatchThreshold) {
+            LOG.fine("Purging asset datapoints: assetId=" + assetId + ", datapoints=" + datapointCount + ", threshold=" + assetDeleteDatapointBatchThreshold);
+            try {
+                deleteAssetDatapointsByAsset(assetId);
+                LOG.fine("Purged asset datapoints: assetId=" + assetId + ", durationMillis=" + (System.currentTimeMillis() - start));
+            } catch (Exception e) {
+                LOG.log(WARNING, "Purge asset datapoints failed: assetId=" + assetId, e);
+            }
```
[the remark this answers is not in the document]
richturner: The foreign key cascade will take care of the datapoints and is the fallback

richturner:
This PR is ready for another review

richturner:
Not looked into UI test failures

manager/src/main/resources/org/openremote/manager/setup/database/V20260629_1100__AssetDeletePending.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE ${schemaName}.asset
+    ADD COLUMN IF NOT EXISTS delete_pending boolean NOT NULL DEFAULT false;
```
wborn, line 2: This migration version is older than a migration already present on the base branch (`V20260715_1444__AddNotificationRoles`), while `PersistenceService` configures Flyway with `outOfOrder=false` by default. An installation already migrated to current `master` therefore cannot apply this as a normal later migration. Please rename it to a version newer than the latest migration on the base branch so existing installations reliably get the `delete_pending` column.

model/src/main/java/org/openremote/model/asset/Asset.java
```diff
@@ -460,6 +465,16 @@ public T setAccessPublicRead(boolean accessPublicRead) {
     return (T) this;
   }
 
+  @JsonInclude(JsonInclude.Include.NON_DEFAULT)
+  public boolean isDeletePending() {
+    return deletePending;
+  }
+
+  public T setDeletePending(boolean deletePending) {
```
wborn, line 473: `deletePending` is server-managed lifecycle state, but this setter also makes it writable through the Asset JSON model. The create path clones the supplied asset and passes it to `merge()` without clearing this field, so a client can create an asset with `deletePending: true`. That asset is hidden from normal queries but has never been queued for deletion. Please ignore or reject client-supplied delete state on creation and expose this as read-only API state.

model/src/main/java/org/openremote/model/alarm/AlarmAssetLink.java
```diff
@@ -100,11 +100,11 @@ public String toString() {
       columnDefinition = "TIMESTAMP WITH TIME ZONE")
   protected Instant createdOn;
 
-  @Formula("(select a.NAME from ASSET a where a.ID = ASSET_ID)")
+  @Formula("(select a.NAME from ASSET a where a.ID = ASSET_ID and a.DELETE_PENDING is false)")
```
wborn, line 103: Filtering `assetName` and `parentAssetName` prevents deleted asset names from being exposed, but the `AlarmAssetLink` itself remains queryable until physical deletion. More importantly, alarm-link creation still validates assets through `getAssetRealms()`, which does not exclude `deletePending` rows, so a new alarm link can be created after the asset has already been logically deleted. Please make alarm-link reads and validation treat pending assets as nonexistent rather than only hiding their names.

manager/src/main/java/org/openremote/manager/asset/AssetResourceImpl.java
```diff
@@ -261,6 +261,39 @@ public Asset<?> getPartial(RequestParams requestParams, String assetId) {
     return get(requestParams, assetId, false);
   }
 
+  @Override
+  public boolean isDeletePending(RequestParams requestParams, String assetId) {
```
wborn, line 265: This endpoint does not apply the restricted-user asset-link authorization used by the normal asset read path. A restricted user can therefore probe IDs for same-realm assets that are not linked to them and learn their delete-pending state. Please apply the equivalent restricted-user authorization before returning the status.

manager/src/main/java/org/openremote/manager/security/ManagerKeycloakIdentityProvider.java
```diff
@@ -1167,7 +1166,17 @@ public void deleteRealm(String realmName) throws ClientErrorException {
       throw new NotFoundException("Realm does not exist: " + realmName);
     }
 
+    int assetCount =
```
wborn, line 1169: Rejecting deletion of a realm that still contains assets appears to be an intentional consequence of making asset deletion asynchronous, but existing callers have not all been migrated to that new contract. The current UI app test run fails while deleting the seeded `smartcity` realm because cleanup still expects realm deletion to remove contained assets. Please update the affected callers/tests and make the Manager UX/API contract consistent with the new behavior before merging.

wborn (changes requested):

The overall approach is appropriate: marking an asset as logically deleted immediately and moving potentially expensive datapoint cleanup into the background avoids blocking asset deletion on very large histories.

Several issues still need to be addressed before merging. The new Flyway migration is ordered before a migration already present on the base branch, and clients can create assets with the server-managed `deletePending` state already set. The previously raised [batched-delete retry concern](https://github.com/openremote/openremote/pull/2916#discussion_r3533616320) also still applies on the current head: failed physical deletions are retried through `deletePendingAsset()` directly instead of re-entering the batching-aware purge path.

The soft-delete abstraction is also not consistently enforced across externally reachable APIs. Normal asset reads, datapoint reads, and datapoint exports correctly treat pending assets as deleted, but the previously raised [`AssetQuery.includeDeletePending` concern](https://github.com/openremote/openremote/pull/2916#discussion_r3533616322) still allows REST query/tree/count endpoints and client-event reads to expose them explicitly.

Dependent-link APIs have similar inconsistencies. `getUserAssetLinks()` queries `UserAssetLink` directly, so links to a logically deleted asset remain externally visible until the final FK cascade removes them. Alarm APIs likewise continue returning links to pending assets, and alarm-link creation can still accept an asset after it has been logically deleted because its validation path does not exclude `deletePending` rows.

The new delete-pending status endpoint also misses the restricted-user asset-link authorization used by normal asset reads. Finally, the intentional change to populated-realm deletion has not yet been propagated consistently to existing callers/tests, which is reflected by the current failing UI test run.

PR #2934, opened 2026-06-24
Enhancement upgrade from Jackson 2 to Jackson 3

ebariaux:
After upgrading the dependencies, fixing the issues has been done using an LLM, iterating on the failing tests.
This is its explanation of work that has been done.

The important thing is that this was not a simple import rename. It ended up being three migrations at once:

  1. move OpenRemote’s own runtime JSON handling to Jackson 3,
  2. keep REST compatibility with RESTEasy’s Jackson 2 provider,
  3. keep generated TypeScript and existing clients stable.

**Dependency Split**

Jackson 3 artifacts moved to tools.jackson.*, but RESTEasy still uses com.fasterxml.jackson.* through resteasy-jackson2-provider.

So the build now needs both families:

  - Jackson 2: RESTEasy, JSON schema tooling, compatibility bridge.
  - Jackson 3: OpenRemote runtime model serialization.

That is why gradle/libs.versions.toml:31 now keeps jackson = 2.21.3 and adds jackson3 = 3.1.3.

**Runtime Jackson 3 Migration**

The core model and runtime code moved from Jackson 2 databind/core APIs to Jackson 3 APIs:

  - com.fasterxml.jackson.databind.* to tools.jackson.databind.*
  - com.fasterxml.jackson.core.* to tools.jackson.core.*
  - serializer/deserializer API changes such as SerializerProvider to SerializationContext
  - updated custom serializers/deserializers in model classes like model/src/main/java/org/openremote/model/asset/agent/AgentLink.java:1, model/src/main/java/org/openremote/model/value/ValueDescriptor.java:1,
    and model/src/main/java/org/openremote/model/event/TriggeredEventSubscription.java:1

The shared runtime mapper setup moved accordingly in model/src/main/java/org/openremote/model/util/ValueUtil.java:1, container/src/main/java/org/openremote/container/json/JacksonConfig.java:1, and Hibernate JSON mapping in container/src/main/java/org/openremote/container/persistence/JsonFormatMapper.java:1.

**Jackson 2 REST Bridge**

The tricky part: HTTP REST payloads are still handled by RESTEasy’s Jackson 2 provider. Jackson 2 cannot see Jackson 3 databind annotations and cannot directly serialize Jackson 3 tree types like tools.jackson.databind.JsonNode.

So we added a bridge:

  - container/src/main/java/org/openremote/container/json/Jackson2Config.java:1
  - model/src/main/java/org/openremote/model/jackson/Jackson2ModelModule.java:1
  - Jackson 2 serializers/deserializers for Attribute, MetaMap, AgentLink, AssetDescriptor, Jackson 3 JsonNode, colors, and asset type IDs.

That bridge is what preserves existing REST contracts while internal code uses Jackson 3.

**Contract Preservation Fixes**

Some annotations had to exist in both worlds. The review comments were about this.

Examples:

  - Datapoint query time parsing: model/src/main/java/org/openremote/model/datapoint/query/AssetDatapointQuery.java:1 and model/src/main/java/org/openremote/model/datapoint/query/AssetDatapointQueryLocalDateTimeDeserializerJackson2.java:1

  - GeoJSON coordinates still serialize as GeoJSON arrays through Jackson 2: model/src/main/java/org/openremote/model/geo/GeoJSONPoint.java:1
  - Calendar recurrence still serializes as RRULE text: model/src/main/java/org/openremote/model/calendar/CalendarEvent.java:1
  - UserSession needed explicit Jackson 2 @JsonProperty annotations because RESTEasy deserializes it: model/src/main/java/org/openremote/model/security/UserSession.java:30

**TypeScript Generation**

  The TypeScript generator also needed care because it runs over the Java model classpath.

Changes included:

  - updating model-util extensions to recognize both com.fasterxml.jackson and tools.jackson
  - remapping Jackson 3 ObjectNode/tree types to stable TS shapes
  - fixing Gradle’s jackson2Configuration to use a real Jackson2Configuration, not a Groovy map, in project.gradle:333

**Tests Added Or Updated**

  The important new coverage is:

  - test/src/test/groovy/org/openremote/test/model/Jackson2ModelBridgeTest.groovy:1: verifies bridge behavior.
  - test/src/test/groovy/org/openremote/test/rest/Jackson2RestContractTest.groovy:1: raw REST contract coverage for the Jackson 2 provider.
  - model/src/test/java/org/openremote/model/security/UserSessionTest.java:1: direct regression for the RESTEasy/Jackson 2 UserSession[] failure.


Ekhorn:
It seems that this commit dc85a5eb2bceee2c9006b0d706de95de8be0934e causes additional properties in the AssetInfo response to be included when saving an asset responds with the following additional properties:


key | value
-- | --
assetName | "New Asset"
attributeNames | [ "notes", "location" ]
assetType | "ThingAsset"
assetClass | "org.openremote.model.asset.impl.ThingAsset"

Looking at the following method I think these above properties are not intended for clients https://github.com/openremote/openremote/blob/a2838215eccd33ef17b0180b5dd3d4253394f490/model/src/main/java/org/openremote/model/asset/Asset.java#L514

Also there was a slight change on the `createdOn` property value:

1782748630947 -> 1782747231.993297000

Looking at the rest of the test failures whether they are related or if it is something else.

PR #2943, opened 2026-06-25
Adjust dependabot rebase strategy

Ekhorn:
## Description

Disabled rebasing for Java and NPM ecosystems to avoid unnecessary pipeline runs.

I've also added a cooldown period for the Java ecosystem with the max allowed value of 90 days (may need to be adjusted). You can also specify different cooldown periods for major,minor,patch release versions.

See https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

Also see https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/dependabot-2.0.json

## Checklist


MartinaeyNL:
I would personally use a 7 day cooldown period (like NPM) on the Java dependency updates.
But everyone will probably have something to say about that haha
Definitely worth deciding upon.

Disabling the rebase strategy is 100% worth merging 👍

MartinaeyNL:
Or maybe, as you can define cooldown periods per major,minor,patch release;
Put `patch` and `minor` releases to 7 days, and put `major` releases to at least 30 days.

Ekhorn:
We've decided to revert the cooldown for now. Ideally a cooldown could be added for problematic dependencies, but that level of granularity requires some more discussion.

See pattern include/excludes in: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#configuration-of-cooldown

MartinaeyNL (approved):

Issue #1390, opened 2024-08-28
Notifications: remove old asset user links when a new one is created on a console asset


DonWillems:
At the moment if multiple users have logged in on a phone, a notifications for any of those users will be sent to that phone.

Make it so that any old asset-user links of a console are removed when a new one is created.
This way you will not receive notifications for users that are not currently logged in on a phone (console).

PR #2965, opened 2026-06-30
Align javadoc with what code actually does

ebariaux:
Code changes was made when fixing [Notifications: remove old asset user links when a new one is created on a console asset · Issue #1390 · openremote/openremote](https://github.com/openremote/openremote/issues/1390) but the javadoc was not updated at this time, this fixes it.


wborn (approved):
Thanks!

PR #2981, opened 2026-07-02
Implement or-vaadin-toggle

### LINKED ISSUE #2637: UI: Integrate new toggle design (or-vaadin-toggle)
## Description
As investigated (and agreed upon) in #2555, we'd like to build a new toggle component that is
in line with the Vaadin design specification. This is because the old Material Design implementation is
very out dated (nor maintained), and the design is nowhere close to our new design system.

This issue is completed once the new toggle design shown below has been integrated.
Image

## Implementation Notes
The plan is to stay as close as the Web / HTML specification as possible.
It's common that a `type="checkbox"` is used, but with different styling to act as a toggle.
You can find plenty of tutorials online on how to do this using CSS, without any HTML changes.
Maybe we can inherit a Vaadin component? Otherwise use a raw HTML element instead.

Ekhorn:
## Description

I connected the Figma MCP server plugin to claude-code (using Opus 4.8 on high effort) and asked it if it could see the toggle from the specified URL. That worked perfectly and then proceeded with the prompts below.

It seemed get most, if not all the styling correct, but it ran into a couple hurdles when it comes to Vaadin, Storybook and the Playwright component tests. Those were expected and I pointed those things out and it figured it out after ~18 prompts (in about 2-3hrs). The `AGENTS.md` file has been updated with this information so we can hopefully get closer to a one-shot prompt in the future.

- Closes #2637  

## Changelog

- Add `or-vaadin-toggle` component
- Add `or-vaadin-toggle` storybook stories
- Add `or-vaadin-toggle` component tests
- Update `AGENTS.md` with Vaadin, Storybook and component test information

## Prompts

> [!NOTE]  
> Certain prompts and responses have been left out like updating the `AGENTS.md` file

**Prompt 1:** Initial prompt - "Can you implement the new Vaadin based toggle and replace the old material UI toggles? The Vaadin toggle should be a new component under @ui/component/or-vaadin-components/src/"

Response ~2 mins later
> `Claude` The architecture here is clear now. There's a genuine scope decision that changes the work significantly, so let me confirm before building.
> 
> What I found:
> - Vaadin has no native switch, so the new or-vaadin-toggle will be a thin component (styling a Vaadin Checkbox as a toggle, matching the Figma design — green --lumo-primary-color track, white knob).
> - The old material toggle is InputType.SWITCH in or-mwc-input. It shows up in two very different ways:
>   a. Via the input pipeline — value-input-provider.ts maps boolean asOnOff/asOpenClosed attributes to SWITCH, currently falling back to or-mwc-input. Wiring the new toggle in here is a small, low-risk change (~3 files) that auto-replaces every attribute-driven switch.
>   b. Direct call sites — ~33 hand-written <or-mwc-input .type="${InputType.SWITCH}"> across ~13 files (page-realms, page-gateway, dashboard settings, or-rules, etc.), each with its own @or-mwc-input-changed handler, label, and styling. Converting these means rewriting each call site to the new element's API (checked prop, change/checked-changed event).

Then Claude asked:

> `Claude` For reference, the decision I'm trying to settle is how far the replacement should go — from just building the or-vaadin-toggle component, to wiring it into the input pipeline (auto-replaces attribute-driven switches, ~3 files), up to also rewriting all ~33 hand-written <or-mwc-input type="switch"> call sites across ~13 files.
> 
> What would you like to clarify?

**Prompt 2:** My response - "ideally pipeline and all callsites but or-json-forms."

5 mins later - Initial toggle implementation 15216b08abd62ac5e7e1cc9d3a3242df81dc9da0
30 mins later - Toggle migration fb46ad794cf1a2f3ea7671c4b18b7c0c4ab273cd

**Prompt 3:** well the styling doesn't apply at all when I look at it in storybook (also there isn't an entry for it under `@ui/app/storybook/docs/components/`). The component should integrate with the defined theme in `@ui/component/theme/` however it should also look like a toggle without the theme. So I'm not very confident directly using the checkbox will work out well. Ideally the component base is the same but we need full control over the styles here (even without applying the theme). You can also use `registerStyles`, but I don't know the limitations of that.

Resolved the issues in f9781c161c232e85a198f8e2ea0bde34cd2d4fc0

**Prompt 4:** Always align the vaadin versions in this case there is a 25.1.4 of the themable mixin.

> `Claude` Good catch needed on CSS specificity: the Vaadin base has a high-specificity :host(:not([checked],[indeterminate])) [part='checkbox']::after { opacity: 0 } rule that would hide my knob in the off state. Let me harden the selectors so my styles reliably win regardless of source order:

3 mins later - it also made some more styling changes

**Prompt 5:** write everything that took time to figure out into the agents.md make it about "how to add new vaadin components" and about "storybook"

44s later - resolved it in 9164a04dc7d0fdfc8987e8841904844795bc1764

**Prompt 6:** Storybook is not just for vaadin components it is for any component under @ui/component/

36s resolved

**Prompt 7:** Let's add some additional stories or at least I want to have a nice overview of all the toggle button variants in 1 go. So all attributes differences must be covered and a combined checkbox(toggle) group should be shown as well.

2m 18s later

**Prompt 8:** why isn't there helper text being shown for any of the toggles? Secondly the toggle group story should also show horizontal variant.

1m 57s - it added 2c8eb577622eddcbaf2d733248e4e175e30dd2cf

**Prompt 9:** Set gap 12 px horizontally and gap 2px vertically for the groups

 1m 8s later

**Prompt 10:** no you made the vertical 1 horizontal

 54s later fixed in c47accbff4ae0d5a66697e4b2a3ac02e1b9bef0c

**Prompt 11:** Now include all story scenarios as component tests, make sure to check the event values that come through.

1m 54s later + answering some questions 53370747a02df24c4d4fe13e24ae435b8d3b1cdd

**Prompt 12:** Some tests failed - pasted in all test failure output from Playwright UI

Screenshot From 2026-07-02 12-08-53

Interrupted, took a few minutes I think.

**Prompt 13:** for the groups the toggles are literally not rendered.
**Prompt 14:** It renders the group fine in storybook but only its label in the ui component tests.

Interrupted, it came up with a shitty solution.

**Prompt 15:** I mentioned something about the component not being registered and pointed it to the component mounting source code (ref https://github.com/sand4rt/playwright-ct-web/blob/master/playwright-ct-web/registerSource.mjs)

After 6m 6s - It resolved the issue with an ok solution for now: edadb54725ef2ff1ca49d5a9354f09c3cd83c24a

**Prompt 16:** I see you also changed the group tests while I wonder if that was even necessary in the first place. I think it would be better to use the web first assertions.

44s later - resolved it in 7516322a9146297775c73689c77235e2493519d1

**Prompt 17:** Yep they still work. Now update the agents.md and then look at why the "vertical groups" aren't actually vertical in the component tests.

2m 54s later - resolved in 6da2264333a60f2438449e112d018f3593647637

**Prompt 18:** Are you also checking that no events are fired when you click disabled/readonly toggles?

See 0e5c0279dbeee03145b60556aa9a584d87707b0a

## Checklist


ui/component/or-vaadin-components/src/value-input-provider.ts
```diff
@@ -105,7 +105,8 @@ export const getValueHolderInputTemplateProvider: ValueInputProviderGenerator =
...
     const format: ValueFormat | undefined = (valueHolder && ((valueHolder as MetaHolder).meta) || (valueDescriptor && (valueDescriptor as MetaHolder).meta) ? Util.getAttributeValueFormat(valueHolder as Attribute<any>, valueHolderDescriptor as AttributeDescriptor, assetType) : Util.getMetaValueFormat(valueHolder as Attribute<any>, valueHolderDescriptor as AttributeDescriptor, assetType));
 
-    const supportsVaadinInput = (type: InputType) => (OrVaadinInput.TEMPLATES.has(type) && !format?.asOnOff);
+    // A boolean rendered as an on/off toggle is now natively supported by or-vaadin-toggle, so it no longer needs the or-mwc-input fallback.
+    const supportsVaadinInput = (type: InputType) => (OrVaadinInput.TEMPLATES.has(type) && (type === InputType.SWITCH || !format?.asOnOff));
```
Ekhorn: Needs further looking at. The comment shouldn't point to or-mwc-input.

AGENTS.md
```diff
@@ -14,3 +14,49 @@ In sandboxed environments, add `--offline` (requires dependencies to already be
...
 Running `./gradlew clean` deletes the root `tmp/` directory that is mounted into PostgreSQL (see `profile/dev-testing.yml`), so recreate it and restart the stack before running tests again.
+
+## UI
+
+### Adding a new Vaadin component
+
+Vaadin-based components live in `ui/component/or-vaadin-components/src/` as thin wrappers that extend a Vaadin class and re-register it under an `or-vaadin-<name>` tag, e.g. `export class OrVaadinX extends (X as new () => X & LitElement)` with `@customElement("or-vaadin-x")`. Keep all `@vaadin/*` dependency versions aligned.
+
+**Styling the component's shadow-DOM internals (`[part='...']`) — do NOT use `static get styles()`.** Vaadin's `ThemableMixin.finalizeStyles()` injects `static` styles *before* the Lumo theme injector, so the theme overrides them and nothing appears to apply. Styles registered via `registerStyles()` are injected *after* the injector, so they always win. Recipe (see `or-vaadin-toggle.ts`):
+- Call `registerStyles("or-vaadin-<tag>", css\`...\`)` from `@vaadin/vaadin-themable-mixin/register-styles.js` at module top (before the element is finalized). Add `@vaadin/vaadin-themable-mixin` to the package deps.
```
Ekhorn:
```suggestion
- Call `registerStyles("or-vaadin-<tag>", css\`...\`)` from `@vaadin/vaadin-themable-mixin/register-styles.js` at module top (before the element is finalized).
```

AGENTS.md
```diff
@@ -14,3 +14,49 @@ In sandboxed environments, add `--offline` (requires dependencies to already be
...
+- Override `static get is() { return "or-vaadin-<tag>"; }`. The base class reports its own tag (e.g. `vaadin-checkbox`), and `registerStyles`/`getStylesForThis()` key off `is`; without this the styles either don't match or leak onto every instance of the base component.
+- Out-specify base rules with `:host(...)` prefixes (e.g. the checkbox base hides its marker with `:host(:not([checked])) [part='checkbox']::after { opacity: 0 }`).
+- Give every themed CSS var a hard-coded fallback (e.g. `var(--lumo-primary-color, #47a942)`) so the component also looks correct standalone, with no theme loaded. Theme integration otherwise lives in `ui/component/theme/src/components/` via Lumo module injection.
+
+To make a type usable through the input pipeline, register it in `or-vaadin-input.ts` (`TEMPLATES` map + a `getXTemplate`, and `nativeValue` if it is a boolean that exposes `checked` instead of `value`) and update `value-input-provider.ts` (boolean/checked types use the `checked` attribute, not `value`).
```
Ekhorn:
```suggestion
To make an input type usable through the input pipeline, register it in `or-vaadin-input.ts` (`TEMPLATES` map + a `getXTemplate`, and `nativeValue` if it is a boolean that exposes `checked` instead of `value`) and update `value-input-provider.ts` (boolean/checked types use the `checked` attribute, not `value`).
```

AGENTS.md
```diff
@@ -14,3 +14,49 @@ In sandboxed environments, add `--offline` (requires dependencies to already be
...
+
+### Storybook
+
+Component docs are MDX under `ui/app/storybook/docs/components/or-vaadin-<name>.mdx`; the stories are under `ui/component/or-vaadin-components/stories/or-vaadin-<name>.stories.ts`. To add one, copy an existing pair (e.g. `or-vaadin-checkbox`), keeping the relative story import in the MDX and the `ComponentDocs` block.
+
+Stories call `getORStorybookHelpers(tagName)` and `setCustomElementsManifest(customElements)` off `../custom-elements.json`. That manifest is generated by `npm run analyze`, so a newly added component will render but show no args/argTypes/description until the manifest is regenerated.
```
Ekhorn: Reads weird

AGENTS.md
```diff
@@ -14,3 +14,49 @@ In sandboxed environments, add `--offline` (requires dependencies to already be
...
+
+Stories call `getORStorybookHelpers(tagName)` (from `ui/component/storybook-utils.js`) and `setCustomElementsManifest(customElements)` off the package's own `../custom-elements.json`. That manifest is generated per-package by `npm run analyze` (cem), so a newly added component will render but show no args/argTypes/description until you regenerate its package's manifest. `custom-elements.json` (and `custom-elements-jsx.d.ts`) are generated and untracked.
+
+
+### Component testing
+
+Component tests use Playwright component testing (`@sand4rt/experimental-ct-web`). They live in each package's `test/*.test.ts`, import `{ ct, expect }` from `@openremote/test`, and `mount(ComponentClass, { props, slots, on })`. Run them with
+ `npm test` in the package (which does `tsc -b && playwright test`). Prefer web-first, role-based assertions (`getByRole("checkbox", { name }).toBeChecked()`, `toHaveCount(...)`) over poking at JS properties (`toHaveJSProperty`) or internal l
+ocators. A few non-obvious gotchas that cost real time:
```
Ekhorn:
```suggestion
ocators. Some important quirks to know about:
```

AGENTS.md
```diff
@@ -14,3 +14,49 @@ In sandboxed environments, add `--offline` (requires dependencies to already be
...
+
+- **Custom elements used as slotted/appended children must be eagerly registered.** Playwright CT turns each imported component into a *lazy* dynamic import that only runs when that component is `mount()`ed, so a child element that is never m
+ounted itself (e.g. `or-vaadin-toggle` slotted inside `or-vaadin-checkbox-group`) never gets `customElements.define`d and stays an inert, unupgraded tag — it appends to the DOM but does not render. Register such elements eagerly in the shared
+ browser bootstrap `ui/test/playwright/index.js` (where icons/theme are also registered) with a side-effect import. When you do, also add that component's package to `ui/test/package.json` dependencies, otherwise it is an undeclared dependency (knip fails and the module may not resolve) — e.g. registering `or-vaadin-toggle` requires `@openremote/or-vaadin-components` in `ui/test`.
+- **The `on` handler receives `event.detail`, not the event** (`listener(event.detail)`). Assert event values via events whose detail carries them — for Vaadin fields that is the `<prop>-changed` notify event (detail `{ value }`); the native
+`change` event has no useful detail so it can only be counted.
```
Ekhorn: We need to think about this.

AGENTS.md
```diff
@@ -14,3 +14,49 @@ In sandboxed environments, add `--offline` (requires dependencies to already be
...
+- **Vaadin fires an initial `<prop>-changed` at mount** (notify-on-first-commit), so a freshly mounted toggle emits `checked-changed(false)` before any interaction. Assert on the user-driven transitions (e.g. the last two values), not exact-a
+rray equality, and prove "emits nothing" with `not.toContain(true)` rather than `toEqual([])`.
+- **Vaadin `<prop>-changed` events do not bubble or compose**, so a listener on a parent/group will not catch a child field's event — attach it to the field itself.
```
Ekhorn: This has come up before and we might want a toggle group component??

ui/component/or-vaadin-components/stories/or-vaadin-toggle.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+            description: "Toggle is an input field representing a binary on/off choice. It reuses the Vaadin checkbox as its base, so it supports the same states (checked, disabled, readonly, indeterminate, required) and events (`change`, `checked-changed`)."
+        }
+    }
+};
+
+export const Primary: Story = {
+    args: {
+        label: "Toggle",
+        checked: true,
+        helperText: "Helper text shown below the toggle"
```
Ekhorn: This seems to be wrong

```suggestion
        "helper-text": "Helper text shown below the toggle"
```

ui/component/or-vaadin-components/stories/or-vaadin-toggle.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+    name: "Toggle group",
+    parameters: {
+        title: "Grouped toggles",
+        summary: "Multiple toggles can be combined inside an `<or-vaadin-checkbox-group>` to share a group label, helper text and validation. Under the theme the group is horizontal by default; use `theme=\"vertical\"` to stack them."
+    },
+    render: () => html`
+        <style>
+            or-vaadin-checkbox-group::part(group-field) {
+                gap: 2px 12px; /* row-gap (vertical) / column-gap (horizontal) */
+            }
```
Ekhorn: Discuss the gaps with Remy

ui/component/or-vaadin-components/stories/or-vaadin-toggle.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+            <or-vaadin-toggle label="Indeterminate" indeterminate></or-vaadin-toggle>
+            <or-vaadin-toggle label="Required" required></or-vaadin-toggle>
+            <or-vaadin-toggle label="With helper text" checked helper-text="Additional context shown below the toggle"></or-vaadin-toggle>
+            <or-vaadin-toggle checked title="No label"></or-vaadin-toggle>
+        </div>
+    `
+};
+
+/** Several toggles combined inside a checkbox group, in both orientations. */
+export const Group: Story = {
```
Ekhorn: Make story height `280px`.

ui/component/or-vaadin-components/stories/or-vaadin-toggle.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+export const Primary: Story = {
+    args: {
+        label: "Toggle",
+        checked: true,
+        helperText: "Helper text shown below the toggle"
+    }
+};
+
+/** Every attribute-driven visual state of the toggle in a single view. */
+export const States: Story = {
```
Ekhorn: Make story height `320px`.

ui/component/or-vaadin-components/src/or-vaadin-toggle.ts
```diff
@@ -0,0 +1,135 @@
...
+        /* Theme variables with standalone fallbacks */
+        --_or-toggle-track-width: 30px;
+        --_or-toggle-track-height: 20px;
+        --_or-toggle-knob-size: 14px;
+        --_or-toggle-knob-inset: 3px;
+        --_or-toggle-radius: var(--lumo-border-radius-l, 12px);
+        --_or-toggle-track-off-color: var(--lumo-contrast-30pct, #c4c8c4);
+        --_or-toggle-track-on-color: var(--lumo-primary-color, #47a942);
+        --_or-toggle-knob-color: var(--lumo-base-color, #ffffff);
+        --_or-toggle-gap: 6px;
```
MartinaeyNL: Why the underscore? I think it would make sense if they're interpreted as "modifiable" by consumers.
Or create separate variables that are allowed to be changed by consumers.

ui/component/or-vaadin-components/src/or-vaadin-toggle.ts
```diff
@@ -0,0 +1,135 @@
...
+        width: var(--_or-toggle-knob-size);
+        height: var(--_or-toggle-knob-size);
+        min-width: 0;
+        border-radius: 50%;
+        background: var(--_or-toggle-knob-color);
+        mask: none;
+        -webkit-mask: none;
+        filter: none;
+        transform: translateY(-50%);
+        transition: left 0.15s ease-in-out;
```
MartinaeyNL: I would use a CSS variable with `0.15s` as a fallback.
If we in the future would like to streamline transition durations, we would be able to do that.

ui/component/or-vaadin-components/src/value-input-provider.ts
MartinaeyNL: `format.asOnOff` would have to point to the new `or-vaadin-toggle` component.
So there is no need to exclude it anymore with a fallback to Material UI.

MartinaeyNL: Instead of excluding `InputType.SWITCH`, it should now point to the new `or-vaadin-toggle` component.
Same counts for the `format.asOnOff`, as there is no need anymore to fallback to Material UI.

ui/component/or-vaadin-components/stories/or-vaadin-toggle.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+            <or-vaadin-toggle label="Disabled, on" checked disabled></or-vaadin-toggle>
+            <or-vaadin-toggle label="Readonly, off" readonly></or-vaadin-toggle>
+            <or-vaadin-toggle label="Readonly, on" checked readonly></or-vaadin-toggle>
+            <or-vaadin-toggle label="Indeterminate" indeterminate></or-vaadin-toggle>
+            <or-vaadin-toggle label="Required" required></or-vaadin-toggle>
+            <or-vaadin-toggle label="With helper text" checked helper-text="Additional context shown below the toggle"></or-vaadin-toggle>
+            <or-vaadin-toggle checked title="No label"></or-vaadin-toggle>
+        </div>
+    `
+};
```
MartinaeyNL: I don't like the "summarize all attributes under one story", because that would not make sense.
Professionally, they would all need separate stories. **However**, we don't practice that to overcomplicate it.

I would personally group them as separate stories for "Disabled & Readonly", "Groups", and "Helper text".
You can provide examples, with for example usage of `required` and `checked` here an there.
And "Indeterminate" doesn't exist or work for toggles right?

ui/component/or-vaadin-components/stories/or-vaadin-toggle.stories.ts
MartinaeyNL: I wouldn't do CSS modifications to stories themselves, unless they're necessary to provide an example.
In this case, if we'd like to apply gaps, we should put it as a CSS variable in the theme, or by default in the component code.

ui/component/or-vaadin-components/test/or-vaadin-toggle.test.ts
```diff
@@ -0,0 +1,244 @@
...
+
+/*
+ * These tests mirror the scenarios in `stories/or-vaadin-toggle.stories.ts`:
+ * the Primary story, every variant of the "States & attributes" story, and the
+ * "Toggle group" story (vertical + horizontal).
+ *
+ * Note: the component-test runner (@sand4rt/experimental-ct-web) forwards only
+ * `event.detail` to `on` handlers, so the boolean state that comes through is
+ * asserted via the `checked-changed` event (detail = `{ value: boolean }`); the
+ * native `change` event carries no useful detail so it is only counted.
```
MartinaeyNL, line 33: ???

ui/component/or-vaadin-components/test/or-vaadin-toggle.test.ts
MartinaeyNL, line 1: Some of these tests might feel overdone, but I think it's good in this case.
Because we should prevent that breaking changes from Vaadin would alter behavior for the custom toggle.

Same counts for styling related changes; I would personally add some checks for those as well.
Would be pity if we upgrade Vaadin dependencies, and that we visually broke the toggle because of it.

ui/test/playwright/index.js
```diff
@@ -8,6 +8,12 @@ import HttpBackend from "i18next-http-backend";
 import { IconSets, OrIconSet, createMdiIconSet, createSvgIconSet } from "@openremote/or-icon";
 import { AssetEventCause, AssetModelUtil, ClientRole } from "@openremote/model";
 
+// Eagerly register web components that are used as slotted/appended children (rather than mounted
+// as the test root). Playwright CT lazy-imports a component's module only when it is `mount()`ed, so
+// a custom element that only appears inside another mounted component would otherwise never be
+// `customElements.define`d and would not upgrade (e.g. or-vaadin-toggle inside or-vaadin-checkbox-group).
+import "@openremote/or-vaadin-components/or-vaadin-toggle";
+
```
MartinaeyNL: I assume this is necessary, but it also feels odd.
Please double check.

Ekhorn: It is. We need a better mechanism for it. Because now it is not clear for what tests this component is meant to be registered.

ui/component/or-vaadin-components/stories/or-vaadin-toggle.stories.ts
Ekhorn: See https://github.com/openremote/openremote/pull/2981#discussion_r3535015580

ui/component/or-vaadin-components/src/or-vaadin-toggle.ts
Ekhorn, line 1: Somehow `font-weight: 500` is being applied to the text of the toggle, but not to the checkbox. It should be the default weight of 400.

ui/component/or-vaadin-components/stories/or-vaadin-toggle.stories.ts
Ekhorn: Will be spacing-xs spacing-m

wborn:
@codex review

ui/component/or-vaadin-components/src/value-input-provider.ts
```diff
@@ -105,7 +105,7 @@ export const getValueHolderInputTemplateProvider: ValueInputProviderGenerator =
     const constraints: ValueConstraint[] = (valueHolder && ((valueHolder as MetaHolder).meta) || (valueDescriptor && (valueDescriptor as MetaHolder).meta) ? Util.getAttributeValueConstraints(valueHolder as Attribute<any>, valueHolderDescriptor as AttributeDescriptor, assetType) : Util.getMetaValueConstraints(valueHolder as NameValueHolder<any>, valueHolderDescriptor as AttributeDescriptor, assetType)) || [];
     const format: ValueFormat | undefined = (valueHolder && ((valueHolder as MetaHolder).meta) || (valueDescriptor && (valueDescriptor as MetaHolder).meta) ? Util.getAttributeValueFormat(valueHolder as Attribute<any>, valueHolderDescriptor as AttributeDescriptor, assetType) : Util.getMetaValueFormat(valueHolder as Attribute<any>, valueHolderDescriptor as AttributeDescriptor, assetType));
 
-    const supportsVaadinInput = (type: InputType) => (OrVaadinInput.TEMPLATES.has(type) && !format?.asOnOff);
+    const supportsVaadinInput = (type: InputType) => OrVaadinInput.TEMPLATES.has(type);
```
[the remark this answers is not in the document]
Ekhorn: Will be done in #2982

MartinaeyNL (approved):
Thanks! 👍 🥳

PR #2996, opened 2026-07-06
Improve container lifecycle management and cleanup logic

richturner:
## Summary

This PR improves container lifecycle handling and test container cleanup reliability.

The container now tracks its running state independently from the background waiting thread, which makes `isRunning()` reflect whether the container services have successfully started rather than whether `startBackground()` happened to create a keep-alive thread. This also keeps `start()` and `startBackground()` semantics distinct and avoids treating the waiting thread as lifecycle state.

## Changes

- Added an explicit `running` lifecycle flag to `Container`.
- Made `running` `volatile` so cross-thread lifecycle checks observe the latest state.
- Set `running = true` only after all services have successfully started.
- Added forced cleanup on startup failure so partially initialized services and executors are still stopped even though the container is not marked as running yet.
- Made `startBackground()` synchronized to prevent races with `stop()`.
- Avoided creating duplicate background waiting threads when `startBackground()` is called repeatedly.
- Made stop handling tolerant of containers that were started without a waiting thread.
- Guarded metrics registry cleanup when metrics are disabled.
- Simplified `ContainerTrait.stopContainer()` so fixture state and test clients are always cleared in `finally`.
- Fixed the recursive restart path in `ContainerTrait` to return the newly started container when agent deployment does not settle.

## Motivation

The previous implementation used `waitingThread != null` as the container running check. That worked for `startBackground()` users, but it did not accurately represent the lifecycle of containers started with `start()`, and it made cleanup behavior depend on whether a background thread had been created.

The test container reuse path also needed more robust cleanup so failed reuse/restart attempts do not leave stale fixture references or JAX-RS clients behind.


container/src/main/java/org/openremote/container/Container.java
```diff
@@ -150,13 +151,14 @@ public boolean isDevMode() {
...
     public boolean isRunning() {
-        return waitingThread != null;
+        return running;
     }
 
     public synchronized void start() throws Exception {
         if (isRunning())
             return;
         LOG.log(INFO, ">>> Starting runtime container...");
+        running = true;
```
ebariaux: Do we want to set it this soon in the method as init has not been done yet ?
Why not at exit of this method?

This would also more closely match the old semantics of isRunning() as waitingThread is set in startBackground() after start() returns.

richturner: This was an AI produced PR after I saw some strange race conditions in tests while the cleanup logic was deliberately broken.

This change seems sensible to me as the flag is set in the synchonrised method now whereas previously it wasn't and stop could return even though waitingThread gets populated

richturner (commented):
See my comment

container/src/main/java/org/openremote/container/Container.java
ebariaux: I don't think there's an issue in the current code base, but isRunning() is not synchronized and is public.
Highly unlikely / no valid reason to do so but if I would call start() in a background thread and call isRunning() in //, I would receive true while the init is in progress.
Making isRunning synchronized would fix this but maybe it's too big performance hit.

richturner: I've done another pass of this please take a look @ebariaux

richturner:
@ebariaux seems like it was a flaky test good for final review

denniskuijs (approved):

PR #3017, opened 2026-07-11
fix: commit location coordinates input on blur so Save enables

### LINKED ISSUE #695: Save button in modify mode does not work
**Describe the bug**
When change the value of location directly in modify mode, only when press enter button will enable the save button

**To Reproduce**
1. Go to modify mode
2. Change value of location directly instead of go to map panel
3. Click somewhere else 

Recap the process and press Enter button in step3, Save button is enabled

**Expected behavior**
Save button should be enabled as long as there is an update

mvanhorn:
## Description
Typing new coordinates into the location attribute's text input and clicking away now commits the value and enables Save. Previously the edit was silently dropped unless the user pressed Enter (reported in #695 with a GIF reproduction; re-confirmed by a contributor on 2026-06-29).

The cause: `getCoordinatesInputKeyHandler` in `ui/component/or-map/src/controls/coordinates.ts` only invokes `valueChangedHandler` on Enter/NumpadEnter keyup, and neither call site — `CoordinatesControl.onAdd` (the coordinates input on the map in modify mode) nor the compact-mode inline text field in `ui/component/or-map/src/index.ts` — listens for `change`, so a blur after an edit committed nothing and Save stayed disabled.

The parse-and-commit logic (split "lat, lng", build `LngLat`, call `valueChangedHandler`) is extracted into a shared function; the Enter keyup path is unchanged. A `change` listener is registered in `CoordinatesControl.onAdd` and as `@change` on the compact-mode field. Vaadin text fields fire `change` on blur after an edit, so clicking away now commits the value and `valueChangeNotifier` fires, enabling Save.

Added a Playwright component test under `ui/component/or-map/test` following the repo's component-test convention, covering commit-on-change (blur) and commit-on-Enter.

Fixes #695

## Checklist


wborn:
@codex review

Ekhorn (changes requested):
@mvanhorn Thanks for the contribution!

The implementation resolves the issue, though the input validation is incomplete e.g. consider 180/-180 lon constraints.

Also the tests don't run in the pipeline, because the `build.gradle` file doesn't register the `npmTest` task. Locally the tests fail to compile. I've updated the branch which includes important information for writing UI component tests in the `AGENTS.md` file. The map controls are non-standard UI components and no component test infrastructure exists yet. Please take inspiration from the recently added `beforeMount` hook in the `ui/test/playwright/index.js` file, which we use to register additional web components in e.g. `ui/component/or-vaadin-components/test/or-vaadin-toggle.test.ts`.

Please align the test code with the `AGENTS.md` file.

mvanhorn:
Thanks for the pointers to `AGENTS.md` and the `beforeMount` hook, both were exactly what I needed. Pushed 492ca81.

Validation: longitude is now bounded to -180..180 alongside the existing latitude bound, and the input is parsed with an anchored pattern instead of a naive split on `[ ,]`. The old split read the two numbers positionally, so `Lat: 52.37, Lng: 4.89` was parsed in the wrong order; the pattern now has separate labeled and bare alternatives. Clearing the field commits `null` rather than `undefined`, and whitespace-only input counts as cleared.

Pipeline: registered `npmTest` and `npmTestUI` in `ui/component/or-map/build.gradle`, copied from a sibling package as `AGENTS.md` describes. That was why the tests never ran in CI.

Tests: the control is not mountable directly, so the test-only element lives in `test/fixtures/coordinates-control.js` as plain JS, per the note that CT compiles `.ts` against `ui/test/tsconfig.json` and a TS component elsewhere fails with TS6059.

One quirk worth recording, since it cost me a cycle: two `mount()` calls in a single `ct()` block hang and then fail on the 30s test timeout rather than erroring. The boundary cases originally paired the low and high value in one test each; splitting them to one mount per test made them pass. Might be worth a line in `AGENTS.md` next to the other CT quirks.

Verified: `yarn test` in `ui/component/or-map` gives 10 passed. I did need `./gradlew :ui:component:model:build` first, since `@openremote/model` is generated and `tsc -b` cannot resolve it otherwise.

Ekhorn:
> One quirk worth recording, since it cost me a cycle: two `mount()` calls in a single `ct()` block hang and then fail on the 30s test timeout rather than erroring. 

Yeah faced this as well once - having multiple `mount()` calls in the same test doesn't resolve to separate locator paths, thus you'll probably get a timeout on a strict mode violation.

@mvanhorn If you don't mind adding an entry for it in the `AGENTS.md`, would be great to have documented. Just mention to avoid doing this.

mvanhorn:
Recorded it in `AGENTS.md` next to the other component-test quirks (05d3db1), since you'd hit it too -- figured it was worth saving the next person the cycle. Happy to drop that commit if you'd rather keep this PR to just the coordinates fix.

Ekhorn (approved):
Thanks!

mvanhorn:
Grateful for the merge. Committing coordinates on blur so Save actually enables is a small thing that was quietly annoying, thanks.

PR #3018, opened 2026-07-13
Only apply delay when enabling providers

Ekhorn:
## Description

In c4b36206d5a139c826139f2f00caa68637e1b796 the `console.ts` provider changes applied a delay that was never applied before which causes the console registration to take 2s, which resulted in multiple UI app tests failing when navigating to the assets page and increased the test duration for each test by 2 seconds.

~~These changes adhere to the following comment, without defaulting to a 2 second delay for every `sendRegistration` call https://github.com/openremote/openremote/blob/4f208d7071a361653035db6d77f7c74de7944007/ui/component/core/src/console.ts#L301~~

I've decided to remove the delay. After hardcoding providers and testing these changes it seems that it waits for 2 seconds for every provider synchronously. Previously it didn't delay, so I've removed the delay now.

## Checklist

### DIFF
--- a/ui/component/core/src/console.ts
+++ b/ui/component/core/src/console.ts
@@ -174,7 +174,7 @@ export class Console {
 
             // Get an ID for this console if it doesn't have one
             if (!this._registration.id) {
-                await this.sendRegistration(0);
+                await this.sendRegistration();
             }
 
             this._initialised = true;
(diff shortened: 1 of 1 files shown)


ui/component/core/src/console.ts
```diff
@@ -237,7 +237,7 @@ export class Console {
         if (index >= 0) {
             this._pendingProviderEnables.splice(index, 1);
             if (this._pendingProviderEnables.length === 0) {
-                this.sendRegistration().catch(() => undefined);
+                this.sendRegistration(2000).catch(() => undefined);
```
richturner: The intention here is to now do the backend registration call as all providers have responded to the `PROVIDER_ENABLE` message.

```suggestion
                this.sendRegistration(0).catch(() => undefined);
```

MartinaeyNL (dismissed):

richturner (approved):

PR #3027, opened 2026-07-14
Make simulator replay data a singular input

### LINKED ISSUE #2311: Add JSON Forms pagination/infinite scrolling
## User Story
As a [user] I want to have pagination/infinite scrolling for the JSON Forms, to remain working when I add a large dataset.

## Description

I tried adding a dataset with more than 8000 entries to the SimulatorAgentLink, but this made it unusable in Firefox and basically unusable on chrome.

## Acceptance Criteria
- [ ] 1. TBD - How many items do we want show at any time?

## Implementation Notes

Consider adding a scrollable virtualized list.

### LINKED ISSUE #2922: Can't insert newlines in textarea on asset view page
**Describe the bug**
Inserting newlines on multi-line fields on the asset view page doesn't insert the new line but instead submits the existing value.

**To Reproduce**
1. Go to 'Asset page'
2. Create an asset
3. Add the 'Multiline' configuration item to the 'notes' attribute
4. Go to the asset view page and try to insert newlines (with shift + enter)
5. It just submits the value without the newline

**Expected behavior**
New lines to be inserted (on shift + enter)

**Additional context**

The issue is specific to asset view page submit logic and should be reviewed and tested.

Ekhorn:
## Description

This scales to at least 86400 entries now on all major browsers. The related issue describes ~48k entries so that use case should be covered by these changes.

I've had to do several optimizations besides just replacing the replay data forms with a singular input. I've also had to:

1. Cache the text and internal replay datapoint representation
2. Replace `objectsEqual` conditions with referential checks. The json-forms change notification falls back to a deep compare when the reference differs, so unchanged content no longer reports a change; `objectsEqual` got a linear index-based path for arrays (the key-based path is quadratic there) to keep that affordable
3. Most important for chromium browsers (but also a bit for Firefox): the new `manualresize` attribute disables Vaadin's automatic resizing, which caused a layout reflow for every line in the textarea

With `manualresize` the field starts at `min-rows`, keeps a fixed height, and shows the native resize handle; on touch devices (no usable handle) the height simply stays fixed. Attribute inputs default to `min-rows="5"`, matching the historical `or-mwc-input` default.

- Closes #2311

This PR also closes the following issue

- Closes #2922

## Changelog

- Make Simulator replay data render as a single textarea
- Add a native resize handle to the textarea (`manualresize`, hidden on touch devices)
- Fix shift+enter causing the textarea to submit
- Add Storybook stories and a docs page for `or-vaadin-text-area`
- Mark inherited underscore members as non-public in the custom elements manifest, so Storybook no longer overwrites Vaadin internals
- Fix `objectsEqual` being quadratic for arrays

## Checklist


Ekhorn:
Just tried with over 600k entries and it is still somewhat usable. It does start to take more than 10 seconds on initial load to expand the attributes and other operations also start to take several seconds.

ui/component/theme/src/default.css
```diff
@@ -26,6 +26,12 @@
 :where(:root),
 :where(:host) {
     --vaadin-lumo-theme: 1;
+
+    /* Distance from a labelled Lumo field's top edge to its input box:
+       host padding-top (space-m when has-label) plus the label block
+       (font-size-s, line-height 1, 0.5em bottom padding). Offset elements
+       placed next to a labelled field by this to align with the input box. */
+    --or-field-input-offset: calc(var(--lumo-space-m) + var(--lumo-font-size-s) * 1.5);
```
Ekhorn: We would need to take multiple different states into account: no label, with label etc.

I think we probably should just revert these changes to be picked up in #3034

ui/component/or-vaadin-components/src/or-vaadin-text-area.ts
```diff
@@ -17,16 +17,74 @@
...
-        this.dispatchEvent(new CustomEvent("submit", {bubbles: true, composed: true}));
+        // Shift+Enter inserts a newline without submitting the value.
+        if (!ev.shiftKey) {
+            this.dispatchEvent(new CustomEvent("submit", {bubbles: true, composed: true}));
+        }
         return super._onEnter(ev);
     }
+
+    // Vaadin's default _updateHeight reads scrollHeight on every value change — O(content lines).
+    // When --or-text-area-height is set, size via CSS variable (O(1)) instead.
```
MartinaeyNL: Is that `--or-text-area-height` variable documented? I think annotations would work
In terms of naming, I think it'd be better to stick to the HTML element name; `--or-vaadin-text-area-height`,
since it is specific to this component anyway (and not applicable to Material textareas for example)

ui/component/or-vaadin-components/src/or-vaadin-text-area.ts
```diff
@@ -17,16 +17,74 @@
...
+import {type LitElement, type PropertyValues} from "lit";
 
 @customElement("or-vaadin-text-area")
-export class OrVaadinTextArea extends (TextArea as new () => TextArea & LitElement) implements OrVaadinComponent {
+export class OrVaadinTextArea extends (TextArea as new () => TextArea & LitElement & { _updateHeight(): void }) implements OrVaadinComponent {
+
+    // Enables the native resize handle in fixed-height mode (--or-text-area-height).
+    // Irrelevant in the default auto-grow mode, where the height always tracks the content.
+    @property({type: Boolean})
+    resize = true;
```
MartinaeyNL: Using the default `true` wouldn't allow consumers to set it back to `false`, only when set through property.
The normal practice is to use default = `false`. So I guess it's either;
- Use `resize` attribute with default = `false`, or
- Use `static` (or similar) attribute with default = `false`, which inverts the mechanic.

ui/component/or-vaadin-components/src/or-vaadin-text-area.ts
```diff
@@ -17,16 +17,74 @@
...
+            // Fixed height disables auto-grow, so the textarea offers the native
+            // resize handle by default (inline style beats Vaadin's ::slotted()
+            // resize: none). min-height keeps drags from shrinking below default.
+            // Only an explicit false disables, so undefined behaves like the default.
+            if (this.resize !== false) {
+                this.inputElement.style.resize = "vertical";
+                this.inputElement.style.minHeight = h;
+            } else {
+                this.inputElement.style.resize = "none";
+                this.inputElement.style.minHeight = "";
```
MartinaeyNL: I don't know if it is easy to check, but what happens if the `resize` CSS variable is not supported by the browser?
I assume resize wouldn't work anymore (which is totally OK), but let's make sure it doesn't break the component.
(on Safari it seems to be recognised OK, but let's prevent syntax issues)

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/resize

ui/component/or-vaadin-components/src/or-vaadin-text-area.ts
```diff
@@ -17,16 +17,74 @@
...
-import {customElement} from "lit/decorators.js";
+import {customElement, property} from "lit/decorators.js";
 import {TextArea} from "@vaadin/text-area";
 import {OrVaadinComponent} from "./util";
-import {type LitElement} from "lit";
+import {type LitElement, type PropertyValues} from "lit";
 
 @customElement("or-vaadin-text-area")
-export class OrVaadinTextArea extends (TextArea as new () => TextArea & LitElement) implements OrVaadinComponent {
+export class OrVaadinTextArea extends (TextArea as new () => TextArea & LitElement & { _updateHeight(): void }) implements OrVaadinComponent {
```
MartinaeyNL, line 27: What's happening here? Is `_updateHeight` inherited from Vaadin or a custom function?
Based on this code, I'd assume it's from another Vaadin class that is not `TextArea`, but let's document that.

ui/component/theme/src/default.css
MartinaeyNL: I don't have the context here, but I think picking this up separately makes sense.
Maybe along with other Vaadin styling changes that make it in line with the OR design system from Remy,

ui/component/or-attribute-input/src/index.ts
```diff
@@ -438,19 +454,24 @@ export class OrAttributeInput extends subscribe(manager)(translate(i18next)(LitE
...
+                // so meta keeps the same reference and this check is O(1).
+                // A full asset reload produces a new meta reference, which is sufficient
+                // to detect a structural change without deep-comparing large meta values.
+                const structurallyEqual = oldAttr.name === attr.name
+                    && oldAttr.type === attr.type
+                    && oldAttr.meta === attr.meta;
+                if (structurallyEqual) {
                     // Compare value and timestamp
-                    if (!Util.objectsEqual(oldValue, attr.value) || oldTimestamp !== attr.timestamp) {
+                    if (oldValue !== attr.value || oldTimestamp !== attr.timestamp) {
```
MartinaeyNL, line 464: The impact if this isn't 100% backwards compatible is quite large.
You've probably tested this extensively, but maybe worth a double check.

ui/component/or-attribute-input/src/index.ts
```diff
@@ -325,11 +330,22 @@ export class OrAttributeInput extends subscribe(manager)(translate(i18next)(LitE
...
+            /* Align the send button with the input box below the field label. */
+            #wrapper:has(or-vaadin-input[label]) #send-btn {
+                margin-top: var(--or-field-input-offset);
+            }
+
+            or-vaadin-input {
+                /* Fixed input-area height enables the O(1) sizing path in or-vaadin-text-area.
+                   The host height stays content-driven so label/helper/error rows aren't clipped. */
+                --or-text-area-height: 96px;
+            }
```
MartinaeyNL: If you'd apply the refactor of this CSS variable to `--or-vaadin-text-area-height`, I would just apply this to `:host {}`, because it should apply to everywhere within this component anyhow.

ui/component/or-attribute-input/src/renderers/replay-data.ts
```diff
@@ -0,0 +1,163 @@
...
+    mapStateToControlProps,
+    mapDispatchToControlProps,
+    schemaMatches,
+} from "@openremote/or-json-forms";
+import "@openremote/or-vaadin-components/or-vaadin-text-area";
+
+interface ReplayDatapoint {
+    timestamp: number;
+    value: any;
+}
```
MartinaeyNL: Don't we have a model for this?

ui/component/or-attribute-input/src/renderers/replay-data.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (!props.required && props.path) {
+        deleteHandler = () => {
+            props.handleChange(props.path, undefined);
+        };
+    }
+
+    return getTemplateWrapper(html`
+        <or-vaadin-text-area
+            label="${props.label}"
+            helper-text="Format: seconds, value — one entry per line"
```
MartinaeyNL: Untranslated helper text

ui/component/or-attribute-input/src/renderers/replay-data.ts
```diff
@@ -0,0 +1,163 @@
...
+            style="width: 100%;"
+            .manualValidation="${true}"
```
MartinaeyNL: I don't remember it exactly, but was it possible to use `manualValidation` attribute instead of the `.manualValidation="${true}"` property? Or was the manual validation property only?

ui/component/or-attribute-input/src/renderers/replay-data.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    const onChanged = (event: Event) => {
+        const textArea = event.target as HTMLInputElement & { invalid: boolean; errorMessage: string };
+        const { datapoints, invalidLine } = textToDatapoints(textArea.value);
+        if (invalidLine !== null) {
+            // Don't commit: the data stays untouched and, since no state changes,
+            // the textarea keeps the user's draft instead of dropping bad lines.
+            // Requires manualValidation, otherwise Vaadin's focusout auto-validation
+            // (which runs after this change handler) resets invalid to false.
+            textArea.errorMessage = `Invalid entry on line ${invalidLine}, expected: seconds, value`;
```
MartinaeyNL: Untranslated message. (and there might be more that I didn't spot yet 😂)

ui/component/or-attribute-input/src/renderers/replay-data.ts
```diff
@@ -0,0 +1,163 @@
...
+            } else if (text.slice(lineStart, lineEnd).trim() !== "") {
+                return { datapoints, invalidLine: line };
+            }
+        }
+
+        lineStart = lineEnd + 1;
+    }
+
+    return { datapoints, invalidLine: null };
+}
```
MartinaeyNL: Maybe thinking outside of the box, but isn't there a simple tool / JS library for these kind of formats?
Or something we can include in `Util` ? (maybe out of scope)

I don't have a problem with this code, just a thought.

MartinaeyNL (commented):
I've read through the code an left some comments. Please take a look!
Let me know when something is unclear 😉

ui/component/or-vaadin-components/src/or-vaadin-text-area.ts
Ekhorn: will do a quick manual test, but it shouldn't break anything.

ui/component/or-vaadin-components/src/or-vaadin-text-area.ts
Ekhorn: documented

ui/component/or-vaadin-components/src/or-vaadin-text-area.ts
Ekhorn: Will disable it on mobile.

ui/component/or-attribute-input/src/renderers/replay-data.ts
Ekhorn: Had a look — the catch is that the format isn't real CSV: values are JSON-typed and can contain commas (5, {"a": 1, "b": 2}), so CSV parsers would mis-split them, and their quoting rules clash with JSON strings. A library would only replace the line/comma splitting while the JSON coercion, validation, and invalid-line reporting stay custom, at the cost of a new dependency (we currently have no CSV lib in the workspace). A shared Util helper makes sense the moment a second consumer of this format appears.

ui/component/or-vaadin-components/src/or-vaadin-text-area.ts
Ekhorn: I've decided to call it "manual resize" and disable it by default. Instead of the auto resize feature from Vaadin "manual resize" will require the user to manually resize the textarea with the handle. See [4fa8bc1](https://github.com/openremote/openremote/pull/3027/commits/4fa8bc1fe1d687263d48faeb07ace30db356ee31)

ui/component/or-vaadin-components/src/or-vaadin-text-area.ts
Ekhorn: See [a578865](https://github.com/openremote/openremote/pull/3027/commits/a578865bf5095d344cd0aae4927cef8054e6183e)

ui/component/or-attribute-input/src/index.ts
Ekhorn: I've tested a couple different things

1. live value ticks in the asset viewer
2. edit mode: change a value, dirty state + save (behaves like before where re-entering the previous value is still considered new - need to fix that separately)
3. a rules-editor attribute action write
4. editing a complex configuration item (agent link/schedule)

MartinaeyNL (approved):

PR #3060, opened 2026-07-19
fix: prevent chart tooltip cutoff in dashboard widgets

### LINKED ISSUE #2395: Bar chart widget tooltip cutoff
## User Story
The tooltip of bar chart widget is cutoff/incorrectly placed when it is pretty wide
Image

## Expected behavior
Text cropping , better placement or overflow outside of container such that the full tooltip is always readable

amanjain57-gif:
Fixes #2395
  
  The bar chart and line chart tooltips were being clipped by the widget container's overflow boundaries. This occurred
  because `confine: true` restricts the tooltip to the ECharts container element, which itself is constrained by the
  dashboard widget's `overflow: hidden` CSS.

  Replace `confine: true` with `appendToBody: true` so the tooltip renders as a direct child of the document body,
  outside the Shadow DOM and widget container clipping boundaries. Also add `extraCssText` for max-width and word-wrap
  to handle long tooltip content gracefully.

  Applied to both:
  - `or-attribute-barchart` (bar chart widget)
  - `or-chart` (line chart widget)

  ## Checklist

  Items 2 and 4 are left unchecked because this is a CSS/config fix with no new tests or docs needed — that's normal for
  this type of change. Paste that, submit, and let me know once done.

### DIFF
--- a/ui/component/or-attribute-barchart/src/index.ts
+++ b/ui/component/or-attribute-barchart/src/index.ts
@@ -1446,7 +1446,9 @@ export class OrAttributeBarChart extends LitElement {
       backgroundColor: this._style.getPropertyValue("--internal-or-asset-tree-background-color"),
       tooltip: {
         trigger: "axis",
-        confine: true, // make tooltip not go outside frame bounds
+        appendTo: 'body',
+        confine: true,
+        extraCssText: "white-space: normal; word-wrap: break-word;",
         axisPointer: {
           type: "cross",
           label: {
--- a/ui/component/or-attribute-barchart/src/index.ts
+++ b/ui/component/or-attribute-barchart/src/index.ts
@@ -1446,7 +1446,7 @@ export class OrAttributeBarChart extends LitElement {
       backgroundColor: this._style.getPropertyValue("--internal-or-asset-tree-background-color"),
       tooltip: {
         trigger: "axis",
-        appendTo: 'body',
+        appendTo: "body",
         confine: true,
         extraCssText: "white-space: normal; word-wrap: break-word;",
         axisPointer: {
--- a/ui/component/or-attribute-history/src/index.ts
+++ b/ui/component/or-attribute-history/src/index.ts
@@ -530,7 +530,9 @@ export class OrAttributeHistory extends translate(i18next)(LitElement) {
           backgroundColor: this._style.getPropertyValue("--internal-or-asset-viewer-panel-color"),
           tooltip: {
             trigger: "axis",
+            appendTo: 'body',
             confine: true,
+            extraCssText: "white-space: normal; word-wrap: break-word;",
             axisPointer: { type: "cross" },
             formatter: (params: any) => {
               if (Array.isArray(params) && params.length > 0) {
--- a/ui/component/or-attribute-history/src/index.ts
+++ b/ui/component/or-attribute-history/src/index.ts
@@ -530,7 +530,7 @@ export class OrAttributeHistory extends translate(i18next)(LitElement) {
           backgroundColor: this._style.getPropertyValue("--internal-or-asset-viewer-panel-color"),
           tooltip: {
             trigger: "axis",
-            appendTo: 'body',
+            appendTo: "body",
             confine: true,
             extraCssText: "white-space: normal; word-wrap: break-word;",
             axisPointer: { type: "cross" },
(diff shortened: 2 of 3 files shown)


ui/component/or-attribute-barchart/src/index.ts
```diff
@@ -1187,7 +1187,8 @@ export class OrAttributeBarChart extends LitElement {
             backgroundColor: this._style.getPropertyValue("--internal-or-asset-tree-background-color"),
             tooltip: {
                 trigger: "axis",
-                confine: true, //make tooltip not go outside frame bounds
+                appendToBody: true,
```
MartinaeyNL: According to the ECharts documentation, `appendToBody` is deprecated.
Use `appendTo` instead.

https://echarts.apache.org/en/option.html#tooltip.appendToBody

MartinaeyNL: I found out that appending the tooltip to the body, can also lead to other issues.
For example, the tooltip will fall outside of the screen, as shown in the screenshot below.
Along with that, it doesn't flip to a "right hand aligned" tooltip when it's close to the left side of the screen.

Image

ui/component/or-attribute-barchart/src/index.ts
MartinaeyNL, line 1: The same fix of `<or-attribute-barchart>` and `<or-chart>`, can be applied to `<or-attribute-history>` as well 👍

MartinaeyNL (changes requested):
Thanks for opening the pull request! 🥳 

I've left a few comments while I was testing out your changes on my machine.
Feel free to read them, and ask if further clarification is needed.

amanjain57-gif:
Thanks for the review! Addressed all three points:
1. Replaced deprecated `appendToBody` with `appendTo: 'body'`
2. Added `confine: true` alongside it to prevent the tooltip from overflowing outside the viewport
3. Applied the same fix to `<or-attribute-history>`

ui/component/or-attribute-barchart/src/index.ts
```diff
@@ -1187,7 +1187,9 @@ export class OrAttributeBarChart extends LitElement {
             backgroundColor: this._style.getPropertyValue("--internal-or-asset-tree-background-color"),
             tooltip: {
                 trigger: "axis",
-                confine: true, //make tooltip not go outside frame bounds
+                appendTo: 'body',
+                confine: true,
+                extraCssText: "max-width: 300px; white-space: normal; word-wrap: break-word;",
```
MartinaeyNL: I think we can get rid of the `max-width` property, right?
In my testing, when the tooltip goes off the screen, thanks to the white-space and word-wrap properties, they already wrap correctly.
Removing it also helps to read longer asset and/or attribute names better, when the screen size allows it.

MartinaeyNL (commented):
Thanks for addressing the comments!
I've only got one small remark on the `max-width` property.

Besides that, can you check the CLA error that appeared after your last commit?
I think the commit was accidentally authored without an attached GitHub account.
Should be simple to fix 😉 

I'm happy to approve once that is resolved ✅ 👍

amanjain57-gif:
Addressed:
- Removed `max-width` from `extraCssText` in all three components — the white-space and word-wrap properties handle wrapping on their own.
- Fixed the CLA issue — previous commit was authored with the wrong email. Force-pushed with the correct GitHub-linked email.

amanjain57-gif:
Rebased on latest master to resolve merge conflicts in all three files. The tooltip fix is now cleanly applied on top of the current upstream:

- `or-attribute-barchart`: added `appendTo: 'body'`, kept `confine: true`, added `extraCssText` for word-wrap (no max-width)
- `or-chart`: same approach
- `or-attribute-history`: same approach

CLA should also be clean now — single commit with correct author email.

ui/component/or-attribute-barchart/src/index.ts
amanjain57-gif: Done — switched to `appendTo: 'body'` as recommended.

amanjain57-gif: Addressed — added `confine: true` alongside `appendTo: 'body'` so the tooltip stays within viewport bounds while still rendering outside the Shadow DOM.

ui/component/or-attribute-barchart/src/index.ts
amanjain57-gif: Done — applied the same fix to `<or-attribute-history>` as well.

ui/component/or-attribute-barchart/src/index.ts
amanjain57-gif: Good call — removed `max-width` entirely. The `white-space: normal` and `word-wrap: break-word` handle wrapping on their own.

amanjain57-gif:
@MartinaeyNL All feedback addressed and CLA is now green — ready for re-review when you get a chance. Thanks!

MartinaeyNL (approved):
Thanks!

PR #3077, opened 2026-07-21
fix: Change timeout Maven artifacts to 40 minutes

denniskuijs:


denniskuijs:
Included fix in PR #3079

PR #3078, opened 2026-07-21
Bugfix for attribute value filter not working in asset tree

### LINKED ISSUE #2575: Asset tree filter by value not working
## Bug description
This afternoon @wborn reported an issue with the "attribute value" filter not working on his deployment.
I did a bit of digging, and came to the same conclusion.

Applying a filter like this, which should return 0 results, caused the list of assets to include all Light Assets with the "notes" attribute, without comparing value.

Image

Image

## Reproduce steps
1. Go to the assets page on any environment
2. Click on the filter icon, select an asset type, and fill in an required attribute of that asset type.
3. Type any value into the "attribute value" field.
4. Click on "FILTER" and you'll see the same result.

## Expected behavior
The list of assets should only show the assets that are in line with the filter.
So if a user looks for assets with the "notes" attribute set on "my special note", it should only return those.
When using other types, like `number` or `boolean`, it should behave the same.
Other complex expressions like `>25` should also work again.

## Bug investigation details
Important to note, is that since the introduction of the `/assets/tree` endpoint, no attributes are included in the payload anymore. The `AssetTreeAsset` shows only a limited amount of fields are included nowadays.
This causes the filtering of attribute values, which occurs in the frontend, not being able to compare the values.

Since the new asset tree pagination only returns a limited number of assets, say 100 out of 500.000, we can't locally compare the attribute values anymore. Because there is no guarantee that all assets are retrieved through the HTTP API. This means that the new implementation needs to process attribute values on database level.

It's important to investigate whether complex expressions like `>25` still work on the database level,
which I assume is not the case.

## Implementation notes
- As explained above, filtering by attribute value is processed on the frontend. As asset tree pagination got introduced, we need to move this filtering to the database level.
- We need to investigate if complex expressions like `attribute value > 25` works on the database level through the `AssetQuery` object.

For additional context, please contact @MartinaeyNL.

MartinaeyNL:
## Description
This PR fixes #2575 where the value filter was not working at all.
It was a left-over issue after the Asset tree scaling got introduced with dynamic HTTP fetching.
I've now resolved the issue, and added test coverage for it 👍 

## Changelog
- Add attribute value filter to the fetching function in `<or-asset-tree>`.
It now correctly filters out any assets that don't align with them.
- Add frontend testing coverage for this bug, alongside other bugs I encountered in the process.
- Minor cleanup of whitespaces 😅 

## Checklist


MartinaeyNL:
Will write a UI test to provide code coverage for this PR once I find the time.

MartinaeyNL:
This pull request is ready for review! 👍

ui/app/manager/test/assettree.test.ts
```diff
@@ -458,6 +441,100 @@ test(`Searching for an asset and removing it keeps the tree and viewer in tact`,
   await expect(assetTree.getAssetNodes()).toHaveCount(0); // Nothing is visible anymore, since there is nothing matching the "Battery 10" text filter.
 });
 
+/**
+ * @given 4 assets are created in the "smartcity" realm
+ * @and the assets are visible in the tree (a total of 4)
+ * @when the user applies a combination of name, type and value in the filtering menu
+ * @then the asset tree should show the assets that comply with those requirements.
+ */
+test(`Should update asset list correctly when applying filters`, async ({ manager, assetsPage, assetTree }) => {
```
Ekhorn, line 450: If it is possible, it would be good test numbers and related operators. I tested greater and less than, but not sure I was doing it correctly.

Ekhorn (changes requested):
Also found a bug where the tree is loading continuously and calling the count endpoint continuously. Please ensure the test also checks after clicking the "clear" that shows the asset correctly.

ui/app/manager/test/assettree.test.ts
MartinaeyNL: Good point, I didn't think of that when writing the tests. I'll add those scenarios

PR #3102, opened 2026-07-24
Replace LitElement with OrElement

Ekhorn:
## Description

This migrates the remaining components that weren't yet using the base `OrElement` component to apply global styles.

## Checklist


Ekhorn:
@MartinaeyNL we should discuss whether we actually want to commit to this considering some alternatives:

1. A mixin -> research whether there are benefits
2. A custom header component
3. Only apply this solution to those components that use header elements (and maybe for other use cases.) Reasons being: Reduces unnecessary inheritance and unused styles. Would require clear documentation about this in the `AGENTS.md`.

Ekhorn:
Discussed that this base component allows inheritance to carry over the styles to extending classes. Though we can't directly control the Vaadin components and make them extend OrElement, which is fine because we can use their themable mixin instead to inject the styles (no need to introduce our own).

Future use cases for this component would possibly be replacing the translation mixin or sharing DOM utilities.

MartinaeyNL (approved):

Issue #3034, opened 2026-07-15
Make send/submit button part of the relevant inputs themselves


Ekhorn:
### Description

Currently some fields in the `or-attribute-input` have a send button alongside it. This button is not always present for every field type. We should discuss which field types should have this button and consider moving it to the relevant components as a decoration option.

What prompted me to write this issue is that it is currently not easy to align the send button with the field alongside it (without a fragile solution). Centering- or putting the button at the bottom looks weird. Aligning it from the top requires you to take into account the label height and other spacing/margins/paddings.

### Acceptance criteria

- [ ] The fields that may have a submit button have been discussed.
- [ ] Use slots to decorate the `or-vaadin-input`.
- [ ] ~~The `--or-field-input-offset` is removed and the styles that use it.~~
- [ ] ~~The components have a property `send`/`submit` that shows this button, which emits a submit event.~~

Ekhorn:
We might want an option to just pass a decorator button to a field and allow positioning to be configured as well.

ebariaux:
Decided to use slots instead of JS (OK for @MartinaeyNL and @richturner)

wborn:
Another reason to reconsider the send/submit buttons is that they suggest the attribute value is only updated after clicking the button, while this is not actually how the Asset view behaves.

For example, changing a timestamp or boolean value immediately updates the attribute. The adjacent send button therefore gives the misleading impression that there is still a separate submit step.

I think it would make sense to remove the send/submit button for inputs where changing the input already updates the attribute immediately. We already effectively do this for text attributes with allowed values, where only the dropdown is shown and selecting a value updates the attribute directly.

The button probably only really makes sense for free text input, where the value otherwise only gets submitted by pressing Enter, which is not particularly obvious or intuitive to users. In that case, having an explicit submit action provides useful affordance.

This would make the behavior more consistent and avoid suggesting that changes are buffered until explicitly submitted.

PR #3129, opened 2026-07-29
Implement datetime field in vaadin input

### LINKED ISSUE #2737: Timestamp ISO8601 attributes do not show values and cannot be modified
Whenever you have an attribute of type "Timestamp ISO8601" it does not show the value in the asset viewer.

You can also not modify the attribute in the asset viewer.
If you try to modify it there is this validation warnings gets logged by the backend:

`2026-05-07 11:47:03.412  WARNING [WebService task-7             ] remote.manager.asset.AssetStorageService : Asset merge failed as asset has failed constraint validation: asset=ThingAsset{id='5tR7zqtfhxvhEX4M8zXnAD', name='Thing 1', type ='ThingAsset', parentId='null', realm='master'}, exception=: Asset is not valid, attributes[timestamp].value: must match "^(?:[1-9]\d{3}-(?:(?:0[1-9]|1[0-2])-(?:0[1-9]|1\d|2[0-8])|(?:0[13-9]|1[0-2])-(?:29|30)|(?:0[13578]|1[02])-31)|(?:[1-9]\d(?:0[48]|[2468][048]|[13579][26])|(?:[2468][048]|[13579][26])00)-02-29)T(?:[01]\d|2[0-3]):[0-5]\d:[0-5]\d(?:[\.,]\d{1,9})?(?:Z|[+-][01]\d:[0-5]\d)$"`

It does show the most recent timestamp value when clicking such a thing on the map.
The asset viewer does show historic timestamp values.

**To Reproduce**

1. Create a thing with "timestamp" attribute of type "Timestamp ISO8601"
2. Add the "Store datapoints" configuration item to the attribute.
3. Configure the location attribute so it show up on the map.

4. Add a Groovy rule like the one below to update the attribute (update the asset ID):

```groovy
package demo.rules

import java.time.OffsetDateTime
import org.openremote.manager.rules.RulesBuilder
import org.openremote.model.asset.Asset
import org.openremote.model.asset.impl.*
import org.openremote.model.rules.Assets
import org.openremote.model.rules.Notifications

Assets assets = binding.assets

rules.add()
        .name("Test rule")
        .when({
    facts -> true
})
        .then({
    facts ->
        assets.dispatch("5tR7zqtfhxvhEX4M8zXnAD", "timestamp", OffsetDateTime.now().toString())
})
```

5. Let the rule run for a minute so there is some data and disable the rule again
6. Now check the "timestamp" attribute value and try to update it

**Expected behavior**

The asset viewer value shows the most recent value and it is possible to update the value.

**Example**

Image

Ekhorn:
## Description

Adds the date and time picker to `or-vaadin-input`, so date and time attributes use Vaadin instead of `or-mwc-input`.

- Closes #2737

## Changelog

- `timestamp`, `timestampISO8601`, `dateAndTime` and number attributes with `asDate` use `or-vaadin-date-time-picker`. Edited values are saved in the attribute's own type: a number, or an ISO 8601 string for `timestampISO8601`, which previously couldn't be saved. Clearing saves `null`, and a value that isn't a date shows an empty picker instead of breaking the input.
- The picker's precision follows the attribute's value format (seconds or milliseconds), and defaults to minutes.
- The date, time and date time pickers format dates and times in the browser locale, and show month and weekday names and button labels in the app language.
- The asset viewer's "Created on" field shows its date and time.
- Component tests cover every date and time value type, several browser locales, and "Created on".
- For manual testing, a `datetime` setup type (`OR_SETUP_TYPE=datetime ./gradlew setup:run`) creates an asset with one attribute per case, each labelled with what the picker should do.

## Checklist

### DIFF
--- a/ui/component/or-vaadin-components/src/value-input-provider.ts
+++ b/ui/component/or-vaadin-components/src/value-input-provider.ts
@@ -55,6 +55,7 @@ import {
   type SupportedWellknownValueTypes,
 } from "./util";
 import { OrVaadinInput } from "./or-vaadin-input";
+import { OrVaadinDateTimePicker } from "./or-vaadin-date-time-picker";
 
 export interface ValueInputProviderOptions {
   label?: string;
@@ -361,6 +362,13 @@ export const getValueHolderInputTemplateProvider: ValueInputProviderGenerator = 
     }
   }
 
+  // or-vaadin-date-time-picker works with local ISO strings, while the value is a timestamp
+  if (inputType === InputType.DATETIME) {
+    min = min === undefined ? undefined : OrVaadinDateTimePicker.getLocalizedISOString(new Date(min));
+    max = max === undefined ? undefined : OrVaadinDateTimePicker.getLocalizedISOString(new Date(max));
+    valueConverter = (v) => (v ? Date.parse(v) : null);
+  }
+
   if (inputType === InputType.NUMBER && format?.resolution) {
     step = format.resolution;
   }
@@ -389,7 +397,7 @@ export const getValueHolderInputTemplateProvider: ValueInputProviderGenerator = 
       if (elem?.checkValidity()) {
         const doRemoteUpdate =
           !OrVaadinInput.CHANGE_EVENTS.has(inputType) || OrVaadinInput.CHANGE_EVENTS.get(inputType) === ev.type;
-        valueChangeNotifier(valueConverter?.(elem.nativeValue) ?? elem.nativeValue, doRemoteUpdate);
+        valueChangeNotifier(valueConverter ? valueConverter(elem.nativeValue) : elem.nativeValue, doRemoteUpdate);
       }
     };
 
@@ -399,6 +407,8 @@ export const getValueHolderInputTemplateProvider: ValueInputProviderGenerator = 
       if (inputType === InputType.CHECKBOX || inputType === InputType.SWITCH) {
         checked = Boolean(value);
         value = undefined;
+      } else if (inputType === InputType.DATETIME && value !== undefined && value !== null) {
+        value = OrVaadinDateTimePicker.getLocalizedISOString(new Date(value));
       }
 
       return html`
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -63,7 +63,7 @@ Pick the approach by how much of the base component you need to keep. The base +
 
 #### Registering an input type in the input pipeline
 
-To make an input type usable through the input pipeline, register it in `or-vaadin-input.ts` (`TEMPLATES` map + a `getXTemplate`, and `nativeValue` if it is a boolean that exposes `checked` instead of `value`) and update `value-input-provider.ts` (boolean/checked types use the `checked` attribute, not `value`).
+To make an input type usable through the input pipeline, register it in `or-vaadin-input.ts` (`TEMPLATES` map + a `getXTemplate`, and `nativeValue` if it is a boolean that exposes `checked` instead of `value`) and update `value-input-provider.ts` (boolean/checked types use the `checked` attribute, not `value`). Where the Vaadin component represents the value differently than the attribute does, such as the date time picker taking a local ISO string instead of a timestamp, convert the value and any min/max there and convert back through `valueConverter`.
 
 ### Storybook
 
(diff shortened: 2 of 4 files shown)


ui/component/or-vaadin-components/src/value-input-provider.ts
```diff
@@ -361,6 +362,20 @@ export const getValueHolderInputTemplateProvider: ValueInputProviderGenerator =
     }
   }
 
+  // or-vaadin-date-time-picker works with local ISO strings, while the value is a timestamp or an ISO 8601 string
+  if (inputType === InputType.DATETIME) {
+    step = OrVaadinDateTimePicker.getStep(format);
+    min = min === undefined ? undefined : OrVaadinDateTimePicker.getLocalizedISOString(new Date(min), step);
```
wborn: Minimum bounds are truncated down to the picker precision here. For example, with a `Future` constraint at 10:30:45 and the default minute precision, `min` becomes `10:30`; the picker accepts that value, but it is saved as 10:30:00 and `ValueConstraint.Future` rejects it as no longer being in the future. The same applies to `Min` values that are not aligned to the picker step.

Please round minimums up to the next representable picker value and add a regression test for the constraint behavior. Maximum bounds can safely continue to round down.

ui/component/or-vaadin-components/src/date-time-i18n.ts
```diff
@@ -0,0 +1,240 @@
...
+  const pad = (value: number | string, length = 2) => String(value).padStart(length, "0");
+  const hours = Number(time.hours);
+  let text = `${hours % 12 || 12}:${pad(time.minutes)}`;
+  if (time.seconds !== undefined) {
+    text += `:${pad(time.seconds)}`;
+  }
+  if (time.milliseconds !== undefined) {
+    text += `.${pad(time.milliseconds, 3)}`;
+  }
+  return `${text} ${hours < 12 ? am : pm}`;
```
wborn: `formatTime()` always puts the day period after the time, and `parseTime()` only accepts that layout. This does not match the browser locale for locales such as `ko-KR` (`오전 10:30`) or `zh-TW` (`上午10:30`), so the picker does not consistently provide the browser-locale formatting/parsing described by this change.

Please derive the ordering and literals from `Intl.DateTimeFormat.formatToParts()` (or an equivalent locale-aware mechanism) and add a test for a locale where the day period precedes the time.

wborn (changes requested):

Two functional issues remain in the current implementation and should be addressed before merge. The existing Copilot findings have been resolved in the current revision, and the current CI checks pass.

Ekhorn:
1 thing that you'll run into when the constraints require a round up. The options in the time picker all get shifted, while you might expect either the first option to roundup or for it to not show at all. This is a known issue, see `https://github.com/vaadin/web-components/issues/12621`.

image
image

wborn (approved):
The two functional issues from the previous review have been addressed in the current revision. Minimum bounds are now rounded up to the picker precision with regression coverage, and the 12-hour locale handling now derives its ordering and literals from `Intl.DateTimeFormat.formatToParts()`, including coverage for locales where the day period precedes the time.

The complete current change was re-reviewed and no further blocking issues were identified. The build and test checks for the current HEAD pass, including the UI component and UI app tests.

*This review was AI-assisted.*

wborn:
I still need to do a bit of testing.

MartinaeyNL:
I'm happy to do a review as well, taking a look at the UI changes 😉

wborn:
I gave it a test and it seems to work fine now. :+1: 

It's a bit confusing that a submit button is shown next to the control on the asset view page for which I added https://github.com/openremote/openremote/issues/3034#issuecomment-5793726680.

ui/component/or-vaadin-components/test/fixtures/value-input-provider-harness.js
MartinaeyNL: This feels overcomplicated

ui/component/or-vaadin-components/src/date-time-i18n.ts
MartinaeyNL, line 1: After deep diving for about an hour into this PR, I'm still struggling with this file.
On one hand it's a useful file, although it feels a bit bloated and difficult to maintain.
Because it feels like this file is reinventing the wheel what we're doing with Moment everywhere else.

I understand the goal here, but I'm not really sure what the solution is.

MartinaeyNL (changes requested):
Overall I think the code is OK-ish, although I still seem to struggle with the code complexity of these changes.
We're also adding a lot of cross-component dependencies, making the date picker less independent / isolated.
Not sure what to do about it really..

Also see the comment below.

PR #3146, opened 2026-08-03
Convert Java JUnit unit tests to Groovy Spock specs

### LINKED ISSUE #2752: Convert JUnit java test to Spock
We standardize on using Spock for all our tests.
Some tests have been written as pure JUnit java test, migrate those to Spock unit tests for consistency.

ebariaux:


container/src/test/groovy/org/openremote/container/web/WebServiceEntitySizeConfigTest.groovy
```diff
@@ -0,0 +1,63 @@
...
+    @Unroll
+    def "uses #scenario entity size limits"() {
+        given:
+        def config = [
+            (WebService.OR_WEBSERVER_LISTEN_HOST): "127.0.0.1",
+            (WebService.OR_WEBSERVER_LISTEN_PORT): "0"
+        ] + overrides
+        def container = Stub(Container) {
+            getConfig() >> config
+            getService(_) >> null
```
wborn: The previous Java test intentionally returned `null` only for `IdentityService` and threw for every other `getService(...)` call. `getService(_) >> null` removes that guard, so a new service dependency added to `WebService.init()` could be silently accepted by this test.

Could the stub preserve the previous behavior by returning `null` for `IdentityService` and throwing for other service types?

model/src/test/groovy/org/openremote/model/util/JSONSchemaUtilTest.groovy
```diff
@@ -16,55 +16,46 @@
...
-  static void setup() {
-    ValueUtil.doInitialise();
+  private static final Reflections reflections = new Reflections("org.openremote")
+
+  def setupSpec() {
+    ValueUtil.doInitialise()
   }
 
   private static InputStream loadResourceAsStream(String resourcePath) {
     return JSONSchemaUtilTest.class.getResourceAsStream(
```
wborn: Could this conversion use more idiomatic Groovy where possible? There are still quite a few Java-style constructs left over from the original test, such as semicolons, explicit local variable types, `return`, and `.class` where Groovy doesn't require them.

Since this is being converted to Spock/Groovy anyway, simplifying those would make it more consistent with the existing Spock tests.

model/src/test/groovy/org/openremote/model/util/LockByKeyTest.groovy
```diff
@@ -0,0 +1,159 @@
...
+
+        int getQueueSize(String key) {
+            LockWrapper lockWrapper = locks.get(key)
+            if (lockWrapper != null) {
+                return lockWrapper.numberOfThreadsInQueue.get()
+            }
+            0
+        }
+    }
+
```
wborn, line 65: Could this test be structured like the existing Spock specs, with a descriptive feature name and `given`/`when`/`then` blocks rather than putting the complete procedural test under `expect:`?

For example, `def "threads waiting for the same key are not starved"()` would also make the reported test name consistent with the other converted specs.

wborn (commented):
This is an AI review before manual review.

Overall, the JUnit-to-Spock migration looks good. The converted tests generally preserve the existing behavior while making several tests more concise and data-driven, and most of the new specs follow the conventions already used by OpenRemote's existing Spock tests.

One substantive issue was found where the converted `WebServiceEntitySizeConfigTest` no longer rejects unexpected service lookups. Two additional non-blocking comments concern making the converted tests more idiomatic and consistent with the existing Spock suite.

No other issues were found in the current changes.

container/src/test/groovy/org/openremote/container/web/WebServiceEntitySizeConfigTest.groovy
ebariaux: This should be fixed by the changes from @Ekhorn

model/src/test/groovy/org/openremote/model/util/LockByKeyTest.groovy
ebariaux: Rewrite from @Ekhorn aligns more with this but I don't want to diverge too much from the initial Java test because the race condition it tests are pretty specific and initial creation of this test required quite a bit of time and back and forth

model/src/test/groovy/org/openremote/model/util/JSONSchemaUtilTest.groovy
ebariaux: Made changes in that direction but kept some statically typed variable required for the test being conducted or because it made code clearer

wborn (commented):
AI re-reviewed the current changes and found no further issues. The previous review comments have been addressed.

A human maintainer review is still needed.

wborn (approved):
Thanks!

PR #3218, opened 2026-08-27
Option to enable OpenTelemetry agent on manager JVM

### LINKED ISSUE #3207: Add OTEL Agent
We want to expose traces via OpenTelemetry for observability.
Add the OTEL Agent to the JVM (potentially a configuration option).
No code change at this stage, we'll only work with the "default" traces provided e.g. by JDBC and Undertow.

ebariaux:
See README.md for some of the validation steps.
Some extra setup / test information has been kept in that file as I thought it made sense to keep it as background info.


README.md
MartinaeyNL, line 1: I would move these `README.md` changes to the documentation website itself, wherever suitable.

ebariaux: I added some documentation, see https://github.com/openremote/documentation/pull/193
But I think it makes sense to keep more details here and have the public doc more a "How to" guide.
I'll review/improve the public doc as we're progressing with using all the "Grafana" features on the different projects and moving to a more stable situation.

MartinaeyNL: I think it makes sense to mention that we're using OpenTelemetry etc in the `README.md`.
However, once we're explaining **how** it works, I'd rather hyperlink it to the docs.
(similar to all other tools/agents we're using)

MartinaeyNL: I also noticed most of the "technical information" of the repo isn't inside the `README.md` file,
but on the docs introduction page; https://docs.openremote.io/docs/introduction

MartinaeyNL (commented):
See my reply on the comment above regarding the `README.md` file changes.
Personally I'd completely leave those changes out, and move them to the documentation repository.

Other that the README, I'm happy with the changes 👍

README.md
ebariaux: I remove extra information from README file in project, pointed to documentation website and added what I thought was still useful to the documentation page.

manager/build.gradle
```diff
@@ -14,6 +14,15 @@ application {
   mainClass = "org.openremote.manager.Main"
 }
 
+configurations {
+  demoExtension
```
wborn, line 18: Looks like a merge conflict that was not properly resolved because the `demoExtension` configuration was removed here in https://github.com/openremote/openremote/pull/3271.

```suggestion
```

gradle/libs.versions.toml
```diff
@@ -55,6 +55,7 @@ niftyModbus = "1.4.0"
 nimbusJoseJwt = "10.9.1"
 objenesis = "3.5"
 moquette = "0.18.0"
+opentelemetryJavaagent = "2.30.0"
```
wborn, line 58: Any reasons for not using [2.31.1](https://central.sonatype.com/artifact/io.opentelemetry.javaagent/opentelemetry-javaagent/2.31.1)?

MartinaeyNL (approved):
Let's merge this. We can always bump the OpenTelemetry version later 👍
(I also assume dependabot will detect it haha)

Issue #2701, opened 2026-04-20
Update Jackson v2 to v3


Ekhorn:
### Description

The upgrade to Jackson 3 contains quite some breaking changes (ref https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md).

This is "required" for #2686. 

### Context

I made this issue so people are aware I already partially started on the migration in #2686

PR #3220, opened 2026-08-27
Migrate JSONSchemaUtil snapshot tests to spock snapshot

Ekhorn:
## Description

Migrates `JSONSchemaUtilTest` to the snapshot testing added in [Spock 2.4](https://spockframework.org/spock/docs/2.4/extensions.html#snapshot-testing). The expected schemas moved out of the spec into snapshot files, so each feature is now a single `schemaMatchesSnapshot(SomeType)` call.

This was done in preparation for #2701, to have a unified way of writing snapshot tests in Spock.

## Updating the snapshots

Run `./gradlew :model:test -PupdateSnapshots`, then review the resulting diff before committing. Without the flag a mismatch is reported as a normal test failure.

Snapshots live in `model/src/test/resources/snapshots/`, one file per feature, named after the feature with an iteration index for the data driven map test.

The snapshot root is set through the `spock.snapshots.rootPath` system property in `model/build.gradle` rather than a `SpockConfig.groovy`, because the `snapshots` config block is silently ignored on Spock 2.4-groovy-5.0. Running the spec straight from an IDE therefore requires Gradle delegation.

## Changelog

- Replace the inline expected JSON and the `loadResourceAsStream` expectation resources with Spock snapshots.
- Move the 7 existing expectation resources under `snapshots/` unchanged, so git records them as renames.
- Add `-PupdateSnapshots` to the `:model:test` task for rewriting snapshots.
- Drop the `jsonassert` dependency from `model` and the version catalog, since comparison is now exact rather than lenient.
- Assert 4 fields that the lenient comparison had been hiding in `schema remaps annotated types`, whose expected JSON was malformed (missing space between after `:` like `"key":value`) and got truncated by `readTree`.
- Document snapshot regeneration in `AGENTS.md`.

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,6 +16,8 @@ In sandboxed environments, add `--offline` (requires dependencies to already be 
 When running integration tests, part of the stack (PostgreSQL and Keycloak) must be running. Start it with `mkdir -pm 777 tmp && docker compose -f profile/dev-testing.yml -p openremote up -d --no-build`.  
 Running `./gradlew clean` deletes the root `tmp/` directory that is mounted into PostgreSQL (see `profile/dev-testing.yml`), so recreate it and restart the stack before running tests again.
 
+Spock snapshots are stored under `model/src/test/resources/snapshots`. Rewrite them with `./gradlew :model:test -PupdateSnapshots` and review the resulting diff, since a mismatch is only reported as a test failure otherwise.
+
 ## REST resources
 
 ### Endpoint roles
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -43,7 +43,6 @@ javaxAnnotationApi = "1.3.2"
 jaywayJsonPath = "2.10.0"
 jdom = "2.0.6.1"
 jserialcomm = "2.11.4"
-jsonassert = "1.5.3"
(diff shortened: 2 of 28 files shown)


MartinaeyNL (commented):
My AI buddy took a look, see below 😉

---

**Verified locally: on a Windows machine with `core.autocrlf=true` (this repo's global default here), all 24 snapshot comparisons fail with `value == snapshotValue` where both sides look identical.** Root cause:

- Snapshot files are committed as LF (`git ls-files --eol` → i/lf, w/crlf), but autocrlf converts them to CRLF in the working tree on Windows.
- Spock's `@Snapshot` extension normalizes the compared value to \n (the .actual dump Spock writes is LF-only), but reads the snapshot file raw — CRLF ≠ LF → failure on every feature.
- CI passes because Linux checkout keeps LF.

Confirmed: converting the working-tree files to LF makes it 26/26 pass, then `git checkout` restores CRLF and it's broken again.

Fix: add an EOL rule for the snapshot dir to `.gitattributes` (it currently only covers `*.sh`):
```
model/src/test/resources/** eol=lf
```
Without it, every Windows dev gets a red test suite after clone, and a "helpful" `-PupdateSnapshots` run just churns files.

PR #3239, opened 2026-08-28
Apply timezone offset for single occurrence events in SimulatorProtocol

### LINKED ISSUE #3226: The simulator timezone field only works for recurring schedules
**Describe the bug**

The timezone field only works for recurring schedules

**To Reproduce**

1. Add a SimulatorAgentLink to a number attribute
2. Add a (one-time) schedule starting the next day with replay data
3. Configure has predicted datapoints
4. Create a line chart widget with a default timeframe of a day
5. The predicted datapoints show up with UTC time in mind even when you specify any different timezone

**Expected behavior**

The predicted datapoints to show in the chart matching the schedule time with your specified timezone in mind.

Ekhorn:
## Description

- Closes #3226

## Checklist


MartinaeyNL (commented):
My AI buddy took a look 😉 

1. **Fall-back (overlap) can produce a negative delay** — `SimulatorProtocol.java:266`. During the repeated hour (e.g. agent restarts at 02:10 CET second pass in Amsterdam), `toEpochMillis` "takes the first of the two", so `nextRunUTC` can be before `nowUTC` → negative delay → executor fires immediately and the whole remaining overlap window replays at once (predicted points also get past timestamps). Only spring-forward is tested. Suggest: if (`nextRunUTC` <= `nowUTC`) `nextRunUTC` = <resolve with later offset at overlap> (e.g. `atZone(timezone).withLaterOffsetAtOverlap().toInstant()` when `ZoneOffset.getRules().getValidOffsets(local).size() > 1`).

2. **Recurrence start inside a spring-forward gap:** if an occurrence starts at a skipped local time (e.g. daily at 02:30), `getTimeUntilNextOccurrence`/`tryAdvanceActive` still use the raw local time, so intra-occurrence offsets for that day shift by an hour. Pre-existing edge, but now partially solvable with the helpers.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -262,13 +261,18 @@ protected ScheduledFuture<?> scheduleReplay(
       }
     }
 
+    // The next run is a distance between local times, which is an hour short or long across a
+    // daylight saving transition, so it is resolved to the instant it occurs on
+    long nextRunUTC = toEpochMillis(now + nextRun.getAsLong(), timezone);
```
wborn, line 266: The fall-back overlap concern raised in the existing review is valid, and the failure mode can be more severe than a single early replay.

During the second occurrence of an ambiguous local time, `toEpochMillis()` always resolves the target using the earlier offset. For example, in `Europe/Amsterdam`, if `nowUTC` is `01:10 UTC` during the second local `02:10`, scheduling a datapoint for local `02:30` resolves that target to the first `02:30` at `00:30 UTC`. This makes `delay` negative.

`ScheduledExecutorService` executes a negative delay immediately. The callback then invokes `scheduleReplay()` again while the local clock is still before `02:30`, so the same datapoint can be selected and immediately scheduled again. This can repeat until the second `02:30` is actually reached.

The overlap resolution therefore needs to consider `nowUTC`: when the earlier candidate has already passed but the later candidate is still in the future, the later offset should be selected. Please also add a test that starts during the second repeated hour of a fall-back transition.

agent/src/main/java/org/openremote/agent/protocol/simulator/SimulatorProtocol.java
```diff
@@ -295,17 +298,40 @@ protected ScheduledFuture<?> scheduleReplay(
...
+   * and a local time that happens twice takes the first of the two.
+   */
+  private static long toEpochMillis(long localMillis, ZoneId timezone) {
+    return LocalDateTime.ofInstant(Instant.ofEpochMilli(localMillis), ZoneOffset.UTC)
+        .atZone(timezone)
+        .toInstant()
+        .toEpochMilli();
+  }
+
   public List<ValueDatapoint<?>> calculatePredictedDatapoints(
```
wborn, line 328: This changes the public signatures of `calculatePredictedDatapoints(...)` and `Schedule.tryAdvanceActive(...)`. Since `openremote-agent` is published as a Maven library, could you check whether there are any downstream consumers of these methods before removing the old signatures? If they are used externally, compatibility overloads may be worth keeping.

wborn (changes requested):
*This is an additional AI-assisted review.*

The timezone handling is moving in the right direction and the added tests cover normal timezone offsets and a spring-forward transition well. The current CI run is also green.

One correctness issue remains around ambiguous local times during a fall-back transition. There is also a public API change worth checking for downstream consumers.

PR #3240, opened 2026-08-28
Fix bar chart navigation between dates

### LINKED ISSUE #3236: Bar chart widget does not correctly navigate between dates
## Bug description
I encounter multiple issues when navigating between dates in the bar chart of the Manager UI.
When the interval is different than the default, navigating changes the timeframe from 24hrs to 23hrs + X mins.
But this also seems to happen when navigating dates without changing the interval.

Along with that, the data ends up malformed within the bar chart, and sometimes invisible.
See the two examples below coming from a production instance of OpenRemote below.

## To Reproduce
**Warning:** I have not attempted to reproduce this using a fresh installation of OpenRemote.
However, these were the steps necessary for that project I encountered it on;
1. Go to the Insights page, and navigate to a dashboard with a bar chart widget.
2. Click on the 'next' arrow, navigating forwards, and after that navigate back a couple of times.

## Expected behavior
- Navigating back and forth should respect the original window, for example 24 hours.
- Navigating should not cause the data to be malformed and/or invisible in the chart.

## Screenshots / videos
Image

## Device information:
- OS: **Windows**
- Browser **Chromium**
- On version v1.26.1 of OpenRemote

## Additional context / developer info
I think it might have to do with the "correction" the frontend does when invalid dates are entered.
For example, when the timeframe is 24 hours, and navigating forwards calculates a timeframe of 24 hours and 1 minute, it resets.
Not sure whether it is related, but this bug is definitely worth looking into.

alexsmolya:
## Description

Fixes #3236 by keeping the requested time range as the authoritative chart axis and request bounds. Bucket-centred datapoints are no longer used to overwrite those boundaries, so 60-minute, 15-minute, and 5-minute intervals no longer shrink the visible range.

Navigation preserves the configured window semantics and keeps the focused regression coverage for repeated forward/backward movement, custom durations, sparse/partial data bounds, and DST-related cases.

### Navigation semantics

The component distinguishes fixed elapsed windows such as `24Hours` from calendar windows such as `Day`, `Month`, and `Year`. Fixed windows retain their elapsed duration when navigating, while calendar windows retain calendar semantics. Using one model for all windows would either make explicit 24-hour ranges vary across DST or make calendar-month navigation depend on the previous month’s elapsed length.

## Checklist

- [ ] 2. Tests are written and all tests pass (focused execution is unavailable in this environment because `yarn` is not installed)

### DIFF
--- a/ui/component/or-attribute-barchart/src/timeframe.ts
+++ b/ui/component/or-attribute-barchart/src/timeframe.ts
@@ -49,7 +49,7 @@ export function shiftTimeframe(
 
   const start = moment(currentStart);
   const end = moment(currentEnd);
-  direction === "previous" ? start.subtract(value, normalizedUnit) : start.add(value, normalizedUnit);
-  direction === "previous" ? end.subtract(value, normalizedUnit) : end.add(value, normalizedUnit);
+  direction === "previous" ? start.subtract(value, unit) : start.add(value, unit);
+  direction === "previous" ? end.subtract(value, unit) : end.add(value, unit);
   return [start.toDate(), end.toDate()];
 }
--- a/ui/component/or-attribute-barchart/src/timeframe.ts
+++ b/ui/component/or-attribute-barchart/src/timeframe.ts
@@ -1,4 +1,10 @@
+import moment from "moment";
+
 export type TimeframeDirection = "previous" | "next";
+
+export type TimeframeUnit = moment.unitOfTime.DurationConstructor;
+
+const CALENDAR_UNITS = new Set<TimeframeUnit>(["days", "weeks", "months", "quarters", "years"]);
 
 export function shiftTimeframeByDuration(
   currentStart: Date,
@@ -8,4 +14,23 @@ export function shiftTimeframeByDuration(
 ): [Date, Date] {
   const offset = direction === "previous" ? -duration : duration;
   return [new Date(currentStart.getTime() + offset), new Date(currentEnd.getTime() + offset)];
+}
+
+export function shiftTimeframe(
+  currentStart: Date,
+  currentEnd: Date,
+  duration: number,
+  unit: TimeframeUnit,
+  value: number,
+  direction: TimeframeDirection
+): [Date, Date] {
+  if (!CALENDAR_UNITS.has(unit)) {
+    return shiftTimeframeByDuration(currentStart, currentEnd, duration, direction);
+  }
+
+  const start = moment(currentStart);
+  const end = moment(currentEnd);
+  direction === "previous" ? start.subtract(value, unit) : start.add(value, unit);
+  direction === "previous" ? end.subtract(value, unit) : end.add(value, unit);
+  return [start.toDate(), end.toDate()];
 }
--- a/ui/component/or-attribute-barchart/src/timeframe.ts
+++ b/ui/component/or-attribute-barchart/src/timeframe.ts
@@ -1,3 +1,22 @@
+/*
+ * Copyright 2026, OpenRemote Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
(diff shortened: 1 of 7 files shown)


MartinaeyNL (changes requested):
The CI/CD does not seem to succeed, as the compile step fails. Can you take a look?
Should be reproducible locally using `./gradlew clean installDist`.

MartinaeyNL (changes requested):
The CI/CD still seems to fail on the component tests and the Spotless formatter check.
It's required to run `./gradlew splotlessApply` to apply the formatting rules of the repository.

wborn (changes requested):
*This is an additional AI-assisted review.*

The overall approach of keeping the requested timeframe authoritative rather than deriving it from bucket-centered datapoints looks appropriate, but the navigation regression is not fully fixed yet.

The existing navigation-duration concern still applies on the latest revision. The first navigation of a preset such as `this Hour` uses the configured 3,600,000 ms duration correctly, but `_shiftTimeframe()` then assigns `timeframe`. During the resulting update, `getNavigationDuration(..., !!this.timeframe, ...)` treats that navigation-generated timeframe as custom and derives the next duration from the inclusive start/end values, producing 3,599,999 ms. The second and subsequent navigation operations can therefore drift again. A regression test should exercise repeated navigation through the actual component lifecycle so this recomputation is covered.

The DST test also still fails in the current CI run. Configuring Playwright with `timezoneId: "Europe/Stockholm"` only changes the browser context, while `shiftTimeframe()` is imported and executed by the Node test process. On the UTC GitHub runner the test consequently receives 24 hours instead of the expected 23 hours. The DST assertion needs to execute in the configured browser timezone or otherwise run with a controlled Node timezone.

The current CI run additionally still fails the Spotless check for the new TypeScript files. That formatting issue has already been raised in the existing review feedback.

ui/component/or-attribute-barchart/src/timeframe.ts
```diff
@@ -0,0 +1,73 @@
...
+  unit: TimeframeUnit,
+  value: number,
+  direction: TimeframeDirection
+): [Date, Date] {
+  const normalizedUnit = normalizeUnit(unit);
+  if (!CALENDAR_UNITS.has(normalizedUnit)) {
+    return shiftTimeframeByDuration(currentStart, currentEnd, duration, direction);
+  }
+
+  const start = moment(currentStart);
```
wborn, line 69: Custom timeframes still inherit the previously selected preset's calendar semantics. For example, with `Day` selected, choosing a 3-hour custom timeframe correctly produces a 3-hour `duration`, but this branch ignores it and shifts both dates by one day. This means the "exact duration for custom windows" behavior only works when the previous preset uses a non-calendar unit. Custom timeframes should shift by their exact duration regardless of the previously selected preset, or the intended custom-window semantics should be represented separately from the preset unit.

wborn (changes requested):
The previously reported preset-navigation and DST test issues are addressed in the latest revision, and the current CI run is green.

One navigation case still remains. Custom timeframes currently retain the calendar semantics of the previously selected preset. For example, starting from `Day` and selecting a custom 3-hour range calculates the correct 3-hour navigation duration, but navigation still shifts the range by one calendar day because `shiftTimeframe()` chooses its behavior from the preset unit.

This should be covered with a component-level regression test using a calendar preset followed by an arbitrary custom duration.

*This review was AI-assisted.*

wborn (approved):
The previously reported custom-timeframe navigation issue is addressed in the latest revision. Custom windows now retain their exact elapsed duration even when the selected preset uses calendar semantics, and the component-level regression test covers that scenario.

The latest CI run is also green, including the distribution build, Spotless, UI component tests, UI app tests, and the NPM publish dry run.

No further blocking issues were identified in the current revision.

*This review was AI-assisted.*

MartinaeyNL (commented):
I'm happy with the code. 👍 
I have also tested the changes locally, ran the Manager UI, and everything looks OK to me.

Before merging, I'd like to double check the component testing changes with @Ekhorn.
Especially the timezone configuration change, together with the "unit testing" approach for the tests.

PR #3249, opened 2026-08-31
hotfix: Upgrade Camel to `4.22.0` to remove vulnerability with CVE ID `CVE-2026-78329`

pankalog:


MartinaeyNL (approved):

MartinaeyNL:
### Additional info on the vulnerability after asking clarification internally;

"We use `camel-undertow` only for the `ClientEventService` WebSocket endpoint. Although the application may contain a Camel version in the affected range, the vulnerable header propagation path is not reachable in our usage.
The Undertow consumer is not used to forward arbitrary request headers to an Undertow WebSocket producer.
Outbound WebSocket sends are created internally by `ClientEventService` and target an explicitly selected session/connection. We do not expose routes that allow untrusted messages or headers to be passed into the Undertow producer."

**"Therefore, we assess [CVE-2026-78329](https://github.com/advisories/GHSA-v7h8-xhh6-gfj4) as not exploitable in our current deployment.**
We should nevertheless upgrade Camel to a fixed version as part of normal dependency maintenance."

PR #3256, opened 2026-09-01
Show client secret length on the interconnect form #1365

### LINKED ISSUE #1365: Interconnect: the form should show client secret requirements and proper error message
Currently the requirement of 36 characters of the client secret is not check before saving and not shown in the helper text of the form input.

Also, the error snackbar that appears does not contain any information on why the save did not work.

AzazelSensei:
## Description

Interconnect client secret must be 36 characters. The field did not say that, save did not check it, and a failed save only showed a generic snackbar.

- Helper text on the client secret field
- Save stays disabled unless the secret is exactly 36 characters
- Snackbar uses the server error when save still fails

Fixes #1365

## Checklist

- [x] 4. Documentation is written or updated

### DIFF
--- a/ui/app/manager/src/pages/page-gateway.ts
+++ b/ui/app/manager/src/pages/page-gateway.ts
@@ -44,6 +44,7 @@ import "@openremote/or-components/or-ace-editor";
 import moment from "moment";
 import type { OrAceEditor } from "@openremote/or-components/or-ace-editor";
 import { showSnackbar } from "@openremote/or-mwc-components/or-mwc-snackbar";
+import { isAxiosError } from "@openremote/rest";
 
 export function pageGatewayProvider(store: Store<AppStateKeyed>): PageProvider<AppStateKeyed> {
   return {
@@ -391,6 +392,9 @@ export class PageGateway extends Page<AppStateKeyed> {
         <or-vaadin-text-field
           id="gateway-clientsecret"
           required
+          minlength="36"
+          maxlength="36"
+          helper-text=${i18next.t("gateway.clientSecretHelper")}
           ?disabled=${disabled}
           value=${connection?.clientSecret}
           @change=${(ev: Event) => this._setConnectionProperty("clientSecret", (ev.currentTarget as HTMLInputElement).value)}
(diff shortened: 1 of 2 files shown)


AzazelSensei:
Opened this for #1365.

pankalog:
@AzazelSensei  

Thanks for your submission! We're really happy to see people contributing to OpenRemote. For us to take a look at the contribution, we need you to sign the OpenRemote CLA, as [this comment](https://github.com/openremote/openremote/pull/3256#issuecomment-5491134580) above says. Once it's signed, we'll be happy to take a look and eventually merge your contribution!

AzazelSensei:
Signed the CLA.

ui/app/manager/src/pages/page-gateway.ts
```diff
@@ -687,17 +691,40 @@ export class PageGateway extends Page<AppStateKeyed> {
...
+      if (error.response?.statusText) {
+        return error.response.statusText;
+      }
+    }
+    if (statusText && statusText.trim()) {
+      return statusText;
+    }
+    return i18next.t("errorOccurred");
+  }
+
```
MartinaeyNL: I would simplify this to simple checks, instead of "modifying the snackbar based on HTTP status text".
Instead, I'd suggest to add a `else if (response.status === XX) { }` similar to the 204 check on line 691.
I would stick to;
- 401 = "Unauthorized",
- 403 = the default unexpected error (errorOccurred),
- and 400 something like "Fields are incorrect", or "Please double check the inputs"

In practice, this snackbar should barely appear, as the save button will be disabled beforehand.
So I wouldn't spend too much time handling edge cases like these 😉

ui/app/manager/src/pages/page-gateway.ts
```diff
@@ -397,6 +398,9 @@ export class PageGateway extends Page<AppStateKeyed> {
         <or-vaadin-text-field
           id="gateway-clientsecret"
           required
+          minlength="36"
+          maxlength="36"
+          helper-text=${i18next.t("gateway.clientSecretHelper")}
```
MartinaeyNL: I believe you can replace the `helper-text` attribute with a slot using `<or-translate>` instead.
It's a better practice in our codebase, along with automatically translating the text when a user switches language.

```
<or-vaadin-text-field ... >
    <or-translate slot="label" value="clientSecret"></or-translate>
    <or-translate slot="helper" value="gateway.clientSecretHelper"></or-translate>
</or-vaadin-text-field>
```

MartinaeyNL (changes requested):
Thanks for opening a pull request 🥳
Sorry for the delay on the review haha
 
I've added a few minor comments that should be relatively easy to resolve.
Feel free to ping me for questions or another review 😉

PR #3271, opened 2026-09-03
Use demo sourceSet for demo extensions

### LINKED ISSUE #3263: Demo extension broken in local development workflow
## Bug description
The last few months, I encountered issues with the Demo Setup run configuration in my development workflow.
It was common to see a Manager UI without the smartcity and manufacturer realms, and no assets at all.
The same can be said about the demo asset types not being present in the system.

I asked around a bit, and it turns out the `build.gradle` file within the `/manager` folder,
contains a dependency on the new Demo Setup extension from the `openremote/extensions` repository.
Apparently when using `demoExtension` instead of `implementation`, it's broken for local development use;
```gradle
  // Select the newest published demo extension up to and including the
  // current project version, falling back to an earlier version when the
  // current version has not been published yet.
  def demoExtensionMaxVersion = project.version.toString().replaceFirst(/-SNAPSHOT$/, "")
  demoExtension "io.openremote.extension:openremote-demo-setup-extension:(,${demoExtensionMaxVersion}]"
```

## To reproduce

1. Clone the `openremote/openremote` repository locally
2. Follow the guide on setting up the repository in the official OpenRemote documentation.
3. Run the "Demo Setup" run configuration within IntelliJ IDEA,
or start OpenRemote through Gradle manually with `OR_SETUP_TYPE=demo`
4. Open the Manager UI on `localhost:8080`, and see missing realms/assets.

## Expected behavior
The "Demo Setup" run configuration that we distribute out-of-the-box, should load the Manager with all realms,
asset types, assets included defined in that setup JAR.

## Device information
It has both been proven reproducible on Windows and Linux (@wborn), and it's expected to be unrelated to OS.
I was using the IntelliJ (v2026.1.5) run configuration, together with Gradle, to run the Manager.

## Additional context
I had a discussion in the office with @denniskuijs and @wborn about this, and they're aware of the issue.
They also have a bit more additional background knowledge on the bug and a possible fix.

wborn:
## Description

Use a dedicated `demo` source set in the `setup` project and move setup-specific Gradle logic out of the `manager` project.

This makes the demo extension available when running the Demo Setup from IntelliJ or with `OR_SETUP_TYPE=demo`, while keeping it out of the Manager's default runtime classpath. As a result, the extension is not included in the Manager distribution or Docker image and does not become part of the published OpenRemote dependencies.

The `setup` project now resolves the newest published demo extension up to and including the current project version, falling back to an earlier version when needed. Its transitive `io.openremote` dependencies are excluded so the local OpenRemote project classes are used instead.

A `setup:run` task now runs the Manager with the source set selected by `OR_SETUP_TYPE`, and the IntelliJ Demo Setup configuration uses the `setup.demo` module.

Fixes #3263

## Checklist


wborn:
I will see if  the demo source set can be moved to the setup project which contains the other setup types.

setup/build.gradle
```diff
@@ -63,3 +73,14 @@ tasks.register('installDist', Copy) {
...
 }
+
+tasks.register("run", JavaExec) {
+  group = "application"
+  description = "Runs the manager with the setup selected by OR_SETUP_TYPE"
+  mainClass = "org.openremote.manager.Main"
+  workingDir = rootProject.projectDir
+
+  def setupType = providers.environmentVariable("OR_SETUP_TYPE").orNull
+  def targetSourceSet = setupType != null ? sourceSets.named(setupType) : null
```
MartinaeyNL: **sourceSets.named(setupType) throws for unknown values**
`OR_SETUP_TYPE` is a free-form string that `SetupService` just forwards to `SetupTasks.createTasks`;
Third-party setup extensions can use arbitrary values, and master's removed code used null-safe with `findByName()`.
With named, e.g. `OR_SETUP_TYPE=custom ./gradlew setup:run` — or even `./gradlew tasks`, fails at configuration time with a cryptic SourceSet with name 'custom' not found.                                                 
I'd suggest `sourceSets.findByName(setupType)` with a fallback to main (optionally logging a warning).

setup/build.gradle
```diff
@@ -63,3 +73,14 @@ tasks.register('installDist', Copy) {
     into project(":manager").layout.buildDirectory.dir("install/manager/deployment/manager/extensions")
   }
 }
+
+tasks.register("run", JavaExec) {
```
MartinaeyNL, line 77: **Silent behavior change for manager:run**
On master, `OR_SETUP_TYPE=demo ./gradlew manager:run` adds the extension to the classpath;
after this PR it silently runs without it. Nothing points users to `:setup:run`.

The PR checklist's "Documentation is written or updated" is unmet;
at minimum a note in the run task description and wherever the dev workflow is documented.

MartinaeyNL: I don't know if there's ever been documentation on the tasks itself `manager:run`,
I guess we've only used the Gradle run configuration (like in IntelliJ IDEA) and nothing else.

setup/build.gradle
```diff
@@ -63,3 +73,14 @@ tasks.register('installDist', Copy) {
...
+  group = "application"
+  description = "Runs the manager with the setup selected by OR_SETUP_TYPE"
+  mainClass = "org.openremote.manager.Main"
+  workingDir = rootProject.projectDir
+
+  def setupType = providers.environmentVariable("OR_SETUP_TYPE").orNull
```
MartinaeyNL: **Empty-string OR_SETUP_TYPE now breaks configuration (regression)**                                                                                                                                                                           
`providers.environmentVariable("OR_SETUP_TYPE").orNull` returns `""` — not null.
When the variable is exported empty, e.g. `OR_SETUP_TYPE=${OR_SETUP_TYPE:-}`),
and mirrors the passthrough pattern in `docker-compose.yml:68`, then `sourceSets.named("")` throws.
Master's code `(if (setupType != null) + findByName)` tolerated an empty value as "unset".

Fix: treat blank as null, e.g. `.map(s -> s.isBlank() ? null : s).orElse(null)` style, or check `!setupType.empty`.

MartinaeyNL (commented):
Seems to work as expected 👍 
I've only checked the common scenarios I run OpenRemote locally with.

Just to double check, here are a few comments that my AI buddy generated;

MartinaeyNL (approved):

PR #3291, opened 2026-09-09
Update security policy

Ekhorn:
## Description

Adds reporter-facing guidance. Nothing changes about how vulnerabilities are reported.

- **Report one vulnerability at a time.** GitHub refuses a CVE for an advisory covering more than one independently fixable vulnerability, so a bundled report has to be split before any of it can be published, which delays the fix and the credit. Cites CNA rules 4.1, 4.2.6 and 4.2.11.
- **What to include.** The version or commit tested against, steps to reproduce, the affected code, and a full CVSS vector rather than a bare number. We prefer 4.0 over 3.1.
- **What happens next.** Outcome communicated either way, we reproduce before fixing, severity discussed with the reporter before it changes, duplicates merged with credit to everyone who found it, advisory published once a release contains the fix.
- Links to handbook section 6 for the full process.

## Checklist


SECURITY.md
```diff
@@ -19,3 +19,32 @@ To report through GitHub go to [https://github.com/openremote/openremote/securit
...
+
+Your report is reviewed and either accepted or rejected, and you are told the outcome either way.
+
+We reproduce the issue before writing a fix. If we cannot, we will come back to you rather than close it silently.
+
+We agree the severity within the team. If we arrive at a different score from the one you assigned, we will discuss it with you before changing it.
+
+If your report turns out to describe the same vulnerability as another one, the reports are merged and everyone who found it is credited on the advisory that is published.
+
+The advisory is published once a release containing the fix is publicly available.
```
wborn: This conflicts with the related advisory template, which explicitly treats a real vulnerability with no planned fix as publishable. GitHub also supports publishing an advisory without a fixed version and specifically documents the `not planned` case.

As written, this policy provides a publication path only when a fixed release exists. Could this distinguish between vulnerabilities with a planned fix and those that are published without one, with mitigations documented?

SECURITY.md
```diff
@@ -19,3 +19,32 @@ To report through GitHub go to [https://github.com/openremote/openremote/securit
...
+## What To Include
+
+- The version, release tag or commit hash you tested against. A floating tag such as `develop` does not identify a build, so we cannot tell what you were running.
+- Steps to reproduce, and what you observed that shows the issue is real rather than theoretical.
+- The affected endpoint, class or file.
+- The full CVSS vector rather than only a number, so the score can be checked. We prefer 4.0 over 3.1.
+
+## What Happens Next
+
+Your report is reviewed and either accepted or rejected, and you are told the outcome either way.
```
wborn: The related advisory process now has four distinct triage outcomes: `publishable`, `duplicate`, `split`, and `rejected`, and this policy itself describes duplicate handling below.

Saying every report is either accepted or rejected does not match that outcome model, particularly for reports that are split or merged as duplicates. Could this simply say that the report is reviewed and the reporter is told the outcome, or otherwise use the same outcome terminology as the tracker?

wborn (changes requested):

Two consistency issues should be resolved before merging. The publication rule currently excludes valid vulnerabilities for which no fix is planned, even though the related advisory workflow explicitly treats those as publishable. The triage wording also describes only accepted/rejected outcomes while the companion workflow distinguishes publishable, duplicate, split, and rejected reports.

Details are in the inline comments.

wborn (dismissed):

SECURITY.md
```diff
@@ -8,14 +8,43 @@ Fixes will only be released for previous releases under special circumstances.
...
 If you have a patch for the issue please use `git format-patch` and attach to the email or issue. Please do not open a
 pull request on GitHub as that may disclose sensitive details around the vulnerability.
+
+### What Happens Next
+
+Your report is reviewed and you are told the outcome, whatever it is.
+
+We reproduce the issue before writing a fix. If we cannot, we will come back to you rather than close it silently.
+
+We agree the severity within the team. If we arrive at a different score from the one you assigned, we will discuss it with you before changing it.
```
ebariaux, line 44: The handbook does not specifically indicate that the score must be discussed with the reporter before being changed. Also not 100% sure we want this.
Also, if we want to be really formal, we should define exceptions, like if we believe the vulnerability is really critical or if the reviewer does not respond to our proposed change within x hours.

What about

If we arrive at a different score from the one you assigned, we will provide a clear rationale for the reason we made the change.

ebariaux (commented):
In general, great job on all the vulnerability related process !

ebariaux (approved):

Issue #3106, opened 2026-07-24
Demo setup is missing when deploying to test servers


denniskuijs:
## Description
After the changes to our Docker images (core), the demo setup is no longer automatically included when deploying to test servers.

Francisemeka:
I am currently working on this and i will be dropping a PR soon

Issue #2598, opened 2026-03-09
Introduce Extensions mechanism


wborn:
Decouple domain-specific logic (protocols, asset models, and services) from the platform Core. This ensures a leaner Core repository and optimizes resource usage for edge deployments.

PR #3327, opened 2026-09-19
Fix demo setup packaging for test deployments

Francisemeka:
The demo setup extension was available when running the manager locally but was not included in the deployment image.

This change exposes the demo setup extension through a consumable manager configuration and includes it in the deployment image under /deployment/manager/extensions.

This allows the demo SetupTasks provider to be discovered through ServiceLoader when deploying to test servers.

Verified that the generated deployment image contains:

openremote-demo-setup-extension-1.30.0.jar


wborn:
> The demo setup extension was available when running the manager locally but was not included in the deployment image.

The demo is nowadays part of the `openremote/manager` image which is created in the [openremote/extensions](https://github.com/openremote/extensions) repo. This is also how we use it on https://demo.openremote.app/

More optional logic will eventually be moved into extensions, see https://github.com/openremote/openremote/issues/2598.

Francisemeka:
Thanks, I understand.

I was following the demo extension dependency in the main repo and found that it was available for the local Manager run but wasn't being included in the deployment image.

If the demo is now packaged in the `openremote/manager` image from the `extensions` repo, should this issue be fixed there instead?

wborn:
I don't think any change is needed in the `extensions` repo. The `demo-setup` extension is already part of the standard `openremote/manager` image, while the core image intentionally excludes extensions.

#3106 is specifically about deploying PR branches to test servers. The PR deployment flow could probably either use the standard `openremote/manager` image and overlay the Manager code built from the PR branch, or only add the `demo-setup` extension to the Manager image built from `openremote/openremote` for these test-server deployments.

PR #3335, opened 2026-09-22
Historic facade security enforcement

### LINKED ISSUE #3333: Enforce realm / asset scope limitation on rules engines for HistoricFacade
Operations exposed to rules engines via HistoricFacade do not perform any checks on scoped access.

Implement proper validation that Realm scoped rules engines and asset scoped rules engine only access assets they are allowed to.

ebariaux:


richturner (dismissed):

manager/src/main/java/org/openremote/manager/rules/facade/FacadeHelper.java
```diff
@@ -0,0 +1,75 @@
...
+import org.openremote.manager.rules.RulesEngineId;
+import org.openremote.model.asset.Asset;
+import org.openremote.model.query.AssetQuery;
+import org.openremote.model.query.filter.PathPredicate;
+import org.openremote.model.query.filter.RealmPredicate;
+import org.openremote.model.rules.AssetRuleset;
+import org.openremote.model.rules.RealmRuleset;
+import org.openremote.model.rules.Ruleset;
+
+public class FacadeHelper {
```
richturner, line 31: Indeed a static singleton should have a private constructor

richturner (changes requested):

manager/src/main/java/org/openremote/manager/rules/facade/FacadeHelper.java
ebariaux: Added

richturner (approved):

